PR #1, opened 2022-11-30
Added Keycloak backstage plugin

sabre1041:
Initial version for the Keycloak backstage plugin

### DIFF
--- a/plugins/keycloak-backend/package.json
+++ b/plugins/keycloak-backend/package.json
@@ -1,0 +1,49 @@
+{
+  "name": "@janus-idp/keycloak-backend",
+  "version": "0.1.0",
+  "description": "A Backend backend plugin for Keycloak",
+  "main": "dist/index.cjs.js",
+  "types": "src/index.ts",
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "backstage": {
+    "role": "backend-plugin"
+  },
+  "scripts": {
+    "start": "backstage-cli package start",
+    "build": "backstage-cli package build",
+    "lint": "backstage-cli package lint",
+    "test": "backstage-cli package test",
+    "clean": "backstage-cli package clean",
+    "prepack": "backstage-cli package prepack",
+    "postpack": "backstage-cli package postpack"
+  },
+  "dependencies": {
+    "@backstage/backend-common": "^0.16.0",
+    "@backstage/backend-tasks": "^0.3.7",
+    "@backstage/config": "^1.0.4",
+    "@backstage/catalog-model": "1.0.0",
+    "@backstage/plugin-catalog-backend": "1.5.1",
+    "@keycloak/keycloak-admin-client": "^18.0.2",
+    "lodash": "^4.14.189",
+    "winston": "^3.2.1",
+    "uuid": "^8.0.0"
+  },
+  "devDependencies": {
+    "@backstage/cli": "0.21.1",
+    "@types/lodash": "^4.14.189",
+    "@types/supertest": "^2.0.12",
+    "@types/uuid": "^8.0.0",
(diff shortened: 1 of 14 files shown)


tumido:
CI fixed by https://github.com/janus-idp/backstage-plugins/pull/2

plugins/keycloak-backend/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "clean": "backstage-cli package clean",
+    "prepack": "backstage-cli package prepack",
+    "postpack": "backstage-cli package postpack"
+  },
+  "dependencies": {
+    "@backstage/backend-common": "^0.16.0",
+    "@backstage/catalog-model": "1.0.0",
+    "@backstage/backend-tasks": "^0.3.7",
+    "@backstage/config": "^1.0.4",
+    "@backstage/plugin-catalog-backend": "1.5.1",
```
tumido, line 30: These should be `peerDependencies`

plugins/keycloak-backend/src/lib/config.ts
```diff
@@ -0,0 +1,143 @@
...
+      username,
+      password,
+      clientId,
+      clientSecret,
+      schedule,
+    }
+  });
+
+
+}
```
tumido, line 143: Missing proper EOL at EOF

plugins/keycloak-backend/src/lib/index.ts
```diff
@@ -0,0 +1,29 @@
...
+} from './types';
+export type {
+    KeycloakProviderConfig,
+  } from './config';
+  export {
+    KEYCLOAK_HOST_ANNOTATION,
+    KEYCLOAK_ID_ANNOTATION,
+    KEYCLOAK_REALM_ANNOTATION,
+  } from './constants';
+
```
tumido, line 29: EOF

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -0,0 +1,239 @@
...
+    config: KeycloakProviderConfig,
+): Promise<{
+    users: UserEntity[];
+}> {
+
+    const users: UserEntity[] = [];
+
+    const keycloakUsers = await client.users.find({ realm: config.realm, });
+
+    for (const keycloakUser of keycloakUsers) {
```
tumido, line 119: I think you'd be better with a `.map()` here than a `for` and `push` here.

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -0,0 +1,239 @@
...
+
+        if (keycloakUser.email) entity.spec.profile!.email = keycloakUser.email;
+        if (displayName) entity.spec.profile!.displayName = displayName;
+
+        users.push(entity);
+
+
+    }
+
+    return { users };
```
tumido, line 150: I think you can return an array here instead.

plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.ts
```diff
@@ -0,0 +1,331 @@
...
+        groupSearch: provider.groupSearch,
+        groupSelect: provider.groupSelect,
+        queryMode: provider.queryMode,
+        groupTransformer: this.options.groupTransformer,
+        userTransformer: this.options.userTransformer,
+        organizationTransformer: this.options.organizationTransformer,
+        logger: logger,
+      },
+    );
+*/
```
tumido, line 248: Not sure this is intended to be here.

tumido (changes requested):
I think we should refactor some of the loops into `.map()` or `.reduce()` calls instead. However that can also be done later...

Can you please run `prettier`/ `yarn lint` against the code? There are quite some formatting issues to be fixed (so it respects the `@spotify/prettier-config` standard)

tumido (approved):
/lgtm
/approve

tumido (approved):

PR #6, opened 2022-12-05
chore(ci): Add tsc before build step

tumido:
`tsc` step was missing from the build step. See logs at:

https://github.com/janus-idp/backstage-plugins/actions/runs/3623692907/jobs/6109842475


tumido:
Ignore failing build, it fails because there's no package to build now. Which is expected. This will go away once we check in any plugin code.

sabre1041 (approved):
LGTM

PR #9, opened 2022-12-06
fix: typo in keycloak annotations

tumido:
I missed that when reviewing #1 .

I think you've meant to use `keycloak.org` everywhere when changing it from `backstage.io`


sabre1041 (approved):
LGTM

PR #11, opened 2022-12-06
refactor: Use arrow functions, maps and inlines

tumido:
First chunk of refactoring on this. Simplify some of the calls, use inlines where it made sense, remove async from where it was not needed.


christophe-f (approved):

sabre1041 (approved):
LGTM

PR #12, opened 2022-12-07
chore(ci): Fix release pipeline [attempt 2]

tumido:
As per https://github.com/dhoulb/multi-semantic-release#npm-v85-npm-err-notarget-no-matching-version-found-for and https://github.com/dhoulb/multi-semantic-release/issues/129

I think this might be it. This is my last attempt to fix the `semantic-release` if this doesn't work I'll experiment with some other releasers like [intuit/auto](https://intuit.github.io/auto/)...


sabre1041 (approved):
LGTM

PR #42, opened 2022-12-08
fix(ocm-backend): Package config schema with the plugin

tumido:
SSIA


tumido (commented):
self merging again. :see_no_evil:

PR #43, opened 2022-12-08
fix(ocm-backend): Package config schema with the plugin - retry

tumido:
Let's try it once more...


tumido (commented):
Shameless self merging leads to not fixing things properly... :facepalm:

PR #44, opened 2022-12-08
docs(ocm): Documentation for OCM plugins

tumido:
Full documentation (without screenshots for now, since I expect the UI to change substantially soon).

Adds an installation and configuration guide for OCM plugins.

### DIFF
--- a/plugins/ocm/README.md
+++ b/plugins/ocm/README.md
@@ -1,10 +1,280 @@
-# Open Cluster Management
+# Open Cluster Management plugin for Backstage
 
-Welcome to the Open Cluster Management plugin!
+This plugin integrates your Backstage instance with Open Cluster Management's MultiClusterHub and MultiCluster Engine.
 
-_This plugin was created through the Backstage CLI_
+## Prerequisites
 
-## Getting started
+1. OCM deployed and configured on a kubernetes cluster
+2. Installed [Kubernetes plugin for Backstage](https://backstage.io/docs/features/kubernetes/overview)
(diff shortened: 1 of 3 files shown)


plugins/ocm/README.md
```diff
@@ -1,10 +1,246 @@
...
+      ...
++     <EntitySwitch.Case if={isKind('resource')} children={resourcePage} />
+      ...
+    </EntitySwitch>
+  );
+
+
+```
+
+#### ClusterStatusCard
```
raffaelespazzoli, line 227: should we add example on how to use these cards

tumido: I've added that. since it's very layout dependent, it's not very helpful though.

plugins/ocm/README.md
```diff
@@ -1,10 +1,278 @@
...
+  - Cluster current status (up, down, etc.)
+  - Cluster details (console link, OCP and Kubernetes version, etc.)
+  - Details about available compute resources on the cluster
+
+## Usage
+
+This plugin is made up of 2 packages:
+
+1. `@janus-idp/backstage-plugin-ocm-backend` connects backstage server to OCM.
+2. `@janus-idp/backstage-plugin-ocm` contains frontend components, requires `*-backend` to be present and set up.
```
SamoKopecky, line 42: I think the `@janus-idp/backstage-plugin-ocm-common` should also be listed here since it contains the definition of the `ClusterDetails` object.

SamoKopecky (commented):
Looks good! I just have two small suggestions bellow.

plugins/ocm/README.md
SamoKopecky: Yea that makes sense.

SamoKopecky:
Looks good!
/lgtm

SamoKopecky (approved):

PR #46, opened 2022-12-12
fix(keycloak): Downgrade to keycloak client v18

tumido:
Keycloak client decided to drop support for CommonJS by v19. Backstage doesn't properly support ESM yet, therefore ESM only dependencies don't build properly via `backstage-cli package build`. This is a temporary workaround until Backstage works out proper ESM support. Locking to `<19` version of Keycloak client. 

Related to: #47


sabre1041 (approved):
Old is the new new! LGTM

PR #67, opened 2023-01-06
feat(ocm): Add update information to the `/status` API endpoint

SamoKopecky:
Serve as the backend for https://github.com/operate-first/service-catalog/issues/179

Add additional API call to gather information about cluster updates and provide it in the `/status` API endpoint. Also  create tests for newly created functions.

I couldn't find an endpoint that only provides a single cluster, so I have to use the endpoint (`internal.open-cluster-management.io/v1beta1/managedclusterinfos`) even in the `/status/:clusterName` API endpoint.

Example of the new API response on `/status/smaug` (see the `update` section at the end):
```json
{
  "name": "smaug",
  "status": {
    "available": true,
    "reason": "Managed cluster is available"
  },
  "consoleUrl": "https://console-openshift-console.apps.smaug.na.operate-first.cloud",
  "kubernetesVersion": "v1.22.3+fdba464",
  "oauthUrl": "https://oauth-openshift.apps.smaug.na.operate-first.cloud/oauth/token/implicit",
  "openshiftId": "5d448ae7-05f1-42cc-aacc-3122a8ad0184",
  "openshiftVersion": "4.9.21",
  "platform": "BareMetal",
  "allocatableResources": {
    "cpuCores": 1136.5,
    "memorySize": "7469511796Ki",
    "numberOfPods": 7750
  },
  "availableResources": {
    "cpuCores": 1152,
    "memorySize": "7505192052Ki",
    "numberOfPods": 7750
  },
  "update": {
    "available": true,
    "version": "4.9.53",
    "url": "https://access.redhat.com/errata/RHBA-2022:8714"
  }
}
```

### DIFF
--- a/plugins/ocm-backend/src/helpers/parser.ts
+++ b/plugins/ocm-backend/src/helpers/parser.ts
@@ -1,5 +1,6 @@
 import { CONSOLE_CLAIM } from '../constants';
 import { ClusterDetails } from '@janus-idp/backstage-plugin-ocm-common';
+import { maxSatisfying } from 'semver';
 
 const convertCpus = (cpus: string | undefined): number | undefined => {
   if (!cpus) {
@@ -56,5 +57,31 @@ export const parseManagedCluster = (cluster: any): ClusterDetails => {
   return {
     ...status,
     ...parsedClusterInfo,
+  };
+};
+
+export const parseUpdateInfo = (clusterInfo: any) => {
+  const { availableUpdates, versionAvailableUpdates } =
+    clusterInfo.status.distributionInfo.ocp;
+  /*
+   * We assume here that if availableUpdates is empty
+   * versionAvailableUpdates also has to be empty
+   */
+  if (!availableUpdates || availableUpdates.length === 0) {
+    return {
+      update: {
+        available: false,
+      },
+    };
+  }
+
+  const version = maxSatisfying(availableUpdates, '*');
+
+  return {
+    update: {
+      available: true,
+      version,
+      url: versionAvailableUpdates[availableUpdates.indexOf(version)].url,
+    },
   };
 };
--- a/plugins/ocm-backend/package.json
+++ b/plugins/ocm-backend/package.json
@@ -33,12 +33,13 @@
     "express": "^4.17.1",
     "express-promise-router": "^4.1.0",
     "node-fetch": "^3.0.0",
+    "semver": "^7.3.8",
     "winston": "^3.2.1",
     "yn": "^5.0.0"
   },
   "devDependencies": {
-    "@types/express": "4.17.14",
     "@backstage/cli": "0.21.1",
+    "@types/express": "4.17.14",
     "@types/supertest": "2.0.12",
     "msw": "0.49.1",
     "nock": "13.2.9",
--- a/plugins/ocm-backend/src/helpers/kubernetes.test.ts
+++ b/plugins/ocm-backend/src/helpers/kubernetes.test.ts
@@ -4,6 +4,7 @@ import {
   hubApiClient,
   getManagedCluster,
   getManagedClusters,
+  getManagedClustersInfo,
 } from './kubernetes';
 import { createLogger } from 'winston';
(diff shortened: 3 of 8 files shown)


plugins/ocm-backend/src/helpers/parser.ts
```diff
@@ -58,3 +60,41 @@ export const parseManagedCluster = (cluster: any): ClusterDetails => {
...
+  };
+};
+
+export const getClusterInfoByName = (
+  clusterInfos: Array<any>,
+  clusterName: string,
+  logger: Logger,
+) => {
+  const clusterInfo = clusterInfos.filter(
+    (cluster: any) => cluster.metadata.name === clusterName,
```
tumido, line 91: There is no way there are multiple kubernetes resources with the same name. This is not allowed by design. You can use `.find()` instead of filter directly.

plugins/ocm-backend/src/helpers/parser.ts
```diff
@@ -58,3 +60,41 @@ export const parseManagedCluster = (cluster: any): ClusterDetails => {
     ...parsedClusterInfo,
   };
 };
+
+export const parseUpdateInfo = (clusterInfo: any) => {
+  const versionAvailableUpdates =
+    clusterInfo.status.distributionInfo.ocp.versionAvailableUpdates;
+  const versionsAvailable =
```
tumido, line 67: Why this name differs from the property name if the previous constant is named the same?

Can't you just unpack both from the object?

```ts
const { versionAvailableUpdates, versionsAvailable } = clusterInfo.status.distributionInfo.ocp;
```

plugins/ocm-backend/src/helpers/parser.ts
```diff
@@ -58,3 +60,41 @@ export const parseManagedCluster = (cluster: any): ClusterDetails => {
...
+    : undefined;
+  const url = versionAvailableUpdates
+    ? versionAvailableUpdates[versionsAvailable.indexOf(version)].url
+    : undefined;
+
+  return {
+    update: {
+      available: versionsAvailable ? versionsAvailable.length !== 0 : false,
+      version: version,
+      url: url,
```
tumido, line 80:
```suggestion
      version,
      url,
```

plugins/ocm-backend/src/helpers/parser.ts
```diff
@@ -58,3 +60,41 @@ export const parseManagedCluster = (cluster: any): ClusterDetails => {
...
+  const version = versionsAvailable
+    ? maxSatisfying(versionsAvailable, '*')
+    : undefined;
+  const url = versionAvailableUpdates
+    ? versionAvailableUpdates[versionsAvailable.indexOf(version)].url
+    : undefined;
+
+  return {
+    update: {
+      available: versionsAvailable ? versionsAvailable.length !== 0 : false,
```
tumido, line 78: We knew this much earlier. Why don't we return right away if there are no updates? Instead you condition on the `versionsAvailable` 3 times (lines 69,73 and here).

tumido (changes requested):

plugins/ocm-backend/src/helpers/parser.ts
SamoKopecky: Resolved in the lastest commit.

plugins/ocm-backend/src/helpers/parser.ts
SamoKopecky: Resolved in the lastest commit.

plugins/ocm-backend/src/helpers/parser.ts
SamoKopecky: Resolved in the lastest commit.

plugins/ocm-backend/src/helpers/parser.ts
SamoKopecky: Resolved in the lastest commit.

SamoKopecky:
Thanks for the feedback, changes pushed with a new commit.

tumido (approved):
/lgtm
/approve

PR #73, opened 2023-01-18
feat(ocm): Add a frontend indicator for cluster upgrades

SamoKopecky:
Resolves https://github.com/operate-first/service-catalog/issues/179

Add an indicator for displaying whether an upgrade is available. Additionally add a mouse over tooltip showing which version is available linking to the Red Hat Product Erreta.

Since we want to make this look like the ACM indicator (see the comparison bellow) an external icon would have to be used (ACM uses patternfly for icons). This would require us to import an external `svg` which I don't think is very clean.
image
image

The closest match I found in the material ui icons is this.
image


tumido (approved):
/lgtm
/approve

PR #74, opened 2023-01-19
test(ocm): Add router tests

SamoKopecky:
Part of https://github.com/operate-first/service-catalog/issues/176

Add tests for the `router.ts` file which contains endpoint definition for the `ocm-backend` plugin.

Kubernetes API is mocked by [nock](https://www.npmjs.com/package/nock).


tumido (approved):
/lgtm
/approve

Issue #81, opened 2023-01-26
🔌 Plugin: Web Terminal


serenamarie125:
## Goal
The goal of this epic is to develop a new Backstage UI plugin in the Janus repo. This UI plugin should provide similar functionality from the OpenShift Web Terminal which is available in the OpenShift Console.

This plugin will provide an initial set of features, which will continue to be enhanced, based on the success metrics of the initial version of this plugin, including customer feedback.

**What problem does this solve? Please describe.**
-- tbd -- 

**Use cases**
As a backstage user, I want easy access to a web terminal running on a cluster in context

**Acceptance criteria**
-- tbd -- 

## Additional context

Web Terminal Operator: https://github.com/redhat-developer/web-terminal-operator 

## Screenshots of
![Screenshot 2023-02-07 at 1 16 34 PM](https://user-images.githubusercontent.com/12504403/217331372-092417b2-46bd-455b-9cc5-cc0ed6869c42.png)

![Screenshot 2023-02-07 at 1 16 22 PM](https://user-images.githubusercontent.com/12504403/217331376-88990020-42e1-460c-9a10-83f109da5b0c.png)

![Screenshot 2023-02-07 at 1 16 10 PM](https://user-images.githubusercontent.com/12504403/217331378-6879c33b-2cbe-4969-878a-237dcb3f494b.png)
 OpenShift web terminal for reference

PR #94, opened 2023-02-01
feat: Add API server annotation

mimotej:
Adds annotation `kubernetes.io/api-server` to a Cluster resource
Related to: https://github.com/operate-first/service-catalog/pull/210
Part of: https://github.com/janus-idp/backstage-plugins/issues/81

### DIFF
--- a/plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
+++ b/plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
@@ -24,6 +24,7 @@ import {
 import {
   ANNOTATION_ORIGIN_LOCATION,
   ANNOTATION_LOCATION,
+  ANNOTATION_KUBERNETES_API_SERVER,
 } from '@backstage/catalog-model';
 import { CustomObjectsApi } from '@kubernetes/client-node';
 import {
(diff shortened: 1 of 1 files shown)


plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
```diff
@@ -91,6 +92,8 @@ export class ManagedClusterProvider implements EntityProvider {
           name: i.metadata.name,
           title: i.metadata?.labels?.name,
           annotations: {
+            [ANNOTATION_KUBERNETES_API_SERVER]:
+              i.spec?.managedClusterClientConfigs?.[0]?.url,
```
tumido, line 96: Is this always located on index 0, isn't there a better marker?

mimotej: Not sure... The `managedClusterClientConfigs` key looks like this:
`managedClusterClientConfigs: [ [Object] ]`
where Object is `{ caBundle: val, url: val}`, I am not sure what other Objects can be inserted into `managedClusterClientConfigs` but the clusters which are managed by our OCM don't have anything else in here. I also don't see there anything that could be used for example with `.find()`. Do you have any ideas?

tumido: I think you are right... Unless we try to pull this from `ManagedClusterInfo` (`.spec.masterEndpoint`). That may introduce unnecessary processing complexity. Let's keep it this way unless we face issues. Can you please add a comment that it can also be pulled from ^ with a link to this discussion?

mimotej: Resolved in latest commit :+1:

tumido (approved):
/lgtm
/approve

tumido (approved):

Issue #100, opened 2023-02-08
fix(quay): Format date, size and Manifest information to be same as on the quay UI


fmenesesg:
The "LAST MODIFIED" field is showing a full timestamp  "Thu, 23 Jun 2022 20:31:28 -0000" and should be like this "Jun 23, 2022, 4:31 PM" as the Quay interface show.

The SIZE is on bytes like this "1656016288" and should be MB or GB according to a calculation.

MANIFEST value should be a substring(0, 19) instead of the all value

tumido:
Resolved via #101

PR #101, opened 2023-02-08
fix(quay): Format date, size and Manifest information to be same as on the quay UI

fmenesesg:
#100

 Format date, size and Manifest information to be same as on the quay UI

### DIFF
--- a/plugins/quay/src/components/utils.tsx
+++ b/plugins/quay/src/components/utils.tsx
@@ -1,45 +1,43 @@
-import { VulnerabilitySeverity } from "../types";
+import { VulnerabilitySeverity } from '../types';
 
 export function getSeverityColor(severity: VulnerabilitySeverity) {
-    switch (severity) {
-      case VulnerabilitySeverity.Critical:
-        return '#7D1007';
-      case VulnerabilitySeverity.High:
-        return '#C9190B';
-      case VulnerabilitySeverity.Medium:
-        return '#EC7A08';
-      case VulnerabilitySeverity.Low:
-        return '#F0AB00';
-      case VulnerabilitySeverity.None:
-        return '#3E8635';
-      default:
-        return '#8A8D90';
-    }
+  switch (severity) {
+    case VulnerabilitySeverity.Critical:
+      return '#7D1007';
+    case VulnerabilitySeverity.High:
+      return '#C9190B';
+    case VulnerabilitySeverity.Medium:
+      return '#EC7A08';
+    case VulnerabilitySeverity.Low:
+      return '#F0AB00';
+    case VulnerabilitySeverity.None:
+      return '#3E8635';
+    default:
+      return '#8A8D90';
   }
-  
-  export function formatDate(date: string | number) {
-    if (date == -1) {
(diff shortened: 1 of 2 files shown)


tumido (changes requested):
There seems to be some formatting errors in here. Can you try running `prettier` please?

fmenesesg:
I execute prettier, now should be ok, can you PTAL

plugins/quay/src/components/utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function formatSize(sizeInBytes: number) {
+  if (!sizeInBytes) {
+    // null or undefined
+    return 'N/A';
+  }
+
+  const i = Math.floor(Math.log(sizeInBytes) / Math.log(1024));
+  return `${Number((sizeInBytes / Math.pow(1024, i)).toFixed(2)) * 1} ${
```
tumido, line 40: Hm, can you explain me what is the `* 1` magic about? I'm unable to find a functional difference.

```ts
  return `${Number((sizeInBytes / Math.pow(1024, i)).toFixed(2)) * 1} ${
                                                                 ^^^^
```

plugins/quay/src/components/utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+    ['B', 'kB', 'MB', 'GB', 'TB'][i]
```
tumido, line 41: Wouldn't it be better if this array was a global constant? I may be too picky here, just a nit.

plugins/quay/src/components/utils.tsx
```diff
@@ -0,0 +1,43 @@
...
+    case VulnerabilitySeverity.Medium:
+      return '#EC7A08';
+    case VulnerabilitySeverity.Low:
+      return '#F0AB00';
+    case VulnerabilitySeverity.None:
+      return '#3E8635';
+    default:
+      return '#8A8D90';
+  }
+}
```
tumido, line 18: I think a constant would be nice here, but I may be too picky:

```suggestion
export function getSeverityColor(severity: VulnerabilitySeverity) {
export const SEVERITY_COLORS = new Proxy(
  {
    VulnerabilitySeverity.Critical: '#7D1007',
    VulnerabilitySeverity.High: '#C9190B',
    VulnerabilitySeverity.Medium: '#EC7A08',
    VulnerabilitySeverity.Low: '#F0AB00',
    VulnerabilitySeverity.None: '#3E8635'
  },
  {
    get: (o, k) => k in o ? o[k] : '#8A8D90'
  }
)
```

This should be easily addressable via `SEVERITY_COLORS[severity]`.


Again, just a nit, if you don't think agree, feel free to leave as it is, your implementation is completely reasonable and understandable.

tumido (commented):
LGTM overall, I've left couple of suggestions. Feel free to dismiss them, this is mergeable as is.

tumido (approved):
After DM discussion the nits are filed as a separate refactor issue. It's minor changes anyways... We can merge this as is. LGTM :+1:

Issue #89, opened 2023-01-30
App Topology - Topology view in backstage plugin should update based on selected theme (Dark/Light/System)


invincibleJai:
Description
As a user, I want to view the topology in the theme as set by the user in the settings i.e Dark, Light, System

Acceptance Criteria

- [x]  Topology view in the backstage plugin should be shown as the selected theme in backstage
- [x]  Push the code to GitHub and any changes need to be documented as required.(Needs to check on with GitHub repo)

 
Additional Details:

invincibleJai:
Creating stories for https://github.com/janus-idp/backstage-plugins/issues/58 

cc @serenamarie125 / @rtaniwa

invincibleJai:
associated PR https://github.com/janus-idp/backstage-plugins/pull/104

Issue #85, opened 2023-01-27
App Topology - Visualize resources of the Deployment kind as readonly


serenamarie125:
@invincibleJai please update this with initial story for https://github.com/janus-idp/backstage-plugins/issues/58

**Description**
As a developer, I want to visualize resources of the Deployment kind with pods ring demonstrating pod status and decorators for the route in a sample plugin for topology as Read-only. The visualization should be the same as in ODC as part of this work mock the data for the required resources and create the required data model.

**Acceptance Criteria**

- [x] Visualize resource of kind deployment and deployment config in Topology, show associated pod rings along with pod status
- [x] Workload resources like Deployments and Pods will be shown in a visual group based on the below label
   ```yaml
   'app.kubernetes.io/part-of': <GROUP_NAME>
   ```
- [x] Workload resources like Deployment or Pod will be shown with a visual connector based on the below annotations
   ```yaml
   'app.openshift.io/connects-to': '[{"apiVersion": <RESOURCE_APIVERSION>,"kind": <RESOURCE_KIND>,"name": <RESOURCE_NAME>}]'
   ```
- [x] Push the code to GitHub and any changes need to be documented as required. (Needs to check on with GitHub repo)

**Additional Details:**

For labels and annotations check https://github.com/redhat-developer/app-labels/blob/master/labels-annotation-for-openshift.adoc

invincibleJai:
Creating stories for https://github.com/janus-idp/backstage-plugins/issues/58 

Feel free to update as needed

cc @serenamarie125 @rtaniwa

invincibleJai:
associated PR https://github.com/janus-idp/backstage-plugins/pull/104

Issue #58, opened 2022-12-20
🔌  UI Plugin: Application Topology


serenamarie125:
## Goal
The goal of this epic is to develop a  new Backstage UI plugin in the Janus repo.  This UI plugin should provide similar functionality from the Application Topology view in the OpenShift Developer Perspective of the console.  The current Application Topology supports a large set of features.

This plugin will provide an initial set of features, which will continue to be enhanced, based on the success metrics of the initial version of this plugin, including customer feedback.

- The plugin will initially be view only
- The plugin will initially only provide the graph view
- The plugin will visualize information/status in real time
- The visual representations of objects from a shape and hierarchy standpoint will be consistent with what is shown in OpenShift today.
- Nodes will display a pod ring so that users are easily able to see the status of the pods associated with the workloads
- Provide a route decorator
- Side panel showing details of the kubernetes objects
- Resource tab in the side panel showing the associated resources 
- Implement the GH and DevSpaces decorator (stretch)

**What problem does this solve? Please describe.**
- Users want to be able to see the status of the workloads associated with a service in your catalog which is deployed to any kubernetes cluster. 
- Users want one click access to pod logs
- Users want one click access to run their application which is running on kubernetes

**Features**
_Need to look into this further_
- Provide a Topology tab on the entity page which provides the graph view of the application topology ( list view will. come later)
_OR_
- Provide a Topology Card on that visualizes the directly related workloads of the current entity.
- Provide a Topology Page which can be access from the Topology Card, and is able to view the workloads associated with the current entity as well as workloads associated with their relations.
 
**Use cases**
_Coming after spike review_

**Acceptance criteria**
_Coming after spike review_

## Issues in Epic
- [x] https://github.com/janus-idp/backstage-plugins/issues/85
- [x] https://github.com/janus-idp/backstage-plugins/issues/89
- [x] https://github.com/janus-idp/backstage-plugins/issues/90
- [x] https://github.com/janus-idp/backstage-plugins/issues/91
- [x] https://github.com/janus-idp/backstage-plugins/issues/105
- [x] https://github.com/janus-idp/backstage-plugins/issues/118
- [x] https://github.com/janus-idp/backstage-plugins/issues/121
- [x] https://github.com/janus-idp/backstage-plugins/issues/122
- [x] https://github.com/janus-idp/backstage-plugins/issues/123
- [x] #109 (once initial PR has been merged)
- [x] https://github.com/janus-idp/backstage-plugins/issues/70
[...]

PR #104, opened 2023-02-09
feat(topology): Topology plugin visualisation of workload

invincibleJai:
**Tracks:** https://github.com/janus-idp/backstage-plugins/issues/58

**Fixes:**  https://github.com/janus-idp/backstage-plugins/issues/85, https://github.com/janus-idp/backstage-plugins/issues/89, https://github.com/janus-idp/backstage-plugins/issues/105

**Screenshots/GIF's:**
- Light Theme
image

- Dark Theme
image

- Empty State
image

**Unit Tests:**
image

**Test Setup:**
Install and configure Kubernetes plugin by following these guides:

https://backstage.io/docs/features/kubernetes/installation
https://backstage.io/docs/features/kubernetes/configuration

Kubernetes plugin is properly configured and able to connect to the cluster via a ServiceAccount

Following ClusterRole must be granted to ServiceAccount accessing the cluster:
(k8s plugin by default tries to fetch the following resources and if access is not provided an error is shown)

```
kind: ClusterRole
metadata:
  name: backstage-read-only
rules:
  - apiGroups:
      - '*'
    resources:
      - pods
      - configmaps
      - services
      - deployments
      - replicasets
      - horizontalpodautoscalers
      - ingresses
      - statefulsets
      - limitranges
      - daemonsets
    verbs:
      - get
      - list
      - watch
  - apiGroups:
      - batch
    resources:
      - jobs
      - cronjobs
    verbs:
      - get
      - list
      - watch
  - apiGroups:
      - metrics.k8s.io
    resources:
      - pods
    verbs:
      - get
      - list
 ```
 
Create a service component entity - https://github.com/temp-demo-organisation/nationalparks-py/blob/master/catalog-info2.yaml

To get the resources from a k8s cluster do the following:

K8s plugin identifies if the provided entity has k8s resources and if yes from which namespace it should get the resources based on the following annotations added to the entity's catalog-info.yaml
```
annotations: 
 backstage.io/kubernetes-id: nationalparks-py
 backstage.io/kubernetes-namespace: div
```

For k8s plugin to get the desired entity's k8s resources the following label should be added to the resources `'backstage.io/kubernetes-id': nationalparks-py`

A custom label selector can also be added which will then be used by Backstage to find the resources. The label selector takes precedence over the id annotation.
`'backstage.io/kubernetes-label-selector': 'app=my-app,component=front-end`

In order to use k8s plugin to get the resources for Topology view, the above annotations/labels should be in place.

Enable additional tab on the entity view page

   ```ts
   import { TopologyPage } from '@janus-idp/backstage-plugin-topology';

   const serviceEntityPage = (
     <EntityPageLayout>
       // ...
       <EntityLayout.Route path="/topology" title="Topology">
        <TopologyPage />
       </EntityLayout.Route>
     </EntityPageLayout>
   );
   ```

**Todo:**
- [x] Visualize resources of the Deployment kind as read-only
- [x] Visualize app group for deployments if present with the label `'[app.kubernetes.io/part-of]': <GROUP_NAME>`
- [x] Visualize a visual connecter between two deployments if present with annotation `'[app.openshift.io/connects-to]': '[{"apiVersion": <RESOURCE_APIVERSION>,"kind": <RESOURCE_KIND>,"name": <RESOURCE_NAME>}]'`
- [x] Light/Dark theme support
- [x] Integrate with the k8s plugin to communicate with the cluster
- [x] Added Empty state for topology if no resources exist
- [x] Clean up of code/add tests

### CODE REVIEW

File: plugins/topology/src/components/Topology/utils/resource-utils.ts [hunk 1]
```diff
@@ -0,0 +1,104 @@
+import * as _ from 'lodash';
+import { DeploymentModel, PodModel } from '../models';
+import { Alert, Alerts } from '../types/monitoring-types';
+import { OverviewItem } from '../types/resource';
+import { K8sModel, K8sResourceKind } from '../types/types';
+import { WORKLOAD_TYPES } from './topology-utils';
+
+export const MONITORABLE_KINDS = [
+  'Deployment',
+  'DeploymentConfig',
```
tumido, line 10: What happens if the plugin is unable to locate resource type on the cluster? Does it raise errors/exceptions? I haven't run the plugin against vanilla Kubernetes myself, I'm just curious. I'm asking on the basis that `DeploymentConfig`s are not Kubernetes resources.

tumido (commented):
I've left few suggestions inline.

I would also suggest to rewrite the use of `lodash` methods into native JS (like those `_.map`s, `_.size`s, `_.head`s, etc...), WDYT?

packages/app/src/components/catalog/EntityPage.tsx
```diff
@@ -192,6 +193,11 @@ const websiteEntityPage = (
     <EntityLayout.Route path="/docs" title="Docs">
       {techdocsContent}
     </EntityLayout.Route>
+
+    {/* TODO : Remove for testing Add entity for topology */}
+    <EntityLayout.Route path="/topology" title="Topology">
+      <TopologyPage />
+    </EntityLayout.Route>
```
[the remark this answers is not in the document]
invincibleJai: Thanks updated.

plugins/topology/src/components/Topology/types/topology-types.ts
```diff
@@ -0,0 +1,51 @@
+import { Model, NodeModel } from '@patternfly/react-topology';
+import { ExtPodKind } from './pods';
+import { OverviewItem } from './resource';
+import { K8sResourceKind, WatchK8sResults } from './types';
+
+export type TrafficData = {
+  nodes: KialiNode[];
```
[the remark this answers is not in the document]
invincibleJai: I have removed it(moved types from the console, and cleaned it up now.

plugins/topology/src/components/Topology/utils/resource-utils.ts
invincibleJai: yeah, have removed DeploymentConfig and this is the initial PR here we are looking for `'deployments', 'pods', 'services', 'ingresses', 'replicasets'` and node workload is formed for Deployment, stadalone Pod others resources are needed to form relationship i.e pods and replicasets to show deployment. There would be follow-up PRs to add a decorator for ingress on workload and sidebar which will open on click to show other associated resources, labels, annotations, etc

divyanshiGupta:
@invincibleJai On changing tabs and then coming back to topology, empty state is being shown first even though the node data is available. It might be because it takes time to load the data and then render the topology view. Instead of empty state in this case we can show a loading visualisation similar to ocp.


https://user-images.githubusercontent.com/20724543/219331295-cf8f13e4-b68f-4bd0-a619-34c4ff4d5ca2.mov

plugins/topology/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+This plugin will help with visualizing the workloads like Deployment, Replicasets, Pods powering any serivce on kubernets cluster.
+
+## Prerequisites
+
+1. Install and configure kubernetes plugin by following these guides:
+   https://backstage.io/docs/features/kubernetes/installation
+   https://backstage.io/docs/features/kubernetes/configuration
+2. Kubernetes plugin is properly configured and able to connect to the cluster via a `ServiceAccount`
+3. Following `ClusterRole` must be granted to ServiceAccount accessing the cluster: (k8s plugin by default tries to fetch the following resources and if access is not provided an error is shown)
```
tumido, line 11: I think we should mention this that this role is already granted if you have the Backstage Kubernetes Plugin properly configured (which is a prerequisite), and link to https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control

The current formulation gives impression that there is something different in the role itself when compared to the one required by kubernetes plugin. While that may become the case in future, I think we should rather formulate in a way that basically says: "Please grant the permissions that Backstage Kubernetes plugin requires + these additional ones" (once we have any additional ones).

plugins/topology/src/components/Graph/TopologyComponentFactory.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const TopologyComponentFactory: ComponentFactory = (
+  kind: ModelKind,
+  type: string,
+): React.ComponentType<{ element: GraphElement }> | undefined => {
+  if (kind === ModelKind.graph) {
+    return withPanZoom()(DefaultGraph);
+  }
+  switch (type) {
+    case 'workload':
```
tumido, line 24: This should be a const as well (it't used in [`data-transforms/data-transformer.ts`](https://github.com/janus-idp/backstage-plugins/pull/104/files#diff-0253d70e4353e47b4ecae9126d55b0479178517ec598e1861b889c9dd04cd53fR19) as well)

plugins/topology/src/const.ts
```diff
@@ -0,0 +1,20 @@
...
+  DEFAULT_GROUP_PAD,
+  DEFAULT_GROUP_PAD,
+  DEFAULT_GROUP_PAD + 20,
+  DEFAULT_GROUP_PAD,
+];
+
+export const TYPE_APPLICATION_GROUP = 'part-of';
+export const TYPE_CONNECTS_TO = 'connects-to';
+
+export const TYPE_TRAFFIC_CONNECTOR = 'traffic-connector';
```
tumido, line 20: This is not used.

plugins/topology/src/utils/topology-utils.ts
```diff
@@ -0,0 +1,134 @@
...
+    type,
+    resource,
+    resources: {
+      ...overviewItem,
+    },
+    data: {
+      kind: kindRef,
+      editURL: deploymentsAnnotations['app.openshift.io/edit-url'],
+      vcsURI: deploymentsAnnotations['app.openshift.io/vcs-uri'],
+      vcsRef: deploymentsAnnotations['app.openshift.io/vcs-ref'],
```
tumido, line 106: Can we document these annotations?

plugins/topology/src/utils/transform-utils.ts
```diff
@@ -0,0 +1,266 @@
...
+  type: string,
+  data: any,
+  nodeProps?: Omit<OdcNodeModel, 'type' | 'data' | 'children' | 'id' | 'label'>,
+  children?: string[],
+  resourceKind?: string,
+  shape?: NodeShape,
+): OdcNodeModel => {
+  const uid = resource.metadata?.uid;
+  const name = resource.metadata?.name;
+  const label = resource.metadata?.labels?.['app.openshift.io/instance'];
```
tumido, line 51: Shouldn't this be

```suggestion
  const label = resource.metadata?.labels?.['app.kubernetes.io/instance'];
```

At least that's what is used at [`utils/topology-utils.ts`](https://github.com/janus-idp/backstage-plugins/pull/104/files#diff-8eed6f305ecf78ad964f1776378d9f7c65b816225105701acbf82ff4773cc8a2R96) above.

Maybe the `app.kubernetes.io/instance` should be a proper global constant.

plugins/topology/src/utils/transform-utils.ts
```diff
@@ -0,0 +1,266 @@
...
+      graphModel.nodes,
+    );
+  }
+};
+
+export const edgesFromAnnotations = (
+  annotations: any,
+): (string | ConnectsToData)[] => {
+  let edges: (string | ConnectsToData)[] = [];
+  if (_.has(annotations, ['app.openshift.io/connects-to'])) {
```
tumido, line 199: `app.openshift.io/connects-to` should be a global constant.

tumido (changes requested):
I've just skimmed the code. It's a lot of content to digest for a thorough review.

I see quite some inconsistent constant usage, can we globalize that and maybe document some annotations that user can define but they are hidden in code?

See my comments inline

plugins/topology/README.md
```diff
@@ -0,0 +1,107 @@
+# Topology plugin for Backstage
+
+This plugin will help with visualizing the workloads like Deployment, Replicasets, Pods powering any serivce on kubernets cluster.
```
riginoommen, line 3:
```suggestion
This plugin will help with visualizing the workloads like Deployment, Replicaset and Pods powering any service on the Kubernetes cluster.
```

plugins/topology/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Prerequisites
+
+1. Install and configure kubernetes plugin by following these guides:
```
riginoommen, line 7:
```suggestion
1. Install and configure Kubernetes plugin by following these guides:
```

plugins/topology/src/hooks/useDeepCompareMemoize.ts
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+
+export const useDeepCompareMemoize = <T = any>(
+  value: T,
+  strinfigy?: boolean,
```
riginoommen, line 6: Looks like type for ```stringify```

plugins/topology/src/hooks/useDeepCompareMemoize.ts
```diff
@@ -0,0 +1,19 @@
...
+): T => {
+  const ref = React.useRef<T>();
+
+  if (
+    strinfigy
```
riginoommen, line 11: Looks like type for ```stringify```

plugins/topology/src/utils/pod-resource-utils.ts
```diff
@@ -0,0 +1,280 @@
...
+import { AllPodStatus } from '../components/Pods/pod';
+import { ReplicaSetModel } from '../models';
+import { PodControllerOverviewItem } from '../types/pods';
+import { K8sModel, K8sResourceKind } from '../types/types';
+
+// List of container status waiting reason values that we should call out as errors in project status rows.
+export const CONTAINER_WAITING_STATE_ERROR_REASONS = [
+  'CrashLoopBackOff',
+  'ErrImagePull',
+  'ImagePullBackOff',
```
riginoommen, line 11:
```suggestion
  'ImagePullBackOff',
  'InvalidImageName'
  'CreateContainerError'
```

plugins/topology/src/utils/pod-resource-utils.ts
```diff
@@ -0,0 +1,280 @@
+import * as _ from 'lodash';
```
riginoommen, line 1: I suggest importing the specific resources which is utilized in the source code.

```suggestion
import { get, isFinite, some } from 'lodash';
```

plugins/topology/src/utils/resource-utils.ts
```diff
@@ -0,0 +1,40 @@
+import * as _ from 'lodash';
```
riginoommen, line 1: I suggest importing the specific resources which is utilized in the source code.

```suggestion
import { isEmpty} from 'lodash';
```

plugins/topology/src/utils/resource-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+import { WORKLOAD_TYPES } from './topology-utils';
+
+export const validPod = (pod: K8sResourceKind) => {
+  const owners = pod?.metadata?.ownerReferences;
+  const phase = pod?.status?.phase;
+  return _.isEmpty(owners) && phase !== 'Succeeded' && phase !== 'Failed';
+};
+
+const isStandaloneJob = (job: K8sResourceKind) =>
+  !_.find(job.metadata?.ownerReferences, owner => owner.kind === 'CronJob');
```
riginoommen, line 13: Can we utilize the find method in the javascript instead of the library

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

plugins/topology/src/utils/resource-utils.ts
```diff
@@ -0,0 +1,40 @@
+import * as _ from 'lodash';
+import { OverviewItem } from '../types/topology-types';
+import { K8sResourceKind } from '../types/types';
+import { WORKLOAD_TYPES } from './topology-utils';
+
+export const validPod = (pod: K8sResourceKind) => {
+  const owners = pod?.metadata?.ownerReferences;
+  const phase = pod?.status?.phase;
+  return _.isEmpty(owners) && phase !== 'Succeeded' && phase !== 'Failed';
```
riginoommen, line 9: can we try utiizing the ```Object.keys()``` method to identify the empty object

plugins/topology/src/utils/topology-utils.ts
```diff
@@ -0,0 +1,134 @@
+import * as _ from 'lodash';
```
riginoommen, line 1:
```suggestion
import { flatten, get } from 'lodash';
```

plugins/topology/src/utils/transform-utils.ts
```diff
@@ -0,0 +1,266 @@
+import {
+  EdgeModel,
+  Model,
+  NodeModel,
+  NodeShape,
+} from '@patternfly/react-topology';
+import * as _ from 'lodash';
```
riginoommen, line 7:
```suggestion
import { get} from 'lodash';
```

riginoommen (changes requested):
I suggest to iupdate the imports and use native code for the better performance

plugins/topology/README.md
```diff
@@ -0,0 +1,107 @@
...
+   ```yaml
+   'app.openshift.io/connects-to': '[{"apiVersion": <RESOURCE_APIVERSION>,"kind": <RESOURCE_KIND>,"name": <RESOURCE_NAME>}]'
+   ```
+
+## Getting started
+
+1. Install the plugin
+
+   ```bash
+   yarn workspace app add @janus-idp/backstage-plugin-topology
```
riginoommen, line 78:
```suggestion
   yarn add --cwd packages/app @janus-idp/backstage-plugin-topology
```

plugins/topology/dev/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { Entity } from '@backstage/catalog-model';
+import { createDevApp } from '@backstage/dev-utils';
+import { topologyPlugin, TopologyPage } from '../src/plugin';
+import {
+  EntityKubernetesContent,
+  KubernetesApi,
+  kubernetesApiRef,
+} from '@backstage/plugin-kubernetes';
+import { TestApiProvider } from '@backstage/test-utils';
+import fixture1 from '../src/__fixtures__/1-deployments.json';
```
riginoommen, line 12: ```fixture1``` variable seems odd i suggest using a variable name which is little more understandable

plugins/topology/package.json
```diff
@@ -0,0 +1,66 @@
...
+  },
+  "devDependencies": {
+    "@backstage/cli": "^0.22.1",
+    "@backstage/core-app-api": "^1.4.0",
+    "@backstage/dev-utils": "1.0.11",
+    "@backstage/test-utils": "^1.2.4",
+    "@testing-library/jest-dom": "^5.10.1",
+    "@testing-library/react": "^12.1.3",
+    "@testing-library/user-event": "^14.0.0",
+    "@types/node": "*",
```
riginoommen, line 59: Can we mention a version and lock it based on the compatiblity

plugins/topology/src/components/Graph/WorkloadNode.tsx
```diff
@@ -0,0 +1,194 @@
...
+      resource.kind,
+    );
+    const donutStatus = loaded && !loadError ? podData : null;
+    const { width, height } = element.getDimensions();
+    const workloadData = element.getData().data;
+    const [hover, hoverRef] = useHover();
+    const size = Math.min(width, height);
+    const { radius, decoratorRadius } = calculateRadius(size);
+    const cx = width / 2;
+    const cy = height / 2;
```
riginoommen, line 141:
```suggestion
    const cx = width / 2;
    const cy = height / 2;
```

can you please use a variable which is more readable

plugins/topology/src/components/Pods/pod.ts
```diff
@@ -0,0 +1,52 @@
+import { global_palette_purple_300 as globalPurple300 } from '@patternfly/react-tokens/dist/js/global_palette_purple_300';
+import { global_palette_white as globalWhite } from '@patternfly/react-tokens/dist/js/global_palette_white';
+import { global_warning_color_100 as globalWarning100 } from '@patternfly/react-tokens/dist/js/global_warning_color_100';
+
+export enum AllPodStatus {
```
riginoommen, line 5:
```suggestion
export enum AllPodStatus {
```

suggest to make the enums in a common format like ALL_POD_STATUS

plugins/topology/src/const.ts
```diff
@@ -0,0 +1,20 @@
+export const DEFAULT_NODE_PAD = 20;
```
riginoommen, line 1: Differentiating the ENUm declaration standard and constant declaration standard helps to make code more readable

plugins/topology/src/hooks/useForceUpdate.ts
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export const useForceUpdate = () => {
+  const [, setTick] = React.useState(0);
```
riginoommen, line 4:
```suggestion
  const [ setTick] = React.useState(0);
```

riginoommen (changes requested):

plugins/topology/src/hooks/useForceUpdate.ts
invincibleJai: am not sure if the above suggestion will work as `useState` returns a tuple with the first position will return an item and the second setter for it

plugins/topology/src/utils/transform-utils.ts
invincibleJai: Thanks, am working on refactoring will replace lodash usage with native methods wherever applicable and prefer named import over all as suggeted

plugins/topology/src/const.ts
invincibleJai: sorry if I get it right all enums I need to update to PascalCase i.e UpperCamelCase and constants should be as it's here or ? 

https://www.typescriptlang.org/docs/handbook/enums.html#:~:text=Enums%20are%20one%20of%20the,numeric%20and%20string%2Dbased%20enums.

spadgett:
@invincibleJai How much of this code was taken from the OpenShift console topology? Should we think about publishing common code as a package to avoid duplication?

invincibleJai:
> @invincibleJai How much of this code was taken from the OpenShift console topology? Should we think about publishing common code as a package to avoid duplication?

@spadgett graph shapes like WorkloadNode, BaseNode, and PodSet have been used and associated utils and types have been pulled from OpenShift console but that also involved a lot of refactoring as Console code has a dependency on a lot of internal / shared packages (some interdependency with devConsole package as well).

In backstage plugin, the initial version is read-only and should work with any k8s cluster.

Moving it to a separate package needs to be evaluated and this will be a lot more work and needs to discuss the strategy for updates as well.

plugins/topology/README.md
tumido: Is there a reason why this should be preferred? Can you please elaborate?

plugins/topology/src/components/Pods/pod.ts
invincibleJai: shouldn't enums be in PascalCase i.e UpperCamelCase , it was not consistent across, and have updated those

https://www.typescriptlang.org/docs/handbook/enums.html#:~:text=Enums%20are%20one%20of%20the,numeric%20and%20string%2Dbased%20enums.

plugins/topology/README.md
riginoommen: For most of the upstream packages for the backstage this format is commonly used to make sync on this I suggested

invincibleJai:
> @invincibleJai On changing tabs and then coming back to topology, empty state is being shown first even though the node data is available. It might be because it takes time to load the data and then render the topology view. Instead of empty state in this case we can show a loading visualisation similar to ocp.

Thanks @divyanshiGupta, have fixed this.

plugins/topology/src/utils/topology-utils.ts
invincibleJai: Have removed this now. These annotations are in the console we had for showing the repo and on click user can directly go to that but we'll bring this as a decorator for workload like one for ingress.

plugins/topology/src/types/types.ts
```diff
@@ -0,0 +1,133 @@
...
+// Properties common to (almost) all Kubernetes resources.
+type K8sResourceCommon = {
+  apiVersion?: string;
+  kind?: string;
+  metadata?: ObjectMetadata;
+};
+
+// Generic, unknown kind. Avoid when possible since it allows any key in spec
+// or status, weakening type checking.
+export type K8sResourceKind = K8sResourceCommon & {
```
spadgett, line 51: We're trying to slowly remove this from OpenShift console, so would be good to avoid using it here if we can. It's the `[key: string]: any` which makes the type checking much weaker even if you extend this type because you don't catch any typos in property names.

plugins/topology/src/utils/topology-utils.ts
```diff
@@ -0,0 +1,116 @@
...
+export const getTopologyResourceObject = (
+  topologyObject: TopologyDataObject,
+): K8sResourceKind | null => {
+  if (!topologyObject) {
+    return null;
+  }
+  return topologyObject.resource || topologyObject.resources?.obj;
+};
+
+const isGroupVersionKind = (ref: string) => ref.split('~').length === 3;
```
spadgett, line 23: How are the `~` delimited reference strings used in topology? This is another OpenShift console-ism that it might be good to remove unless we need it.

spadgett (commented):
Thanks, @invincibleJai. Fantastic progress. I know this is a huge effort.

I'm not super familiar with topology, but I believe @jeff-phillips-18 has looked as well?

I think we should prioritize writing some tests. If we don't hold the initial PR on it, at least after it goes in.

plugins/topology/README.md
```diff
@@ -0,0 +1,67 @@
+# Topology plugin for Backstage
+
+This plugin will help with visualizing the workloads like Deployment, Replicaset and Pods powering any service on the Kubernetes cluster.
+
+## Prerequisites
+
+1. Install and configure Kubernetes plugin by following these guides:
+   https://backstage.io/docs/features/kubernetes/installation
+   https://backstage.io/docs/features/kubernetes/configuration
+2. Kubernetes plugin is properly configured and able to connect to the cluster via a `ServiceAccount`
```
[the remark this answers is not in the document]
invincibleJai: It should be any supported authentication mechanism as provided here https://backstage.io/docs/features/kubernetes/configuration 

and has cluster roles for https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control 

we have added `ServiceAccount` as that's what we have tested. Once we verify other mechanisms will update.

@divyanshiGupta ^^

plugins/topology/src/components/Topology/TopologyComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const htmlTagElement = document.documentElement;
+    if (type === THEME_DARK) {
+      htmlTagElement.classList.add(THEME_DARK_CLASS);
+    } else {
+      htmlTagElement.classList.remove(THEME_DARK_CLASS);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [type]);
+
+  const watchedResources = [
```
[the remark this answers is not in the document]
invincibleJai: Yeah, supported default types from k8s plugin is listed here https://backstage.io/docs/features/kubernetes/configuration#objecttypes-optional

Yeah eventually we would need to bring in support for custom resources as well

plugins/topology/src/components/Graph/WorkloadNode.tsx
```diff
@@ -0,0 +1,194 @@
...
+    'Info',
+    'Unknown',
+    'PipelineNotStarted',
+  ],
+};
+
+const getNodePodStatus = (podStatus: AllPodStatus): number => {
+  switch (podStatus) {
+    case AllPodStatus.Failed:
+    case AllPodStatus.CrashLoopBackOff:
```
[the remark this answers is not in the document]
invincibleJai: Have added `ErrImagePull` as well here. 

image


In podStatus for topology I don't see this being handled in the console as there in topology, error in podrings will be shown only pod status is failed and for errImagePull status will be in Pending

```
...
status:
  phase: Pending
  ...
  containerStatuses:
    - name: container
      state:
        waiting:
          reason: ImagePullBackOff
          message: Back-off pulling image "jai/helloworld56"

```
and the best reasoning I can think of is there we have git import flows where we do s2i or pipeline builds so until the build is complete the image won't be accessible so the user will see a pending state over failed.

plugins/topology/src/components/Topology/TopologyWorkloadView.tsx
```diff
@@ -0,0 +1,113 @@
...
+        graph: {
+          id: 'g1',
+          type: 'graph',
+          layout,
+        },
+        ...dataModel,
+      };
+      controller.fromModel(model, false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
invincibleJai: not really, didn't observe any updated with `controller` in the dependency array.

plugins/topology/src/models.ts
```diff
@@ -0,0 +1,54 @@
+import { K8sModel } from './types/types';
+
+export const ReplicaSetModel: K8sModel = {
```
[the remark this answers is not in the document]
invincibleJai: Have updated it to use group/version/kind and later if we feel the need can add more props like custom Abbr/color if needed for badges in topology.

plugins/topology/src/types/types.ts
invincibleJai: so shall I add strict types for supported resources over generic ones like K8sDeploymentKind, K8sPodKind, etc?

plugins/topology/src/components/Graph/BaseNode.tsx
```diff
@@ -0,0 +1,126 @@
...
+  onShowCreateConnector,
+  onContextMenu,
+  contextMenuOpen,
+  alertVariant,
+  ...rest
+}) => {
+  const [hover, internalHoverRef] = useHover();
+  const nodeHoverRefs = useCombineRefs(
+    internalHoverRef,
+    hoverRef as React.Ref<Element>,
```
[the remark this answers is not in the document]
jeff-phillips-18: There are some know type issues in the PF topology package. These are planned to be fixed in the next major release.

plugins/topology/src/components/Graph/BaseNode.tsx
```diff
@@ -0,0 +1,126 @@
...
+  badgeBorderColor?: string;
+  badgeClassName?: string;
+  badgeLocation?: BadgeLocation;
+  children?: React.ReactNode;
+  attachments?: React.ReactNode;
+  element: Node;
+  hoverRef?: (node: Element) => () => void;
+  dragging?: boolean;
+  dropTarget?: boolean;
+  canDrop?: boolean;
```
jeff-phillips-18, line 41: Don't need these.

plugins/topology/src/components/Graph/BaseNode.tsx
```diff
@@ -0,0 +1,126 @@
...
+  nodeStatus?: NodeStatus;
+  showStatusBackground?: boolean;
+  alertVariant?: NodeStatus;
+} & Partial<WithSelectionProps> &
+  Partial<WithDragNodeProps> &
+  Partial<WithDndDropProps> &
+  Partial<WithContextMenuProps> &
+  Partial<WithCreateConnectorProps>;
```
jeff-phillips-18, line 49:
```suggestion
  Partial<WithContextMenuProps>;
```
We don't need these props if we aren't allowing edge modifications.

plugins/topology/src/types/types.ts
invincibleJai: i checked backstage-k8s plugin  and it fetches native types from https://www.npmjs.com/package/@kubernetes/client-node , have used the same to be more strict

https://github.com/backstage/backstage/blob/master/plugins/kubernetes/package.json#L43
https://github.com/backstage/backstage/blob/master/plugins/kubernetes/src/components/CronJobsAccordions/CronJobsAccordions.tsx#L25

invincibleJai:
> Thanks, @invincibleJai. Fantastic progress. I know this is a huge effort.
> 
> I'm not super familiar with topology, but I believe @jeff-phillips-18 has looked as well?
> 
> I think we should prioritize writing some tests. If we don't hold the initial PR on it, at least after it goes in.

Thanks @spadgett for reviewing, have addressed most of the review comments and have added tests as well for some of the components (will continue adding more and it can be a follow-up PR as well).

plugins/topology/src/components/Graph/WorkloadNode.tsx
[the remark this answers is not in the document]
spadgett: There are a number of ways this can occur (typo in image name, private image that can't be pulled, image registry is down, etc). We probably want to handle `ImagePullBackOff` as well.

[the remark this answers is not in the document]
spadgett: Yeah, I guess I can see the argument that the build hasn't completed yet, though, in which case we want pending. Maybe we should check to see if it wasn't intentional.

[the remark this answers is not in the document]
invincibleJai: currently, it'll show pending only if phase is pending with the exception to `CrashLoopBackOff` ,  `ImagePullBackOff ` or state is terminated with `exitCode != 0` in containerStatuses state

 https://github.com/janus-idp/backstage-plugins/pull/104/files#diff-8da95685163a1fdc3228145eccfedce06be50826d3d5f2e64d135106c5a02b21R65-R76

spadgett (commented):
Overall, looks good. Just a few small comments.

Thank you for the tests!

plugins/topology/src/models.ts
```diff
@@ -0,0 +1,37 @@
...
+  apiGroup: 'apps',
+  kind: 'Deployment',
+};
+
+export const ServiceModel: K8sModel = {
+  apiVersion: 'v1',
+  kind: 'Service',
+};
+
+export enum ModelsPlural {
```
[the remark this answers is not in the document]
invincibleJai: updated model to just be `GroupVersionKind` so have kept it separate.

plugins/topology/src/types/types.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export type K8sResourceKind = V1Deployment | V1Pod | V1Service | V1ReplicaSet;
+
+type ContainerStateValue = {
+  reason?: string;
+  [key: string]: any;
+};
+
+type ContainerState = {
```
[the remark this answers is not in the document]
invincibleJai: Thanks 👍 

yeah we do have `V1ContainerStatus` itself, updated to use those.

plugins/topology/src/models.ts
```diff
@@ -0,0 +1,37 @@
+import { GroupVersionKind } from './types/types';
+
+export const ReplicaSetModel: GroupVersionKind = {
```
spadgett, line 3: nitpick: maybe `ReplicaSetGroupVersionKind` rather than `Model`.

spadgett (commented):
LGTM, just two small naming nits

plugins/topology/src/types/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  V1Service,
+} from '@kubernetes/client-node';
+
+export type GroupVersionKind = {
+  kind: string;
+  apiVersion: string;
+  apiGroup?: string;
+};
+
+export type K8sResourceKind = V1Deployment | V1Pod | V1Service | V1ReplicaSet;
```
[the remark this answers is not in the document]
invincibleJai: will call it `K8sWorkloadResource` as these also include service and ingress in next PR which may not be workload but workload resources

invincibleJai:
Thanks @spadgett , @jeff-phillips-18 for the review 👍

invincibleJai:
@tumido PTAL when you get time, thanks!!

plugins/topology/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@janus-idp/backstage-plugin-topology",
+  "version": "0.1.0",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "license": "Apache-2.0",
+  "private": true,
```
tumido, line 7:
```suggestion
```

Please make the package public so we can release it.

tumido (changes requested):

plugins/topology/package.json
invincibleJai: Thanks Tom, updated

tumido (approved):
/lgtm
/approve

This is a huge PR. kudos to everyone working on it and reviewing it. :slightly_smiling_face: :100: Good job!

@spadgett can you please take one last look? If you approve we can merge and release it :slightly_smiling_face:

spadgett (approved):
LGTM 👍

Issue #119, opened 2023-02-13
🔌 Plugin: 3scale backstage provider


fmenesesg:
### 🔖 Summary

The [Custom entity provider](https://backstage.io/docs/features/software-catalog/external-integrations#custom-entity-providers) should be able to synchronize 3scale Products as API Entity on the Backstage Catalog.

### 🌐 Project website (if applicable)

https://www.3scale.net/
https://github.com/3scale

### ✌️ Context

It will be a [Custom entity provider](https://backstage.io/docs/features/software-catalog/external-integrations#custom-entity-providers) that will synchronize the products defined on the Tenant.
The Product to be synchronized should have an API doc related to it in order to be imported.
The configuration of delay time between sync waves should be configurable at app-config.yaml.

### 👀 Have you spent some time to check if this plugin request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/backstage/backstage/blob/master/CODE_OF_CONDUCT.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #120, opened 2023-02-13
Add 3scale backstage provider

fmenesesg:
#119 
This is the first version of 3scale backstage provider that syncs 3scale Products as API entities on the Backstage Catalog


plugins/3scale-provider/src/providers/ThreeScaleApiEntityProvider.ts
```diff
@@ -0,0 +1,229 @@
...
+      page++;
+    }
+
+    this.logger.info(`Applying the mutation with ${entities.length} entities`);
+
+    await this.connection.applyMutation({
+      type: 'full',
+      entities: entities.map(entity => ({
+        entity,
+        locationKey: 'ThreeScaleApiEntityProvider',
```
tumido, line 172: This should at least match the `getProviderName`. In case you have multiple instances of this provider (for different environments for example) and more than one of them emit an entity with the same kind/name/namespace, it will cause conflicts in catalog database.

See: https://backstage.io/docs/features/software-catalog/external-integrations#provider-mutations

plugins/3scale-provider/src/providers/ThreeScaleApiEntityProvider.ts
```diff
@@ -0,0 +1,229 @@
...
+          service.service.id,
+        );
+        if (apiDoc !== null) {
+          const apiEntity: ApiEntity = this.buildApiEntityFromService(
+            service,
+            apiDoc!,
+            proxy,
+          );
+          entities.push(apiEntity);
+          this.logger.info(`Discovered ApiEntity ${service.service.name}`);
```
tumido, line 153: Please change to debug or remove. We don't need to know about every ingested entity by default.

plugins/3scale-provider/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "dependencies": {
+    "@backstage/backend-common": "^0.17.0",
+    "@backstage/backend-tasks": "^0.4.0",
+    "@backstage/catalog-model": "^1.1.4",
+    "@backstage/config": "^1.0.5",
+    "@backstage/plugin-catalog-backend": "^1.6.0",
+    "@types/express": "*",
+    "express": "^4.17.3",
+    "express-promise-router": "^4.1.0",
+    "js-base64": "^3.7.3",
```
tumido, line 33: I don't see this used anywhere, am I missing something?

plugins/3scale-provider/src/providers/ThreeScaleApiEntityProvider.ts
```diff
@@ -0,0 +1,229 @@
...
+    return {
+      kind: 'API',
+      apiVersion: 'backstage.io/v1alpha1',
+      metadata: {
+        annotations: {
+          [ANNOTATION_LOCATION]: location,
+          [ANNOTATION_ORIGIN_LOCATION]: location,
+        },
+        //  TODO: add tenant name
+        name: `${service.service.system_name}}`.replaceAll('}', ''),
```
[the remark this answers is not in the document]
fmenesesg: This was a mistake, I already fix this on another commit

plugins/3scale-provider/src/providers/ThreeScaleApiEntityProvider.ts
```diff
@@ -0,0 +1,229 @@
...
+      );
+      apiDocs = await listApiDocs(this.baseUrl, this.accessToken);
+      for (const element of services.services) {
+        const service = element;
+        this.logger.debug(`Find service ${service.service.name}`);
+
+        // Trying to find the API Doc for the service.
+        const apiDoc = apiDocs.api_docs.find(obj => {
+          return obj.api_doc.service_id === service.service.id;
+        });
```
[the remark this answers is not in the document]
fmenesesg: I changed this a little to validate if api_doc_service_id is undefined, this is because it is possible to have an Active Doc not associated with a product on 3scale.

fmenesesg:
@tumido I apply all the changes suggested and some other fixes that I found, can you PTAL.

plugins/3scale-backend/src/providers/config.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  for (const id of providerConfigs.keys()) {
+    configs.push(
+      readThreeScaleApiEntityConfig(id, providerConfigs.getConfig(id)),
+    );
+  }
+
+  return configs;
+}
```
tumido, line 25: It may be more efficient to do it like this:

```suggestion
  const providerConfigs = config.getOptionalConfig(
    'catalog.providers.threeScaleApiEntity',
  );

  if (!providerConfigs) {
    return [];
  }

  return providerConfigs.keys().map(id => readThreeScaleApiEntityConfig(id, providerConfigs.getConfig(id))))
}
```

Array `.push()` is many times slower than `.map()`.

tumido (changes requested):
I don't think you want to be removing the `LICENSE` at the repository root as well. Please restore that file.

Otherwise only a nit, slightly effecting performance. But nothing important. It looks good to me otherwise.

fmenesesg:
@tumido all changes were applied.

tumido (changes requested):
image


Can we undo this change, please?

fmenesesg:
Done, sorry about that 

> image
> 
> Can we undo this change, please?

plugins/3scale-backend/config.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  systemLabel?: string;
+  ownerLabel?: string;
+  addLabels?: boolean;
+  schedule?: TaskScheduleDefinitionConfig;
+}
+
+export interface Config {
+  catalog?: {
+    providers?: {
+      threeScaleApiEntity?: Record<string, ThreeScaleConfig>;
```
tumido, line 16: I've played around with this and for some reason `Record<...>` is not generating a validating JSON schema. Current code generates following definition:


```
$schema: https://backstage.io/schema/config-v1
title: Application Configuration Schema
type: object
definitions:
  "Record<string,ThreeScaleConfig>":
    type: object
...
  catalog:
    type: object
    description: Configuration options for the catalog plugin.
    properties:
      providers:
        type: object
        properties:
          threeScaleApiEntity:
            $ref: "#/definitions/Record<string,ThreeScaleConfig>"
```

Compared to `{ [key: string]: ThreeScaleConfig}`:

```
$schema: https://backstage.io/schema/config-v1
title: Application Configuration Schema
type: object
definitions:
  ThreeScaleConfig:
    type: object
    required:
      - baseUrl
      - serviceAccount
      - serviceAccountCredentials
    properties:
      baseUrl:
        type: string
      serviceAccount:
        type: string
      serviceAccountCredentials:
        type: string
      systemLabel:
        type: string
      ownerLabel:
        type: string
      addLabels:
        type: boolean
      schedule: {}
...
  catalog:
    type: object
    description: Configuration options for the catalog plugin.
    properties:
      providers:
        type: object
        properties:
          threeScaleApiEntity:
            type: object
            additionalProperties:
              $ref: "#/definitions/ThreeScaleConfig"
```

Therefore I would suggest to change this to:

```suggestion
      threeScaleApiEntity?: {[key: string]: ThreeScaleConfig};
```

You can check the generated schema via `npx backstage-cli config:schema`

tumido (changes requested):
I've noticed one thing when working on another PR. Sorry for late addition to the review.

plugins/3scale-backend/config.d.ts
tumido: Based on my investigation this tracks back to https://github.com/YousefED/typescript-json-schema/issues/337 bug.

fmenesesg: Done, This is great I'm learning a lot from your reviews. Thanks

tumido (approved):
LGTM :+1: :100: :heart:

PR #144, opened 2023-02-22
feat: Add cluster ID annotation constant

mimotej:
SSIA


mimotej:
Moved to https://github.com/janus-idp/backstage-plugins/pull/143

PR #145, opened 2023-02-22
feat(ocm)!: Relocate OCM config and support multiple hubs

tumido:
BREAKING CHANGE :boom: 

This PR updates OCM plugin to respect/conform better to non-written standards in catalog provider plugins and moves the configuration from top level `ocm` key into `catalog.providers.ocm`. It now supports multiple hub clusters

New configuration looks like this:

```yaml
catalog:
  providers:
    ocm:
      environment1:
        kubernetesPluginRef: this_is_sourced_from_kuberentes_plugin
      environment2:
        name: how_you_want_the_cluster_to_be_named_in_backstage
        url: api_url
        serviceAccountToken: token
        ...
```

This is also addressed in router that changes from `/status/:clusterName` to `/status/:providerId/:clusterName`.

Updated backend test-suite to use fixtures instead of inline objects.


plugins/ocm-backend/config.d.ts
```diff
@@ -0,0 +1,40 @@
...
+  kubernetesPluginRef: string;
+}
+
+interface HubClusterConfig {
+  /**
+   * Url of the hub cluster API endpoint
+   */
+  url: string;
+  /**
+   * Name of the cluster
```
SamoKopecky, line 14: I would change this to a something more descriptive, maybe the same description as in the README example configuration.

plugins/ocm-backend/config.d.ts
```diff
@@ -0,0 +1,40 @@
...
+   */
+  name: string;
+  /**
+   * Service Account Token which is used for querying data from the hub
+   * @visibility secret
+   */
+  serviceAccountToken?: string;
```
SamoKopecky, line 21: Doesn't this have to be required?

plugins/ocm-backend/src/helpers/config.ts
```diff
@@ -1,82 +1,84 @@
...
-    return getHubClusterFromOcmConfig(config);
-  }
-  if (path === CLUSTER_CONFIG_PATH) {
-    logger.info(`Loading config from ${CLUSTER_CONFIG_PATH} config`);
-    return getHubClusterFromKubernetesConfig(config);
+  return {
+    id,
+    url: hub.getString('url'),
+    hubResourceName: hub.getString('name'),
+    serviceAccountToken: hub.getOptionalString('serviceAccountToken'),
```
SamoKopecky, line 68: If the `serviceAccountToken` is made required in the config, this can change to `hub.getString()`.

plugins/ocm-backend/src/helpers/config.ts
```diff
@@ -1,82 +1,84 @@
...
 };
 
-export const getHubClusterFromOcmConfig = (config: Config): Config => {
-  const hub = config.getConfig(HUB_CONFIG_PATH);
+export const getHubClusterFromOcmConfig = (
+  id: string,
+  config: Config,
+): Config => {
   // Check if required values are valid
   const requiredValues = ['name', 'url'];
```
SamoKopecky, line 44: The `serviceAccountToken` has to also be added to the `requiredValues` if it is made required in the config.

plugins/ocm-backend/src/types.ts
```diff
@@ -0,0 +1,93 @@
+import { KubernetesObject, V1PodCondition } from '@kubernetes/client-node';
+
+export type OcmConfig = {
+  id: string;
+  url: string;
+  hubResourceName: string;
+  serviceAccountToken?: string;
```
SamoKopecky, line 7: Same here if this is required in config.

plugins/ocm/src/helpers/apiClient.ts
```diff
@@ -2,24 +2,22 @@ import { ConfigApi } from '@backstage/core-plugin-api';
...
-): Promise<any> => {
+const clusterApiFetchCall = async (configApi: ConfigApi, params?: string) => {
   const backendUrl = configApi.getString('backend.baseUrl');
-  const jsonResponse = fetch(`${backendUrl}/api/ocm/status${params}`).then(r =>
-    r.json(),
+  const jsonResponse = await fetch(
+    `${backendUrl}/api/ocm/status${params || ''}`,
   );
-  return jsonResponse;
+  return jsonResponse.json();
```
SamoKopecky, line 10: I think this can be put on a single line.

SamoKopecky (commented):
Looks really good to me, just some minor suggestions bellow.

plugins/ocm-backend/config.d.ts
tumido: It's not required. If not defined the default context from `kubeconfig` is used. This is consistent with kubernetes plugin for cluster with `authProvider: 'serviceAccount'` and without `serviceAccountToken` defined.

plugins/ocm/src/helpers/apiClient.ts
tumido: It can but it would look cluttered.

```ts
return (await fetch(`${backendUrl}/api/ocm/status${params || ''}`)).json()
```

plugins/ocm-backend/config.d.ts
SamoKopecky: That is true totally forgot about that.

SamoKopecky (approved):
LGTM! :rocket:

SamoKopecky:
I've noticed one small thing, the README has to be updated to iterate over all of the providers here: https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm/README.md?plain=1#L155-L157 since cluster provider now returns a collection of providers.

SamoKopecky:
Another thing I noticed while testing this is that the discovered clusters in the catalog contain the hub cluster still named as `local-cluster` instead of the configured id or name. See images below and the API response:
```json
[
...
  {
    "name": "local-cluster",
    "status": {
      "available": true,
      "reason": "Managed cluster is available"
    }
  },
...
]
```

image
image

plugins/ocm/src/helpers/apiClient.ts
SamoKopecky: Yea I guess the original way is more readable.

tumido:
/hold

@SamoKopecky's review is valid. I need to address few things.

plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
```diff
@@ -84,25 +95,29 @@ export class ManagedClusterProvider implements EntityProvider {
...
+          : i.metadata!.name!;
+
       return {
         kind: 'Resource',
         apiVersion: 'backstage.io/v1beta1',
         metadata: {
-          name: i.metadata.name,
-          title: i.metadata?.labels?.name,
+          name: normalizedName,
+          title: i.metadata?.labels?.name || normalizedName,
```
SamoKopecky, line 110:
```suggestion
          title: normalizedName || i.metadata?.labels?.name,
```
This is probably the the reason why the hub cluster is named `local-cluster` in the catalog.

SamoKopecky (commented):
This may be the issue for the wrong name in the catalog.

SamoKopecky (approved):
LGTM!

riginoommen (approved):

mimotej (approved):

Issue #121, opened 2023-02-14
App Topology: Add filter resources by cluster functionality in Topology view


divyanshiGupta:
**Description:**
Currently only the first cluster resources are being visualised in topology but as a user I would like to be able to filter resources by cluster in case of multiple clusters.

**Acceptance Criteria:**

- [x] Add a dropdown cluster filter in the toolbar
- [x] On selecting a cluster from the filter only the resources of that cluster should be shown
- [x] Add a toolbar in topology view
- [x] Show error (if any while fetching clusters/resources) similar to k8s plugin

PR #146, opened 2023-02-22
feat(topology): Add filter resources by cluster functionality in Topology plugin

divyanshiGupta:
**Story:** https://github.com/janus-idp/backstage-plugins/issues/121

**Description:**
This PR adds a toolbar in topology view consisting of a filter by cluster dropdown which lists all the available clusters and the user can select the cluster they want to see the resources for. It also shows any errors while fetching the resources in the topology view.

https://user-images.githubusercontent.com/20724543/224062140-6bc1ee4b-794e-4455-9e16-bb24542dbd21.mov

**Test Setup:**
Install and configure kubernetes plugin by following these guides:

https://backstage.io/docs/features/kubernetes/installation
https://backstage.io/docs/features/kubernetes/configuration

Kubernetes plugin is properly configured and able to connect to the cluster via a ServiceAccount

Following ClusterRole must be granted to ServiceAccount accessing the cluster:
(k8s plugin by default tries to fetch the following resources and if access is not provided an error is shown)

```
kind: ClusterRole
metadata:
  name: backstage-read-only
rules:
  - apiGroups:
      - '*'
    resources:
      - pods
      - configmaps
      - services
      - deployments
      - replicasets
      - horizontalpodautoscalers
      - ingresses
      - statefulsets
      - limitranges
      - daemonsets
    verbs:
      - get
      - list
      - watch
  - apiGroups:
      - batch
    resources:
      - jobs
      - cronjobs
    verbs:
      - get
      - list
      - watch
  - apiGroups:
      - metrics.k8s.io
    resources:
      - pods
    verbs:
      - get
      - list
 ```
 
Create a service component entity - https://github.com/temp-demo-organisation/nationalparks-py/blob/master/catalog-info2.yaml

To get the resources from a k8s cluster do the following:

K8s plugin identifies if the provided entity has k8s resources and if yes from which namespace it should get the resources based on the following annotations added to the entity's catalog-info.yaml
```
annotations: 
 backstage.io/kubernetes-id: nationalparks-py
 backstage.io/kubernetes-namespace: div
```

For k8s plugin to get the desired entity's k8s resources the following label should be added to the resources `'backstage.io/kubernetes-id': nationalparks-py`

A custom label selector can also be added which will then be used by Backstage to find the resources. The label selector takes precedence over the id annotation.
`'backstage.io/kubernetes-label-selector': 'app=my-app,component=front-end`

In order to use k8s plugin to get the resources for Topology view, the above annotations/labels should be in place.

Enable additional tab on the entity view page

   ```ts
   import { TopologyPage } from '@janus-idp/backstage-plugin-topology';

   const serviceEntityPage = (
     <EntityPageLayout>
       // ...
       <EntityLayout.Route path="/topology" title="Topology">
        <TopologyPage />
       </EntityLayout.Route>
     </EntityPageLayout>
   );
   ```

**To Do:**
- [x] Add a dropdown cluster filter in the toolbar
- [x] On selecting a cluster from the filter only the resources of that cluster should be shown
- [x] Add a toolbar in topology view
- [x] Show error (if any while fetching clusters/resources) similar to k8s plugin

### CODE REVIEW

**Comment by divyanshiGupta** [comment]
cc: @invincibleJai

File: plugins/topology/src/components/Topology/TopologyComponent.tsx [hunk 1]
```diff
@@ -4,18 +4,26 @@ import '@patternfly/react-core/dist/styles/base.css';
...
+import { useEntity } from '@backstage/plugin-catalog-react';
+import { useKubernetesObjects } from '@backstage/plugin-kubernetes';
+import { useK8sResourcesClusters } from '../../hooks/useK8sResourcesClusters';
 
 const THEME_DARK = 'dark';
 const THEME_DARK_CLASS = 'pf-theme-dark';
 
 export const TopologyComponent = () => {
+  const { entity } = useEntity();
+  const k8sObjectsResponse = useKubernetesObjects(entity);
```
invincibleJai, line 22: will be nice to move this to a custom hook which we would have for context, WDYT?

plugins/topology/src/components/Topology/TopologyComponent.tsx
```diff
@@ -34,13 +42,19 @@ export const TopologyComponent = () => {
...
+  const k8sResourcesClusters = useK8sResourcesClusters(k8sObjectsResponse);
 
   return (
-    <K8sResourcesContext.Provider value={k8sResponseData}>
-      <div className="pf-ri__topology">
-        <TopologyWorkloadView />
-      </div>
-    </K8sResourcesContext.Provider>
+    <K8sResourcesClustersContext.Provider value={k8sResourcesClusters.clusters}>
+      <K8sResourcesClusterContext.Provider value={clusterContext}>
```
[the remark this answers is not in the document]
divyanshiGupta: We can have both the values in one context but since both of them gets set at different places so just to be more elaborate kept them in different contexts as well.

plugins/topology/src/components/Topology/TopologyComponent.tsx
divyanshiGupta: So you are saying have a hook which then should be used here or in the hooks I want the `k8sObjectsResponse` response in? If it the later then the api call will be made more than once.

invincibleJai: i was thinking of just one context provider which can return what all need and consumer can pick the data as needed

packages/backend/src/plugins/kubernetes.ts
```diff
@@ -0,0 +1,16 @@
+import { KubernetesBuilder } from '@backstage/plugin-kubernetes-backend';
```
invincibleJai, line 1: we don't want to push this, right?

tumido:
Please change the commit messages (not the PR title, that doesn't work) according to: https://github.com/janus-idp/backstage-plugins#requirements

Otherwise the release process won't be able to generate a proper changelog and release.

invincibleJai:
Thanks @divyanshiGupta, lint and tests need to be fixed PR jobs are failing

divyanshiGupta:
> Thanks @divyanshiGupta, lint and tests need to be fixed PR jobs are failing

Yes, working on to fix those :)

invincibleJai:
@divyanshiGupta UI breaks for me in `TopologyViewWorkloadComponent.tsx`

k8sClusters is an empty list as below and we are trying to access `currentK8sCluster` (0) index

```
clusters: []
errors : []
```

```
const currentClusterErrors = [
    ...k8sClusters.errors[currentK8sCluster],
    ...(k8sResponseError ? [{ message: k8sResponseError }] : []),
  ];
```

image

invincibleJai:
@divyanshiGupta it flicker as well at regular interval

topologyFilterbar

divyanshiGupta:
> @divyanshiGupta it flicker as well at regular interval
> 
> topologyFilterbar 
> 
> It'll look good if we remove the padding below the filter bar
> 
> image
> 
> in the dark theme empty state doesn't look as expected and will be good to use backstage warning panel over pf here to show error https://backstage.io/docs/reference/core-components.warningpanel/
> 
> image

@invincibleJai Have fixed the above issues except for the dark theme not being applied to the empty state.


https://user-images.githubusercontent.com/20724543/224060781-ff15da99-5358-45c5-a3d0-10ee9e6d65f2.mov

plugins/topology/src/hooks/useK8sObjectsResponse.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  useEntity: () => ({
+    entity: {
+      metadata: {
+        name: 'test',
+      },
+    },
+  }),
+}));
+
+describe('useK8sObjectsResponse', () => {
```
invincibleJai, line 31: shall we add more tests here with `setSelectedCluster` and then check on the cluster if updated and one for `selectedClusterErrors` as well?

plugins/topology/src/components/Topology/TopologyToolbar.tsx
```diff
@@ -0,0 +1,68 @@
...
+  SelectOption,
+  SelectVariant,
+  SelectOptionObject,
+} from '@patternfly/react-core';
+import { K8sResourcesContext } from '../../hooks/K8sResourcesContext';
+
+const TopologyToolbar: React.FC<{
+  setClusterContext: React.Dispatch<React.SetStateAction<number>>;
+}> = ({ setClusterContext }) => {
+  const { clusters: k8sClusters } = React.useContext(K8sResourcesContext);
```
invincibleJai, line 14: do we need to take `setSelectedCluster` as props? or

```suggestion
const TopologyToolbar: React.FC = () => {
  const { clusters: k8sClusters , setSelectedCluster} = React.useContext(K8sResourcesContext);
```

plugins/topology/src/hooks/K8sResourcesContext.ts
```diff
@@ -1,4 +1,6 @@
 import * as React from 'react';
-import { K8sResponse } from '../types/topology-types';
+import { K8sResourcesContextData } from '../types/types';
 
-export const K8sResourcesContext = React.createContext<K8sResponse>({});
```
invincibleJai, line 4: Remove the type definition for `K8sResponse`, this is no longer used ?

plugins/topology/src/hooks/useK8sObjectsResponse.ts
```diff
@@ -0,0 +1,32 @@
...
+import { K8sResourcesContextData } from '../types/types';
+import { useAllWatchResources } from './useAllWatchResources';
+import { useK8sResourcesClusters } from './useK8sResourcesClusters';
+
+export const useK8sObjectsResponse = (
+  watchedResource: string[],
+): K8sResourcesContextData => {
+  const { entity } = useEntity();
+  const { kubernetesObjects, loading, error } = useKubernetesObjects(entity);
+  const [selectedCluster, setSelectedCluster] = useState<number>(0);
```
invincibleJai, line 13: we are setting the default to `0`. what will happen if there are no clusters?

divyanshiGupta: It doesnt matter. This will only be used when there are more than 1 available clusters. And by default we want 1st cluster to be selected.

invincibleJai (approved):

PR #157, opened 2023-03-01
feat(ocm): Add a default owner configuration field

### LINKED ISSUE #140: OCM: Set cluster owners
**Is your feature request related to a problem? Please describe.**
Currently the catalog entity that describes each managed cluster, has no `.spec.owner`.

**Describe the solution you'd like**
Allow user to define default owner in config. Optionally, we can consider introducing new annotation to `ManagedCluster` resources that can translate into this value.

**Describe alternatives you've considered**
OCM provides annotation `open-cluster-management.io/user-identity` on `ClusterClaim` and `ClusterDeployment` resources. This annotation appear to be either not present for imported clusters and set to `system:admin` for provisioned clusters. It seems to hold minimal value.

**Additional context**
Replaces https://github.com/operate-first/service-catalog/issues/178

SamoKopecky:
Fixes #140 

Add a configuration option for adding a default owner to the created entities from the cluster discovery. It works by using mimicking the backstage owner references described here: https://backstage.io/docs/features/software-catalog/references/.

### DIFF
--- a/plugins/ocm-backend/config.d.ts
+++ b/plugins/ocm-backend/config.d.ts
@@ -36,7 +36,12 @@ export interface Config {
         /**
          * Key is reflected as provider ID. Defines and claims plugin instance ownership of entities
          */
-        [key: string]: KubernetesPluginRef | HubClusterConfig;
+        [key: string]: (KubernetesPluginRef | HubClusterConfig) & {
+          /**
+           * Owner reference to created cluster entities in the catalog
+           */
+          owner?: string;
+        };
       };
     };
   };
--- a/plugins/ocm-backend/src/helpers/config.ts
+++ b/plugins/ocm-backend/src/helpers/config.ts
@@ -4,6 +4,7 @@ import { OcmConfig } from '../types';
 const KUBERNETES_PLUGIN_CONFIG = 'kubernetes.clusterLocatorMethods';
 const OCM_PREFIX = 'catalog.providers.ocm';
 const KUBERNETES_PLUGIN_KEY = 'kubernetesPluginRef';
+const OWNER_KEY = 'owner';
 
 const isValidUrl = (url: string): boolean => {
   try {
@@ -83,6 +84,7 @@ export const getHubClusterFromConfig = (
     serviceAccountToken: hub.getOptionalString('serviceAccountToken'),
     skipTLSVerify: hub.getOptionalBoolean('skipTLSVerify') || false,
     caData: hub.getOptionalString('caData'),
+    owner: config.getOptionalString(OWNER_KEY) || 'unknown',
   };
 };
 
--- a/plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
+++ b/plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
@@ -47,6 +47,7 @@ export class ManagedClusterProvider implements EntityProvider {
   protected readonly client: CustomObjectsApi;
   protected readonly hubResourceName: string;
   protected readonly id: string;
+  protected readonly owner: string;
   protected readonly logger: winston.Logger;
   protected connection?: EntityProviderConnection;
 
@@ -55,11 +56,13 @@ export class ManagedClusterProvider implements EntityProvider {
     hubResourceName: string,
     id: string,
     options: { logger: winston.Logger },
+    owner: string,
   ) {
     this.client = client;
     this.hubResourceName = hubResourceName;
     this.id = id;
     this.logger = options.logger;
+    this.owner = owner;
   }
 
   static fromConfig(config: Config, options: { logger: winston.Logger }) {
@@ -70,6 +73,7 @@ export class ManagedClusterProvider implements EntityProvider {
         provider.hubResourceName,
         provider.id,
         options,
+        provider.owner,
       );
     });
   }
@@ -138,7 +142,7 @@ export class ManagedClusterProvider implements EntityProvider {
           ],
         },
         spec: {
-          owner: 'unknown',
+          owner: this.owner,
           type: 'kubernetes-cluster',
         },
       };
--- a/plugins/ocm-backend/src/helpers/config.test.ts
+++ b/plugins/ocm-backend/src/helpers/config.test.ts
@@ -245,6 +245,7 @@ describe('getHubClusterFromConfig', () => {
     const ocmConfig = {
       name: 'foo',
       url: 'http://example.com',
+      owner: 'bar',
     };
 
     const config = new ConfigReader(ocmConfig);
@@ -262,11 +263,13 @@ describe('getHubClusterFromConfig', () => {
 
     expect(result.hubResourceName).toEqual('foo');
     expect(result.url).toEqual('http://example.com');
+    expect(result.owner).toEqual('bar');
   });
 
   it('should correctly parse the ocm hub from the config while using kubernetes ref', () => {
     const ocmConfig = {
       kubernetesPluginRef: 'cluster1',
+      owner: 'foo',
     };
     const globalConfig = new ConfigReader({
       kubernetes: {
@@ -297,6 +300,29 @@ describe('getHubClusterFromConfig', () => {
 
     expect(result.hubResourceName).toEqual('cluster1');
     expect(result.url).toEqual('http://example.com');
+    expect(result.owner).toEqual('foo');
+  });
+
+  it('should throw an error if the url is not valid while using ocm config', () => {
+    const ocmConfig = {
+      name: 'foo',
+      url: 'bar',
+    };
+
+    const config = new ConfigReader(ocmConfig);
+    const globalConfig = new ConfigReader({
+      catalog: {
+        providers: {
+          ocm: {
+            env: ocmConfig,
+          },
+        },
+      },
+    });
+
+    const result = () => getHubClusterFromConfig('env', config, globalConfig);
+
+    expect(result).toThrow('"bar" is not a valid url');
   });
 
   it('should throw an error if the url is not valid while using kubernetes ref', () => {
@@ -333,10 +359,10 @@ describe('getHubClusterFromConfig', () => {
     expect(result).toThrow('"bar" is not a valid url');
   });
 
-  it('should throw an error if the url is not valid while using ocm config', () => {
+  it('should fallback to the default owner if no owner is configured', () => {
     const ocmConfig = {
       name: 'foo',
-      url: 'bar',
+      url: 'http://example.com',
     };
 
     const config = new ConfigReader(ocmConfig);
@@ -350,8 +376,8 @@ describe('getHubClusterFromConfig', () => {
       },
     });
 
-    const result = () => getHubClusterFromConfig('env', config, globalConfig);
+    const result = getHubClusterFromConfig('env', config, globalConfig);
 
-    expect(result).toThrow('"bar" is not a valid url');
+    expect(result.owner).toEqual('unknown');
   });
 });
--- a/plugins/ocm-backend/src/types.ts
+++ b/plugins/ocm-backend/src/types.ts
@@ -7,6 +7,7 @@ export type OcmConfig = {
   serviceAccountToken?: string;
   skipTLSVerify?: boolean;
   caData?: string;
+  owner: string;
 };
 
 export interface ClusterClaim {
--- a/plugins/ocm/README.md
+++ b/plugins/ocm/README.md
@@ -168,6 +168,17 @@ This plugin is made up of 2 packages:
      }
    ```
 
+6. (Optional) The default owner for the cluster entities in the catalog for a specific env can additionally be configured. The configuration follows the [upstream string references documentation](https://backstage.io/docs/features/software-catalog/references/#string-references). For example to set `foo` as the owner for clusters from `env` in the catalog use:
+
+   ```yaml
+   # app-config.yaml
+   catalog:
+     providers:
+       ocm:
+         env:
+           owner: user:foo
+   ```
+
 ### Setup `@janus-idp/backstage-plugin-ocm`
 
 Install the frontend plugin:


tumido:
@SamoKopecky #145 is merged now, can you rebase please? 🙂

SamoKopecky:
Rebased to `main`.

plugins/ocm-backend/config.d.ts
```diff
@@ -29,14 +29,24 @@ interface HubClusterConfig {
...
+
 export interface Config {
   catalog?: {
     providers?: {
       ocm?: {
         /**
          * Key is reflected as provider ID. Defines and claims plugin instance ownership of entities
          */
-        [key: string]: KubernetesPluginRef | HubClusterConfig;
+        [key: string]: (KubernetesPluginRef | HubClusterConfig) | Owner;
```
tumido, line 49: This is not correct, following passes schema validation:


```
catalog:
  providers:
    ocm:
      foo:
        kind: User
        name: test
```

plugins/ocm-backend/config.d.ts
```diff
@@ -29,14 +29,24 @@ interface HubClusterConfig {
   caData?: string;
 }
 
+interface Owner {
+  owner:
+    | string
+    | {
+        kind?: string;
```
tumido, line 36: Can this really be any string?

plugins/ocm-backend/src/providers/ManagedClusterProvider.ts
```diff
@@ -48,18 +48,21 @@ export class ManagedClusterProvider implements EntityProvider {
...
     hubResourceName: string,
     id: string,
     options: { logger: winston.Logger },
+    config: Config,
   ) {
     this.client = client;
     this.hubResourceName = hubResourceName;
     this.id = id;
     this.logger = options.logger;
+    this.config = config;
```
tumido, line 65: Let's pass the owner directly here.

SamoKopecky: Will be fixed in the latest commit.

plugins/ocm-backend/config.d.ts
SamoKopecky: Thats true, I changed so that the schema looks like this now:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "Config": {
      "additionalProperties": false,
      "properties": {
        "catalog": {
          "additionalProperties": false,
          "properties": {
            "providers": {
              "additionalProperties": false,
              "properties": {
                "ocm": {
                  "additionalProperties": {
                    "anyOf": [
                      {
                        "additionalProperties": false,
                        "properties": {
                          "kubernetesPluginRef": {
                            "description": "Match the cluster name in kubernetes plugin config",
                            "type": "string"
                          },
                          "owner": {
                            "anyOf": [
                              {
                                "type": "string"
                              },
                              {
                                "additionalProperties": false,
                                "properties": {
                                  "kind": {
                                    "enum": [
                                      "user",
                                      "group"
                                    ],
                                    "type": "string"
                                  },
                                  "name": {
                                    "type": "string"
                                  },
                                  "namespace": {
                                    "type": "string"
                                  }
                                },
                                "required": [
                                  "name"
                                ],
                                "type": "object"
                              }
                            ]
                          }
                        },
                        "required": [
                          "kubernetesPluginRef"
                        ],
                        "type": "object"
                      },
                      {
                        "additionalProperties": false,
                        "properties": {
                          "caData": {
                            "description": "Base64-encoded CA bundle in PEM format used for verifying the TLS cert presented by the API server",
                            "type": "string"
                          },
                          "name": {
                            "description": "Name that the hub cluster will assume in Backstage Catalog (in OCM this is always local-cluster which can be confusing)",
                            "type": "string"
                          },
                          "owner": {
                            "anyOf": [
                              {
                                "type": "string"
                              },
                              {
                                "additionalProperties": false,
                                "properties": {
                                  "kind": {
                                    "enum": [
                                      "user",
                                      "group"
                                    ],
                                    "type": "string"
                                  },
                                  "name": {
                                    "type": "string"
                                  },
                                  "namespace": {
                                    "type": "string"
                                  }
                                },
                                "required": [
                                  "name"
                                ],
                                "type": "object"
                              }
                            ]
                          },
                          "serviceAccountToken": {
                            "description": "Service Account Token which is used for querying data from the hub",
                            "type": "string"
                          },
                          "skipTLSVerify": {
                            "description": "Skip TLS certificate verification presented by the API server, defaults to false",
                            "type": "boolean"
                          },
                          "url": {
                            "description": "Url of the hub cluster API endpoint",
                            "type": "string"
                          }
                        },
                        "required": [
                          "name",
                          "url"
                        ],
                        "type": "object"
                      }
                    ],
                    "description": "Key is reflected as provider ID. Defines and claims plugin instance ownership of entities"
                  },
                  "type": "object"
                }
              },
              "type": "object"
            }
          },
          "type": "object"
        }
      },
      "type": "object"
    },
    "HubClusterConfig": {
      "additionalProperties": false,
      "properties": {
        "caData": {
          "description": "Base64-encoded CA bundle in PEM format used for verifying the TLS cert presented by the API server",
          "type": "string"
        },
        "name": {
          "description": "Name that the hub cluster will assume in Backstage Catalog (in OCM this is always local-cluster which can be confusing)",
          "type": "string"
        },
        "serviceAccountToken": {
          "description": "Service Account Token which is used for querying data from the hub",
          "type": "string"
        },
        "skipTLSVerify": {
          "description": "Skip TLS certificate verification presented by the API server, defaults to false",
          "type": "boolean"
        },
        "url": {
          "description": "Url of the hub cluster API endpoint",
          "type": "string"
        }
      },
      "required": [
        "url",
        "name"
      ],
      "type": "object"
    },
    "KubernetesPluginRef": {
      "additionalProperties": false,
      "properties": {
        "kubernetesPluginRef": {
          "description": "Match the cluster name in kubernetes plugin config",
          "type": "string"
        }
      },
      "required": [
        "kubernetesPluginRef"
      ],
      "type": "object"
    },
    "Owner": {
      "additionalProperties": false,
      "properties": {
        "owner": {
          "anyOf": [
            {
              "type": "string"
            },
            {
              "additionalProperties": false,
              "properties": {
                "kind": {
                  "enum": [
                    "user",
                    "group"
                  ],
                  "type": "string"
                },
                "name": {
                  "type": "string"
                },
                "namespace": {
                  "type": "string"
                }
              },
              "required": [
                "name"
              ],
              "type": "object"
            }
          ]
        }
      },
      "type": "object"
    }
  }
}

```

where the owner field is optional but can't be as the only key in a hub config.

plugins/ocm-backend/src/helpers/config.ts
```diff
@@ -1,9 +1,12 @@
+import { DEFAULT_NAMESPACE } from '@backstage/catalog-model';
 import { Config } from '@backstage/config';
 import { OcmConfig } from '../types';
 
+export const OCM_PREFIX = 'catalog.providers.ocm';
+
```
tumido, line 6: This change is not necessary.

plugins/ocm-backend/src/helpers/config.ts
```diff
@@ -82,3 +85,18 @@ export const readOcmConfigs = (config: Config): OcmConfig[] => {
...
+  }
+
+  if (typeof hubConfig.get(OWNER_KEY) === 'string') {
+    return hubConfig.getString(OWNER_KEY);
+  }
+  const name = hubConfig.getString(`${OWNER_KEY}.name`);
+  const namespace =
+    hubConfig.getOptionalString(`${OWNER_KEY}.namespace`) || DEFAULT_NAMESPACE;
+  const kind = hubConfig.getOptionalString(`${OWNER_KEY}.kind`) || 'group';
+  return `${kind}:${namespace}/${name}`;
```
tumido, line 101: Do we need to do this parsing? Why don't we return it as is and let Backstage parse it? `spec.owner` field can be a compound reference.

https://backstage.io/docs/features/software-catalog/references#compound-references

We should avoid assuming defaults to avoid potentially diverting from upstream.

tumido (changes requested):

plugins/ocm-backend/src/helpers/config.ts
SamoKopecky: The owner field in the spec can only be a string, so thats why I think we have to use this parser. However if the default values from the parser for `kind` and `namespace` are removed we don't assume any defaults and the backstage parser will fill in the defaults on its own so we avoid assuming the defaults.
```suggestion
  if (typeof hubConfig.get(OWNER_KEY) === 'string') {
    return hubConfig.getString(OWNER_KEY);
  }
  const name = hubConfig.getString(`${OWNER_KEY}.name`);
  const namespace =
    hubConfig.getOptionalString(`${OWNER_KEY}.namespace`);
  const kind = hubConfig.getOptionalString(`${OWNER_KEY}.kind`);
  return `${kind}:${namespace}/${name}`;
```

plugins/ocm-backend/src/helpers/config.ts
SamoKopecky: Will be fixed in the latest commit.

plugins/ocm-backend/src/helpers/config.ts
tumido: @SamoKopecky let's do it so we allow the `OWNER_KEY` to be only a value that is already directly consumable by Backstage `.spec.owner`, WDYT?

SamoKopecky: Yes that actually makes a lot more sense then creating just a new parser that basically does the same thing as the backstage owner parser. I'll change it to just a simple string value.

SamoKopecky: The schema looks like this:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "Config": {
      "additionalProperties": false,
      "properties": {
        "catalog": {
          "additionalProperties": false,
          "properties": {
            "providers": {
              "additionalProperties": false,
              "properties": {
                "ocm": {
                  "additionalProperties": {
                    "anyOf": [
                      {
                        "additionalProperties": false,
                        "properties": {
                          "kubernetesPluginRef": {
                            "description": "Match the cluster name in kubernetes plugin config",
                            "type": "string"
                          },
                          "owner": {
                            "description": "Owner reference to created cluster entities in the catalog",
                            "type": "string"
                          }
                        },
                        "required": [
                          "kubernetesPluginRef"
                        ],
                        "type": "object"
                      },
                      {
                        "additionalProperties": false,
                        "properties": {
                          "caData": {
                            "description": "Base64-encoded CA bundle in PEM format used for verifying the TLS cert presented by the API server",
                            "type": "string"
                          },
                          "name": {
                            "description": "Name that the hub cluster will assume in Backstage Catalog (in OCM this is always local-cluster which can be confusing)",
                            "type": "string"
                          },
                          "owner": {
                            "description": "Owner reference to created cluster entities in the catalog",
                            "type": "string"
                          },
                          "serviceAccountToken": {
                            "description": "Service Account Token which is used for querying data from the hub",
                            "type": "string"
                          },
                          "skipTLSVerify": {
                            "description": "Skip TLS certificate verification presented by the API server, defaults to false",
                            "type": "boolean"
                          },
                          "url": {
                            "description": "Url of the hub cluster API endpoint",
                            "type": "string"
                          }
                        },
                        "required": [
                          "name",
                          "url"
                        ],
                        "type": "object"
                      }
                    ],
                    "description": "Key is reflected as provider ID. Defines and claims plugin instance ownership of entities"
                  },
                  "type": "object"
                }
              },
              "type": "object"
            }
          },
          "type": "object"
        }
      },
      "type": "object"
    },
    "HubClusterConfig": {
      "additionalProperties": false,
      "properties": {
        "caData": {
          "description": "Base64-encoded CA bundle in PEM format used for verifying the TLS cert presented by the API server",
          "type": "string"
        },
        "name": {
          "description": "Name that the hub cluster will assume in Backstage Catalog (in OCM this is always local-cluster which can be confusing)",
          "type": "string"
        },
        "serviceAccountToken": {
          "description": "Service Account Token which is used for querying data from the hub",
          "type": "string"
        },
        "skipTLSVerify": {
          "description": "Skip TLS certificate verification presented by the API server, defaults to false",
          "type": "boolean"
        },
        "url": {
          "description": "Url of the hub cluster API endpoint",
          "type": "string"
        }
      },
      "required": [
        "url",
        "name"
      ],
      "type": "object"
    },
    "KubernetesPluginRef": {
      "additionalProperties": false,
      "properties": {
        "kubernetesPluginRef": {
          "description": "Match the cluster name in kubernetes plugin config",
          "type": "string"
        }
      },
      "required": [
        "kubernetesPluginRef"
      ],
      "type": "object"
    }
  }
}

```

plugins/ocm-backend/src/helpers/config.ts
```diff
@@ -83,6 +84,7 @@ export const getHubClusterFromConfig = (
     serviceAccountToken: hub.getOptionalString('serviceAccountToken'),
     skipTLSVerify: hub.getOptionalBoolean('skipTLSVerify') || false,
     caData: hub.getOptionalString('caData'),
+    owner: hub.getOptionalString(OWNER_KEY) || 'unknown',
```
tumido, line 87: If `hub` is resolved via `kubernetesPluginRef` by `getHubClusterFromKubernetesConfig()`, it has no owners field. This needs to be handed/passed down in `getHubClusterFromKubernetesConfig()`

tumido (changes requested):

plugins/ocm-backend/src/helpers/config.ts
```diff
@@ -97,3 +98,7 @@ export const readOcmConfigs = (config: Config): OcmConfig[] => {
     .keys()
     .map(id => getHubClusterFromConfig(id, ocmConfigs.getConfig(id), config));
 };
+
+export const getOwnerFromHub = (hubConfig: Config): string => {
+  return hubConfig.getOptionalString(OWNER_KEY) || 'unknown';
+};
```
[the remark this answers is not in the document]
tumido: I think adding some tests is a good idea. As you can see in my other comment below, I think we need those tests here. 😄

plugins/ocm-backend/src/helpers/config.ts
SamoKopecky: Once #228 is merged I'll push the change that will fix this together with updated tests.

plugins/ocm-backend/src/helpers/config.ts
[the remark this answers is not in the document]
SamoKopecky: I've added tests in #228 together with a fix to the `url` validation since it wasn't working properly. Once it is merged I'll update the tests to also check for the owner configuration.

tumido (approved):
/lgtm
/approve

Issue #156, opened 2023-03-01
Pod label in case of rolling/recreate overlaps with pod ring


invincibleJai:
**Describe the bug**
Pod label in case of rolling/recreate overlaps with pod ring

**To Reproduce**
Run topology plugin , can follow https://github.com/janus-idp/backstage-plugins/tree/main/plugins/topology#readme

Create a deployment with a valid image let pods come 

```
kind: Deployment
apiVersion: apps/v1
metadata:
  name: minikube-dep-6
  labels:
    backstage.io/kubernetes-id: backstage
  annotations:
    deployment.kubernetes.io/revision: '3'
spec:
  replicas: 1
  selector:
    matchLabels:
      app: name
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: name
        backstage.io/kubernetes-id: backstage
    spec:
      containers:
        - name: container
          image: openshift/hello-openshift
          ports:
            - containerPort: 8080
              protocol: TCP
          resources: {}
          terminationMessagePath: /dev/termination-log
          terminationMessagePolicy: File
          imagePullPolicy: Always
      restartPolicy: Always
      terminationGracePeriodSeconds: 30
      dnsPolicy: ClusterFirst
      securityContext: {}
      schedulerName: default-scheduler
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 25%
      maxSurge: 25%
  revisionHistoryLimit: 10
  progressDeadlineSeconds: 600

```

and then edit the deployment image to an invalid image and in topology view can see Pod label in case of rolling/recreate overlaps with pod ring

**Expected behavior**
Pod label text shouldn't overlap with pod rings

**Screenshots**

<img width="663" alt="image" src="https://user-images.githubusercontent.com/5129024/222167066-68143b18-f28d-4f51-9027-378c2f8b5aa9.png">

**Additional context**

PR #159, opened 2023-03-01
fix(topology): fixes pods label overlap with rolling recreate

invincibleJai:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/156

**Description:**
Pod label in case of rolling/recreate overlaps with pod ring

**Before:**

image

**After:**

image


tumido (approved):
/lgtm
/approve

Issue #163, opened 2023-03-09
Bug: Topology not showing correct deployment and pod


PatAKnight:
**Describe the bug**
The deployment and pod are not correctly showing in the topology plugin tab in the showcase app. 
**To Reproduce**

1. Go to 'topology tab in showcase app'

**Expected behavior**
Should show the correct information
**Screenshots**
![Screenshot from 2023-03-09 11-39-46](https://user-images.githubusercontent.com/82788581/224091612-d08be9c3-4666-4dba-b209-183c24813208.png)

![Screenshot from 2023-03-09 11-39-33](https://user-images.githubusercontent.com/82788581/224091658-3a735716-da47-4103-b319-ceda79cb5d26.png)

**Additional context**

tumido:
cc @invincibleJai @divyanshiGupta

invincibleJai:
paylod response for above https://gist.github.com/invincibleJai/09762d354807aa875da39f0da92af8c3

PR #164, opened 2023-03-09
fix(topology): getDeploymentRevision always return false but not the revision for deployment

invincibleJai:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/163

**Before:**

image

**After:**

image

### DIFF
--- a/plugins/topology/src/utils/pod-resource-utils.ts
+++ b/plugins/topology/src/utils/pod-resource-utils.ts
@@ -30,9 +30,12 @@ const isIdled = (deployment: K8sWorkloadResource): boolean => {
   ];
 };
 
-const getDeploymentRevision = (obj: K8sWorkloadResource) => {
-  const revision = getAnnotation(obj, DEPLOYMENT_REVISION_ANNOTATION);
-  return revision && Number.isFinite(revision) && parseInt(revision, 10);
+export const getDeploymentRevision = (
+  obj: K8sWorkloadResource,
+): number | null => {
+  const revisionAnnotation = getAnnotation(obj, DEPLOYMENT_REVISION_ANNOTATION);
+  const revision = revisionAnnotation && parseInt(revisionAnnotation, 10);
+  return revision && Number.isFinite(revision) ? revision : null;
 };
 
 const getOwnedResources = <T extends K8sWorkloadResource>(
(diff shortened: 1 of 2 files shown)


plugins/topology/src/utils/pod-resource-utils.ts
```diff
@@ -32,7 +32,7 @@ const isIdled = (deployment: K8sWorkloadResource): boolean => {
 
 const getDeploymentRevision = (obj: K8sWorkloadResource) => {
   const revision = getAnnotation(obj, DEPLOYMENT_REVISION_ANNOTATION);
-  return revision && Number.isFinite(revision) && parseInt(revision, 10);
+  return revision && parseInt(revision, 10);
```
invincibleJai, line 35: `Number.isFinite` will always return false here as revision is string

spadgett, line 35: Maybe

```suggestion
  const revision = parseInt(revisionAnnotation, 10);
  return Number.isFinite(revision) ? revision : null;
```

invincibleJai: Thaks Sam, have updated it.

plugins/topology/src/utils/pod-resource-utils.ts
```diff
@@ -30,9 +30,12 @@ const isIdled = (deployment: K8sWorkloadResource): boolean => {
...
 
-const getDeploymentRevision = (obj: K8sWorkloadResource) => {
-  const revision = getAnnotation(obj, DEPLOYMENT_REVISION_ANNOTATION);
-  return revision && Number.isFinite(revision) && parseInt(revision, 10);
+export const getDeploymentRevision = (
+  obj: K8sWorkloadResource,
+): number | null => {
+  const revisionAnnotation = getAnnotation(obj, DEPLOYMENT_REVISION_ANNOTATION);
+  const revision = revisionAnnotation && parseInt(revisionAnnotation, 10);
+  return revision && Number.isFinite(revision) ? revision : null;
```
spadgett, line 38: nitpick: I think this can be

```suggestion
  const revision = parseInt(revisionAnnotation, 10);
  return Number.isFinite(revision) ? revision : null;
```

which should handle `undefined`, but I am OK adding the extra checks too

spadgett (approved):

plugins/topology/src/utils/pod-resource-utils.ts
invincibleJai: added checks as linters are not happy as `revisionAnnotation` can be undefined.

invincibleJai:
Thanks Sam/Patrick for reviewing.

@tumido PTAL when you get time.

tumido (approved):
/lgtm
/approve

PR #167, opened 2023-03-14
feat(ocm): Add header to the ocm cluster status page

SamoKopecky:
Part of https://github.com/janus-idp/backstage-showcase/issues/63

Add a `header` instead of a `typography` field to the `ocm` cluster status page.

Result:
image


tumido (approved):
/lgtm
/approve

PR #185, opened 2023-03-21
test(quay): unit tests and refactor utils

### LINKED ISSUE #103: Refactor: Quay plugin code quality improvements
Look into following suggestions and address them if needed:
- [ ] https://github.com/janus-idp/backstage-plugins/pull/101#pullrequestreview-1290936392_

Code is sometimes a copy-paste from quay source and some decisions are not obvious/not great quality. Let's revisit that.

SamoKopecky:
Resolves #103 

Changes:
- Refactor the code that was mentioned in #101 
- Add tests for most of the functions outside of the react components
- Part of adding tests was also adding fixtures so that the API can be tested using the `msw` package

A dev setup could also be created using the defined fixtures, let me know if that is something we want to do.


plugins/quay/src/handlers.ts
```diff
@@ -0,0 +1,72 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { rest } from 'msw';
+
+const LOCAL_ADDR = 'https://localhost:5050/quay/';
+
+export const handlers = [
```
RobotSail, line 6: Since it looks like these handlers are only being used for testing purposes by `plugins/quay/src/index.ts`, you can just move this definition to `index.test.ts` and delete this file. 

In general, I would expect a file named `src/handlers.ts` to be exposing API route handlers that are consumed by a server somewhere. So if you do plan on consuming these routes from multiple test files, you would want to name the file something like `test-handlers.ts`. if they are meant to be consumed by multiple files.

plugins/quay/src/components/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+      o.has(k) ? o.get(k) : '#8A8D90',
+  },
+);
+
+export function formatDate(date: string | number): string {
+  if (date === -1) {
+    return 'N/A';
+  }
+
+  const adjustedDate = typeof date === 'number' ? date * 1000 : date;
```
RobotSail, line 24: Is date here here expressed in seconds and being converted to MS? 
If so, then you can express that with a TS docstring like:
```ts
/**
 * Returns the given date as a formated Date.
 * 
 * @param date - The given date in seconds
 * @return The date formatted to en-US locale
 */
```

plugins/quay/src/components/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+    [VulnerabilitySeverity.Low, '#F0AB00'],
+    [VulnerabilitySeverity.None, '#3E8635'],
+  ]) as any,
+  {
+    get: (o: Map<VulnerabilitySeverity, string>, k: VulnerabilitySeverity) =>
+      o.has(k) ? o.get(k) : '#8A8D90',
+  },
+);
+
+export function formatDate(date: string | number): string {
```
RobotSail, line 19: It's a bit of a nit-pick but you can rewrite this function to follow the single-responsibility principle like this:
```ts
export function formatDate(date: string | number): string {
  if (isNotAvailable(date)) {
    return 'N/A';
  }

  const adjustedDate = adjustDate(date);
  return formatAdjustedDate(adjustedDate);
}

function isNotAvailable(date: string | number): boolean {
  return date === -1;
}

function adjustDate(date: string | number): number | string {
  return typeof date === 'number' ? date * 1000 : date;
}

function formatAdjustedDate(adjustedDate: number | string): string {
  return new Date(adjustedDate).toLocaleString('en-US', {
    timeZone: getUserTimeZone(),
    timeStyle: 'short',
    dateStyle: 'medium',
  });
}

function getUserTimeZone(): string {
  return new Intl.DateTimeFormat().resolvedOptions().timeZone;
}
```

plugins/quay/src/components/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+}
+
+export function formatSize(sizeInBytes: number): string {
+  if (!sizeInBytes) {
+    // null or undefined
+    return 'N/A';
+  }
+
+  const i = Math.floor(Math.log(sizeInBytes) / Math.log(1024));
```
RobotSail, line 38: rename `i` to `unitIndex` and move the calcuation of it to a separate function called `calculateUnitIndex`
```suggestion
  const unitIndex = calculateUnitIndex(sizeInBytes);
	// 
	
function calculateUnitIndex(sizeInBytes: number): number {
	return Math.floor(Math.log(sizeInBytes) / Math.log(1024));
}
```

plugins/quay/src/api/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+  });
+
+  describe('getLabels', () => {
+    it('should correctly get the labels using the manifest digest', async () => {
+      const manifest = require(`${__dirname}/../../fixtures/manifests/bar.json`);
+      const result = await quayApi.getLabels('foo', 'bar', manifest.digest);
+
+      expect(result).toEqual(
+        require(`${__dirname}/../../fixtures/labels/foo.json`),
```
RobotSail, line 106: I looked at the fixtures directory and it seems like only this test is using them + the handlers. You should move `${PROJECT_DIR}/fixtures` to instead be located at `${PROJECT_DIR}/src/api/fixtures`. 
If we decide to add more tests in other directories in the future which reference these test files, we can think about moving it to a higher directory then. 
This way, we aren't emphasizing the importance of test data in our project.

RobotSail (changes requested):
This PR looks good overall, just some minor changes and then it should be good to merge 👌

SamoKopecky:
> This PR looks good overall, just some minor changes and then it should be good to merge ok_hand

Thanks for the feedback!

plugins/quay/src/api/index.test.ts
SamoKopecky: I agree with that, moved the fixtures and the handlers to the api folder.

plugins/quay/src/components/utils.ts
SamoKopecky: I've just extracted the `getUserTimeZone` function since I see how the implementation could change in the future.

SamoKopecky:
I've updated the PR according to your suggestions, let me know what you think.

plugins/quay/src/components/utils.ts
RobotSail: Perfect!

SamoKopecky:
Fixed grammatical errors in docs.

plugins/quay/src/api/index.ts
```diff
@@ -10,7 +10,7 @@ import {
   SecurityDetailsResponse,
 } from '../types';
 
-const DEFAULT_PROXY_PATH = '/quay/api';
+const DEFAULT_PROXY_PATH = '/quay';
```
tumido, line 13: Why is this being changed? What is the benefit? This change is not reflected in the readme therefore users expose `/quay/api` in their proxy settings and are confused.

https://github.com/janus-idp/backstage-plugins/tree/main/plugins/quay#getting-started

tumido (changes requested):

plugins/quay/src/api/index.ts
SamoKopecky: This change was made because the test `QuayApiClient should use a correct default proxy path` was failing if the default proxy path was `/quay/api` since the `/api` part is already inserted in the api functions like [`getTags`](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/quay/src/api/index.ts#L93).

So I think in a real deployment the functions are using paths with `*/api/api/*` inside the encoded url but this works because the quay api sends a redirect see this call for example:
```
curl -v -H "X-Requested-With: XMLHttpRquest" https://quay.io/api/api/v1/repository/janus-idp/redhat-backstage-build/manifest/sha256:e461dc54b4e2469bb7f5bf85a4b7445c175548ba9d56c3f617dd25bc3adf3752/security
```
and the fetcher handles the redirect. However the mocked quay API doesn't send a redirect and that is why the test fails while using the `/quay/api` as the default proxy path. 

I see two ways how to approach this:
- we can either change the documentation so that the `API` in the config is supposed to only be `/quay` and the default should also be `/quay` 
- or we can remove the `/api` from the API function calls in `index.ts` and let the default be `/quay/api` and don't change the documentation

What do you think is the correct approach?

tumido: I think your problem is a product of you locally setting the target to `https://quay.io/api` instead of `https://quay.io`. Not a product of setting the proxy path.

with default proxy path `/quay/api` and target `https://quay.io` the proxy resolves:

`/quay/api/api/v1/repository/janus-idp/redhat-backstage-build...`

gets translated to:

`https://quay.io/api/v1/repository/janus-idp/redhat-backstage-build...`

this is the correct resolution.

SamoKopecky: That's true, I totally forgot about the proxy. I will try to also mock the proxy translation in the tests.

SamoKopecky: I reverted the change to the default proxy path.

tumido (approved):
/lgtm
/approve

PR #190, opened 2023-03-22
fix(ocm-backend): Improve error reporting and check for url

SamoKopecky:
This PR is related to [this slack thread](https://janus-idp.slack.com/archives/C04EDTPJRK5/p1679181248576309) and might help troubleshoot this and similar issues.

Add an `url` validator before the k8s config is created and add a fallback for the error handler if the response doesn't contain a header (which is the case in the slack thread issue).


tumido:
can you please rebase? @SamoKopecky  🙂

SamoKopecky:
Will do!

tumido (approved):
/lgtm
/approve

PR #196, opened 2023-03-23
feat(tekton): frontend plugin setup for tekton

invincibleJai:
Tracks: https://github.com/janus-idp/backstage-plugins/issues/201

**Description:** 
Setup a frontend plugin for tekton with dev setup

**Test Setup**

- For Dev test

```
yarn install
yarn workspace @janus-idp/backstage-plugin-tekton run start

```

- For normal test

1. Install and configure Tekton plugin by following https://github.com/jquad-group/backstage-jquad#configuration
2. Enable an additional tab on the entity view page

   ```ts
   import { TektonPage } from '@janus-idp/backstage-plugin-tekton';

   const serviceEntityPage = (
     <EntityPageLayout>
       // ...
       <EntityLayout.Route path="/pipelines" title="Pipelines">
         <TektonPage />
       </EntityLayout.Route>
     </EntityPageLayout>
   );
   ```
   
**Todo**

- [x]  Setup a frontend plugin for tekton with dev setup
- [x]  Fetch data from upstream tekton
- [x]  Adds patternfly dependencies
- [x]  Adds React testing library dependencies
- [x]  Supports dark theme

### DIFF
--- a/plugins/tekton/README.md
+++ b/plugins/tekton/README.md
@@ -1,0 +1,39 @@
+# Tekton plugin for Backstage
+
+This plugin will help with visualizing the Pipeline Runs and associated Task Runs powering any service on the Kubernetes cluster.
+
+## Prerequisites
+
+1. Install and configure Tekton plugin (only backend plugin) by following https://github.com/jquad-group/backstage-jquad#configuration
+
+## Getting started
+
+1. Install the plugin
+
+   ```bash
+   yarn workspace app add @janus-idp/backstage-plugin-tekton
+   ```
+
+2. Enable an additional tab on the entity view page
+
+   ```ts
+   // packages/app/src/components/catalog/EntityPage.tsx
+   import { TektonPage } from '@janus-idp/backstage-plugin-tekton';
+
+   const serviceEntityPage = (
+     <EntityPageLayout>
+       // ...
+       <EntityLayout.Route path="/pipelines" title="Pipelines">
+         <TektonPage />
+       </EntityLayout.Route>
+     </EntityPageLayout>
+   );
+   ```
+
+## Development
+
+In [Backstage plugin terminology](https://backstage.io/docs/local-dev/cli-build-system#package-roles), this is a `frontend-plugin`. You can start a live dev session from the repository root using below command:
+
+```
+yarn workspace @janus-idp/backstage-plugin-tekton run start
+```
--- a/plugins/tekton/.eslintrc.js
+++ b/plugins/tekton/.eslintrc.js
@@ -1,0 +1,1 @@
+module.exports = require('@backstage/cli/config/eslint-factory')(__dirname);
--- a/plugins/tekton/dev/index.tsx
+++ b/plugins/tekton/dev/index.tsx
@@ -1,0 +1,82 @@
+import * as React from 'react';
+import { EntityProvider } from '@backstage/plugin-catalog-react';
+import { Entity } from '@backstage/catalog-model';
+import { createDevApp } from '@backstage/dev-utils';
+import { TestApiProvider } from '@backstage/test-utils';
+import {
+  Cluster,
+  PipelineRun,
+  PipelineRunsByEntityRequest,
+} from '@jquad-group/plugin-tekton-pipelines-common';
+import { tektonPlugin, TektonPage } from '../src/plugin';
+import { TektonApi, tektonApiRef } from '@jquad-group/plugin-tekton-pipelines';
+import { pipelineRunMock } from '../src/__fixtures__/mockPipelinesData';
+
+const mockEntity: Entity = {
+  apiVersion: 'backstage.io/v1alpha1',
+  kind: 'Component',
+  metadata: {
+    name: 'backstage',
+    description: 'backstage.io',
+    annotations: {
+      'backstage.io/kubernetes-id': 'dice-roller',
+      'tektonci/build-namespace': 'sample-go-application-build',
+    },
+  },
+  spec: {
+    lifecycle: 'production',
+    type: 'service',
+    owner: 'user:guest',
+  },
+};
+
+class MockTektonClient implements TektonApi {
+  getLogs(
+    _baseUrl: string,
+    _authorizationBearerToken: string,
+    _namespace: string,
+    _taskRunPodName: string,
+    _stepContainer: string,
+  ): Promise<string> {
+    const logMock = Promise.resolve('this is example log');
+    return logMock;
+  }
+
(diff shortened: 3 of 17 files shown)


invincibleJai:
@tumido Will it make sense to have the initial PR just have `Setup a frontend plugin for tekton with dev setup` and features will be added in follow-up PRs? Not sure if it's ideal but will help to have smaller PRs. LMK your thoughts.

plugins/tekton/README.md
```diff
@@ -0,0 +1,39 @@
+# Tekton plugin for Backstage
+
+This plugin will help with visualizing the Pipeline Runs and associated Task Runs powering any service on the Kubernetes cluster.
+
+## Prerequisites
+
+1. Install and configure Tekton plugin by following https://github.com/jquad-group/backstage-jquad#configuration
```
tumido, line 7: Please emphasize that the user should install and configure **only** the backend tekton plugin this way. We want to make sure the user is not confused by this.

tumido (commented):
LGTM 👍 I like this approach a lot. I think we just need slight clarification/disclaimer in the readme. WDYT?

invincibleJai:
@tumido Thaks for review.  User needs to install and configure the complete tekton plugin (both frontend and backend) for this plugin to use the hook directly `usePipelineRunObjects`  https://github.com/janus-idp/backstage-plugins/pull/196/files#diff-da201b26d1d64a686ffa285e470876ed967939b13762860b27a31d745850936aR30.

Otherwise, there would be some duplicity in this frontend plugin.

plugins/tekton/README.md
invincibleJai: @tumido user needs to install and configure the complete tekton plugin (both frontend and backend) for this plugin to use the hook directly `usePipelineRunObjects`  https://github.com/janus-idp/backstage-plugins/pull/196/files#diff-da201b26d1d64a686ffa285e470876ed967939b13762860b27a31d745850936aR30.

Otherwise, there would be some duplicity in this frontend plugin.

tumido:
@invincibleJai well... I'm not sure that's what you want though. **If** you want to rely on the upstream tekton plugin to provide you with the API, you **need** the have the `EntityTektonPipelinesContent` upstream compoment mounted somewhere in the JSX static. That means the upstream plugin needs to display its UI somewhere on the page (as another tab for example). If you remove the instance of the upstream component from the page, you'll get:

```
No implementation available for apiRef{plugin.tekton-pipelines.service}
```

In other words, (if I'm doing this right and didn't miss anything) user has no means to display only our UI, it always has to be displayed together with the upstream one. And I don't think we want that.

Since your plugin already lists the `@jquad-group/plugin-tekton-pipelines{,-common}` as its dependencies, the code comes in bundled. You just have to reexport it so backstage can use this plugin as a standalone. As a result, user doesn't have to install the upstream frontend plugin, they will only install this one (+ the obvious upstream backend).

invincibleJai:
> @invincibleJai well... I'm not sure that's what you want though. **If** you want to rely on the upstream tekton plugin to provide you with the API, you **need** the have the `EntityTektonPipelinesContent` upstream compoment mounted somewhere in the JSX static. That means the upstream plugin needs to display its UI somewhere on the page (as another tab for example). If you remove the instance of the upstream component from the page, you'll get:
> 
> ```
> No implementation available for apiRef{plugin.tekton-pipelines.service}
> ```
> 
> In other words, (if I'm doing this right and didn't miss anything) user has no means to display only our UI, it always has to be displayed together with the upstream one. And I don't think we want that.
> 
> Since your plugin already lists the `@jquad-group/plugin-tekton-pipelines{,-common}` as its dependencies, the code comes in bundled. You just have to reexport it so backstage can use this plugin as a standalone. As a result, user doesn't have to install the upstream frontend plugin, they will only install this one (+ the obvious upstream backend).


@tumido Thanks, have updated PR to make use on only backend, effectively needed to add [apiFactory](https://github.com/janus-idp/backstage-plugins/pull/196/files#diff-b480aa0932b1bd7241e6d152f7335a17afc9506ec5a9720895a712af3e2a3fe0R13-R21) and updated README as well. LMK if it makes sense

tumido:
@invincibleJai it totally makes sense and I like that approach. However I'm struggling to make it work with a full instance (not a dev setup). 

I'm getting following:
image

When I've had to change following and than it started working:

```diff
diff --git a/plugins/tekton/src/api/types.ts b/plugins/tekton/src/api/types.ts
index 69b5c34..7a09788 100644
--- a/plugins/tekton/src/api/types.ts
+++ b/plugins/tekton/src/api/types.ts
@@ -2,5 +2,5 @@ import { createApiRef } from '@backstage/core-plugin-api';
 import { TektonApi } from '@jquad-group/plugin-tekton-pipelines';

 export const tektonPluginApiRef = createApiRef<TektonApi>({
-  id: 'plugin.tekton.service',
+  id: 'plugin.tekton-pipelines.service',
 });
diff --git a/plugins/tekton/src/plugin.ts b/plugins/tekton/src/plugin.ts
index 33f7aa9..3c84113 100644
--- a/plugins/tekton/src/plugin.ts
+++ b/plugins/tekton/src/plugin.ts
@@ -9,7 +9,7 @@ import { TektonBackendClient } from './api/TektonBackendClient';
 import { tektonPluginApiRef } from './api/types';

 export const tektonPlugin = createPlugin({
-  id: 'tekton',
+  id: 'tekton-pipelines',
   apis: [
     createApiFactory({
       api: tektonPluginApiRef,
diff --git a/plugins/tekton/src/routes.ts b/plugins/tekton/src/routes.ts
index 78fddea..de829f8 100644
--- a/plugins/tekton/src/routes.ts
+++ b/plugins/tekton/src/routes.ts
@@ -1,5 +1,5 @@
 import { createRouteRef } from '@backstage/core-plugin-api';

 export const rootRouteRef = createRouteRef({
-  id: 'tekton',
+  id: 'tekto-pipelines',
 });
```

I think the provided api name has to match the expected name by the upstream plugin. WDYT? Does it work for you fine?

You can see my setup here: https://github.com/tumido/backstage-plugins/commit/aa82d4ad18e13f9bbde876f1b547f7ced87a10a4 

Please verify if those changes are needed for an user installation to work properly or I'm doing something wrong.

invincibleJai:
Closing this PR as we have https://github.com/janus-idp/backstage-plugins/pull/214

There was a need to get more data like pipeline spec and tasksRuns to get Pipeline run visualisation to show all the tasks structure (parallel, finally etc)

PR #198, opened 2023-03-23
feat(artifactory) First funtional version of the plugin

### LINKED ISSUE #117: 🔌 Plugin: OCI Registry (JFrog)
## Goal
This epic aims to develop a new Backstage plugin in the Janus repo. This plugin should allow Backstage users to view details of an image related to a catalog resource deployed from your JFrog Artifactory repository.

## Acceptance criteria
- [ ] Users should be able to view the tags, last update and SHA of an image related to a catalog deployed by a **JFrog** image 

## Requirements
- [ ] Plugin
- [x] Test plan
- [ ] Installation documentation
- [ ] End user documentation

## Issues in Epic
- [x] https://github.com/janus-idp/backstage-plugins/issues/559 [Doc]
- [x] https://github.com/janus-idp/backstage-plugins/issues/558 [QE]

## Notes
There is a question as to if there should be an OCI Registry frontend plugin which has a number of backend plugins (Quay, ECR, GCR, Nexus, JFrog)

**Additional context**

fmenesesg:
Resolves #117 

Get the image data from an Artifactory/JCR registry

artifactoryPluginPage

### DIFF
--- a/plugins/artifactory/.eslintrc.js
+++ b/plugins/artifactory/.eslintrc.js
@@ -1,0 +1,1 @@
+module.exports = require('@backstage/cli/config/eslint-factory')(__dirname);
--- a/plugins/artifactory/README.md
+++ b/plugins/artifactory/README.md
@@ -1,0 +1,67 @@
+# Artifactory plugin for Backstage
+
+This plugin will show you information about your container images within Artifactory registry
+
+## Getting started
+
+1. Install the plugin
+
+   ```bash
+   yarn workspace app add @janus-idp/backstage-plugin-artifactory
+   ```
+
+2. Set the proxy to desired Artifactory server
+
+   ```yaml
+   # app-config.yaml
+   proxy:
+     '/artifactory/api':
+       target: 'http://<hostname>:8082'
+       headers:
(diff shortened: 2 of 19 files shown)


tumido:
@fmenesesg Nice plugin! However I have one trouble here. This plugin shares a lot of codebase with the `@janus-idp/backstage-plugin-quay`. I don't really want to get into the business of maintaining this much duplicate code. What if we move the shared component into a common/shared package instead? That way we can reuse the same components in both of these plugins. It would essentially mean that the UI layer is shared and these individual plugins provide an API layer for data fetching. WDYT about that?

fmenesesg:
@tumido I totally agree with you it makes perfect sense, we need to decide what information this standard UI component will look like because at least JCR has some differences with Quay. Do you think we should work on that common component on this issue or we should create a new issue for that work?

tumido:
@fmenesesg you are right.. and even if we decide to abstract the shared component into a component library, we can do so later anyways.. so let's roll with this and we can refactor later. I'll try to review this asap 🙂

tumido (approved):
/lgtm
/approve

Issue #177, opened 2023-03-15
Custom action: Create a SonarQube project


serenamarie125:
## Description
As a Platform Engineer, I want to be able to use a custom action to **create a SonarQube project** in my GPT.

jlmayorga:
@serenamarie125: Would it make sense to create an action for SonarQube [Applications](https://docs.sonarqube.org/latest/user-guide/applications/) as well?

christophe-f:
That would be a nice addition. Maybe for a v2?

Currently, I'm not sure if the SonarQube plugin handles that (Applications):
- https://github.com/backstage/backstage/blob/master/plugins/sonarqube/README.md
- https://github.com/backstage/backstage/blob/master/plugins/sonarqube-backend/README.md

regicsolutions:
Agree that supporting both Projects and Applications would be great

rtaniwa:
Delivered via https://github.com/janus-idp/backstage-plugins/pull/206

PR #206, opened 2023-03-27
feat(sonarqube): Custom action to create a SonarQube project

jlmayorga:
New Backstage [custom action](https://backstage.io/docs/features/software-templates/writing-custom-actions#registering-custom-actions) to create a SonarQube project

Takes care of #177


plugins/sonarqube-actions/src/actions/index.ts
```diff
@@ -0,0 +1 @@
+export {createSonarQubeProjectAction} from './createSonarQubeProject'
```
tumido, line 1: Missing EOL at EOF

plugins/sonarqube-actions/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@janus-idp/sonarqube-actions",
```
tumido, line 2:
```suggestion
  "name": "@janus-idp/backstage-scaffolder-backend-module-sonarqube",
```

Looking at all available scaffolder plugins out there, all use a more descriptive name that contains `scaffolder-backend-module`. Let's follow the conventions, so the module is easier to discover for users.

tumido (changes requested):
Hey @jlmayorga! Thanks for contributing this! 👍 I have some formalities/nits inline.

plugins/sonarqube-actions/src/actions/index.ts
jlmayorga: Fixed in c0af8ee

tumido (approved):
/lgtm
/approve

tumido:
> Should the plugin directory be renamed as well?

@jlmayorga I don't see a need to rename the folder at this point. We can give the folder structure a redesign/revisit at later point. Right now, I think this is fine, it's understandable for us (me, you, other maintainers). When it becomes a challenge on that front, we may consider renaming it. 🙂

PR #207, opened 2023-03-27
feat(quay): add quay link to repository

dweber019:
Provide a link to the quay hosted repository in the table header.
This enables users of the plugin to navigate into quay and get more insights.

CleanShot 2023-03-29 at 20 38 08@2x

### DIFF
--- a/plugins/quay/README.md
+++ b/plugins/quay/README.md
@@ -24,6 +24,10 @@ This plugin will show you information about your container images within Quay re
        changeOrigin: true
        # Change to "false" in case of using self hosted quay instance with a self-signed certificate
        secure: true
+
+   quay:
+     # The UI url for Quay, used to generate the link to Quay
+     uiUrl: 'https://quay.io'
    ```
 
 3. Enable additional tab on the entity view page
--- a/plugins/quay/config.d.ts
+++ b/plugins/quay/config.d.ts
@@ -2,9 +2,14 @@ export interface Config {
   /** Configurations for the Quay plugin */
   quay?: {
     /**
-     * The base url of the Quay instance.
+     * The proxy path for the Quay instance.
      * @visibility frontend
      */
     proxyPath?: string;
+    /**
+     * The UI url of the Quay instance.
+     * @visibility frontend
+     */
+    uiUrl?: string;
   };
 }
--- a/plugins/quay/src/components/QuayRepository/QuayRepository.tsx
+++ b/plugins/quay/src/components/QuayRepository/QuayRepository.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { configApiRef, useApi } from '@backstage/core-plugin-api';
 import { Link, Progress, Table } from '@backstage/core-components';
 import { columns, useStyles } from './tableHeading';
 import { useRepository, useTags } from '../../hooks';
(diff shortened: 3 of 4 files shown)


tumido (commented):
Hi @dweber019, thank you for your contribution here. Can you please explain what do you want to achieve by this PR?

Since you have not provided any description I'm gonna be guessing from the code. It seems you're adding a new optional field to the settings that is used for display purposes only, is that correct? It doesn't need to match the proxy target at all and we basically accept any value, correct?

plugins/quay/README.md
```diff
@@ -24,6 +24,9 @@ This plugin will show you information about your container images within Quay re
        changeOrigin: true
        # Change to "false" in case of using self hosted quay instance with a self-signed certificate
        secure: true
+
+   quay:
```
tumido, line 28: Can you please describe the meaning of this setting in the `README.md` as well? We may want to include it as a separate and optional configuration step. When listed as a part of the mandatory proxy setting, it may be viewed as a required step.

plugins/quay/config.d.ts
```diff
@@ -2,9 +2,14 @@ export interface Config {
...
-     * The base url of the Quay instance.
+     * The proxy path for the Quay instance.
      * @visibility frontend
      */
     proxyPath?: string;
+    /**
+     * The base url of the Quay instance.
+     * @visibility frontend
+     */
+    baseUrl?: string;
```
tumido, line 13: Should this be rather named `uiUrl` or something? So it's obvious this value is not used for in API queries at all, only for display and link purposes.

tumido (commented):
Hi @dweber019, sorry for the delay on this.

Very interesting... is it possible to host Quay UI separately from the registry? I'm just curious, I never heard of this.

I'm thinking about the `baseUrl` config property name. I don't want users to confuse it with the proxy `target`... WDYT, can it be confusing?

cc @fmenesesg @christophe-f

plugins/quay/config.d.ts
dweber019: Yes would make more sense for the user. Will update this.

dweber019:
> is it possible to host Quay UI separately from the registry? I'm just curious, I never heard of this.

It's possible to provide different URL for UI and the registry endpoint from a k8s standpoint of view.

plugins/quay/src/components/QuayRepository/QuayRepository.tsx
```diff
@@ -21,10 +21,25 @@ const useLocalStyles = makeStyles({
 
 export function QuayRepository(props: RepositoryProps) {
   const quayClient = useApi(quayApiRef);
+  const configApi = useApi(configApiRef);
   const classes = useStyles();
   const localClasses = useLocalStyles();
   const [tags, setTags] = useState<Tag[]>([]);
-  const title = `Quay repository: ${props.organization}/${props.repository}`;
+
+  const quayBaseUrl = configApi.getOptionalString('quay.uiUrl');
```
tumido, line 29:
```suggestion
  const quayUiUrl = configApi.getOptionalString('quay.uiUrl');
```

tumido (changes requested):
Sorry for delays in review here. I have just a tiny last nit and we can merge this. 👍

tumido (approved):
/lgtm
/approve

PR #240, opened 2023-04-05
feat(quay): add quay security details

### LINKED ISSUE #231: Display the Security scanner result in the Quay plugin
**Describe the solution you'd like**
Quay does provide a security scan of the image. It would be great to see the number of vulnerabilities (and the severity) in the history table. If an image has an issue, a link should redirect to the Quay page with all the information.

https://developers.redhat.com/blog/2019/06/26/using-quay-io-to-find-vulnerabilities-in-your-container-images

RobotSail:
- fix(quay): improve code readability with hooks
- feat(quay): Refactor code and introduce tag details as subroute

This PR adds the ability to view security details for a given tag manifest.
You can access the view by either clicking through the security details on the main tag 🏷️ list, or by
navigating to `/quay/tag/{manifest}` relative to where the Quay plugin is mounted.

Closes #231 
Related to #60 

### Changes to `QuayRepositoryPage`:

image

### Addition of `QuayTagPage`:

Vulnerabilities are parsed and displayed in table format, adhering to the updated Quay UI format.

image


plugins/quay/package.json
```diff
@@ -31,7 +31,10 @@
     "@material-ui/core": "^4.12.2",
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "4.0.0-alpha.61",
-    "react-use": "^17.2.4"
+    "react-router": "^6.3.0",
+    "react-router-dom": "^6.3.0",
```
tumido, line 35: These should be peer dependencies.

packages/backend/package.json
tumido, line 1: Can we please revert this change? We don't need the frontend plugin in the backend app.

plugins/quay/src/components/QuayDashboardPage/QuayDashboardPage.tsx
```diff
@@ -1,15 +1,9 @@
...
+import { useRepository } from '../../hooks';
 
 export const QuayDashboardPage = () => {
-  const { entity } = useEntity();
-  const { repositorySlug } = useQuayAppData({ entity });
-  const info = repositorySlug.split('/');
-
-  const organization = info.shift() as 'string';
-  const repository = info.join('/');
+  const { repository, organization } = useRepository();
```
tumido, line 6: Right now you're using a context to get these values and then simply passing them to `QuayRepository` as props.

I think this would be better if you'd get those values from the context in `QuayRepository` directly. WDYT? Is there any other use for `QuayRepository` when it would not consume the context values anyways and require props?

plugins/quay/src/components/QuayTagPage/component.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { useRepository, useTagDetails } from '../../hooks';
+import { useParams } from 'react-router';
+import { QuayTagDetails } from '../QuayTagDetails';
+import { ErrorPanel } from '@backstage/core-components';
+// import { QuayTagDetails } from '../QuayTagDetails';
+
+type QuayTagPageProps = Record<never, string>;
+
+// TODO: replace useAsync with tanstack query
```
tumido, line 10: There's no `useAsync` in this component.

plugins/quay/src/hooks/state.ts
```diff
@@ -0,0 +1,34 @@
...
+export type QuayStore = {
+  repoTags: Record<string, Tag[]>;
+  tagManifestLayers: Record<string, Layer>;
+  setRepoTags: (org: string, repo: string, tags: Tag[]) => void;
+  setTagManifestLayers: (digest: string, layer: Layer) => void;
+};
+
+/**
+ * This is a global store for Quay data so that it can be shared
+ * between components without having to pass it down through props.
```
tumido, line 13: Why do we need this?

plugins/quay/src/types.ts
```diff
@@ -15,6 +15,7 @@ export interface Tag {
   end_ts?: number;
   manifest_list?: ManifestList;
   expiration?: string;
+  // securityDetails: SecurityDetailsResponse;
```
tumido, line 18: Please cleanup.

plugins/quay/src/components/QuayTagDetails/component.tsx
```diff
@@ -0,0 +1,115 @@
...
+        <>
+          {row.Name}
+          {row.Link.trim().length > 0 ? (
+            <a
+              href={getVulnerabilityLink(row.Link)}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              <Link style={{ marginLeft: '5px' }} />
+            </a>
```
tumido, line 38: Use `Link` from `@backstage/core-components` instead of `a` 

https://github.com/backstage/backstage/blob/master/packages/core-components/src/components/Link/Link.tsx

https://backstage.io/storybook/?path=/story/navigation-link--pass-props

tumido (changes requested):
Hey! Good progress. As an initial pass I've skimmed it through. I don't think we need `zustand` or `tanstack` in here. Can you please explain me the benefits of using those in this case?

packages/app/package.json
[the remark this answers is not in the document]
RobotSail: Oh I didn't mean to commit this, my bad.

packages/backend/package.json
RobotSail: Didn't mean to commit this, my bad.

plugins/quay/src/components/QuayTagPage/component.tsx
RobotSail: This comment was just adding to the code debt

plugins/quay/src/hooks/state.ts
RobotSail: We don't

plugins/quay/src/types.ts
RobotSail: Done

RobotSail:
Hi @serenamarie125, the `Link` icon is there because I was replicating what the new Quay UI does for its tag security page. It's just an icon that you can click on to be directed to the relevant advisory page. 

The severity icon is off because I didn't know whether it was wanted and didn't want to invest the time styling it until I knew. I can fix it now if you'd prefer it though.

RobotSail:
@serenamarie125 I just realized that I didn't answer your first question. Clicking the security scan link takes you to the `TagDetailsPage` which I've shown in the second screenshot.

RobotSail:
@serenamarie125 I fixed the icon styling, the vulnerabilities page now looks like this:
image

Do we want to also add the ability to expand each row to reveal more details about the vulnerability, like how the updated Quay UI currently does it?
If so, should we do that in this PR or another one?

RobotSail:
I did some more work on this and set it up locally so that you can expand each row to display some information. Is this something we want to include? 

### Row Closed
image

### Row Opened
image

The downside to this is that it lacks pagination. It could probably be added with some more effort, but again, I wanted to make sure this is something we want before doing that.

RobotSail:
Here is a screenshot of what the QuayTagPage looks like in the current state of the PR:

image

plugins/quay/src/components/QuayDashboardPage/QuayDashboardPage.tsx
```diff
@@ -1,21 +1,6 @@
...
-  const repository = info.join('/');
-
-  return (
-    <QuayRepository
-      organization={organization}
-      repository={repository}
-      widget={false}
-    />
-  );
+  return <QuayRepository />;
```
tumido, line 5: Thinking about this.. do we need a new component that just renders a single component - basically just reexports it? What about we change the router line (https://github.com/janus-idp/backstage-plugins/pull/240/files#diff-d2fefa7438680cfd44aad206e463e9058a17ea3282f2ca483ea4033a41497806R22 ) to use the `QuayRepository` directly? WDYT?

plugins/quay/src/components/QuayTagPage/component.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const QuayTagPage: React.FC<QuayTagPageProps> = () => {
+  const rootLink = useRouteRef(rootRouteRef);
+  const { repository, organization } = useRepository();
+  const { digest } = useParams();
+  if (!digest) {
+    throw new Error('digest is not defined');
+  }
+  const { loading, value } = useTagDetails(organization, repository, digest);
+  if (loading) {
+    return <div>loading...</div>;
```
tumido, line 20: Can you display `<Progress />` here instead, please? https://backstage.io/storybook/?path=/story/feedback-progress--progress

plugins/quay/src/hooks/quay.tsx
```diff
@@ -0,0 +1,128 @@
...
+    setTags(prevTags => [...prevTags, ...tagsResponse.tags]);
+    return tagsResponse;
+  });
+
+  const data = useMemo(() => {
+    return Object.values(tags)?.map(tag => {
+      const hashFunc = tag.manifest_digest.substring(0, 6);
+      const shortHash = tag.manifest_digest.substring(7, 19);
+      // console.log('tag.securityDetails', tag.securityDetails);
+      console.log('tag', tag);
```
tumido, line 66: Please remove this

tumido (changes requested):
Just a few nits, otherwise LGTM, good job! 👍

tumido (approved):
/lgtm
/approve

Issue #235, opened 2023-04-05
Unify UX on how different plugins schedule workers/tasks


tumido:
**Description**
In each backend plugin we host in Janus-IDP, we offer a different mean to schedule workers. Not only when it comes how and where this is configured, but also how they are scheduled.

See and compare configuration for following plugins:
- 3Scale https://github.com/janus-idp/backstage-plugins/tree/main/plugins/3scale-backend#configure
- Keycloak https://github.com/janus-idp/backstage-plugins/tree/main/plugins/keycloak-backend#installation
- OCM https://github.com/janus-idp/backstage-plugins/tree/main/plugins/ocm#setup-janus-idpbackstage-plugin-ocm-backend

**Expected behavior**

I think 3Scale plugins provides the easiest method to configure the scheduler, please migrate the other plugins to this format:

```yaml
# app-config.yaml

catalog:
  providers:
    <PLUGIN_NAME>:
      <ENV>:
        schedule: # optional; same options as in TaskScheduleDefinition
          # supports cron, ISO duration, "human duration" as used in code
          frequency: { minutes: 1 }
          # supports ISO duration, "human duration" as used in code
          timeout: { minutes: 1 }
```

**Acceptance Criteria**

- [x] Keycloak supports scheduler config through `app-config.yaml`
- [x] OCM supports scheduler config through `app-config.yaml`

SamoKopecky:
I can take a look at this.

PR #246, opened 2023-04-11
feat(ocm-backend)!: add a scheduler configuration option

SamoKopecky:
BREAKING CHANGE :boom:

Part of #235 

The scheduler for the entity provider is now configurable by changing the `app-config.yaml` or by changing code in `catalog.ts`. The old configuration for entity provider in `catalog.ts` is no longer valid.

The configuration options are described in the updated README file.


SamoKopecky:
Can you please take a look when you have the time @tumido :slightly_smiling_face:

PR #247, opened 2023-04-12
feat(keycloak): Add support for scheduler in config file

SamoKopecky:
Part of #235 

The scheduler for the entity provider is now configurable by changing the `app-config.yaml` or by changing code in `catalog.ts`. Since this plugin almost already supported all of the scheduler configuration methods, just one code change was needed and an updated README.

### DIFF
--- a/plugins/keycloak-backend/README.md
+++ b/plugins/keycloak-backend/README.md
@@ -33,52 +33,96 @@ When using client credentials, the access type must be set to `confidential` and
 
 ## Installation
 
-The package must be installed into the backend of Backstage. When not integrating with a published package, the repository must be cloned locally and added to Backstage as shown below:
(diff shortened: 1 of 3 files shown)


christophe-f:
This is great. Just want to make sure that we could still pass the `initialDelay` parameter

SamoKopecky:
> This is great. Just want to make sure that we could still pass the `initialDelay` parameter

I'll test it out and let you know.

tumido (changes requested):
Can you please update `config.d.ts` schema with the new values please?

SamoKopecky:
> Can you please update `config.d.ts` schema with the new values please?

Will do.

tumido (approved):
/lgtm
/approve

Issue #251, opened 2023-04-13
[Topology] Move base pf styles to plugin from component


invincibleJai:
**Describe the bug**
Move base pf styles to plugin from component, this can create styling issues if multiple plugins has base styles loaded on component

PR #252, opened 2023-04-13
fix(topology): moves base styles to plugin form component

invincibleJai:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/251


debsmita1:
verified the changes locally
LGTM!

Issue #237, opened 2023-04-05
Tekton: Filter by Cluster and error handling 


invincibleJai:
**Description:**
Currently, only the first cluster resources are being visualized in tekton but as a user, I would like to be able to filter resources by cluster in case of multiple clusters and sort the list based on columns.

**Acceptance Criteria:**
- [ ] Sort by columns for the list
- [ ] Add a dropdown cluster filter in the toolbar
- [ ]  On selecting a cluster from the filter only the resources of that cluster should be shown
- [ ]  Show error (if any while fetching clusters/resources) similar to k8s plugin

PR #253, opened 2023-04-17
feat(tekton): tekton pipeline run list sort

invincibleJai:
Tracks : https://github.com/janus-idp/backstage-plugins/issues/237

**Screenshots:**

- Pipeline Run List with data

image

- Empty state (will show warning if unable to fetch resource else just empty state)

image


debsmita1:
do we disable the dropdown if there's only one cluster ?

invincibleJai:
> do we disable the dropdown if there's only one cluster ?

we have similar experience in topology as well, we can take it up as a separate PR

debsmita1:
Are we not showing the "Pipeline Runs" heading anymore when the list is not empty ?

debsmita1:
verified the changes locally.
LGTM !

Issue #257, opened 2023-04-17
[Topology] On clicking a node in topology sometimes the sidebar doesn't open


divyanshiGupta:
**Describe the bug**
On clicking a node in topology sometimes the sidebar doesn't open.

PR #258, opened 2023-04-17
fix(topology): Get selected-node if data-model available

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/257

During resource fetch, loaded is set to false till the time the call completes so during that time selected node was not getting set as loaded remained false. I have removed loaded from the check as it is not needed since we just want to have the data-model in place and the selected-node should be set when its available.

https://user-images.githubusercontent.com/20724543/232550733-accff3f4-fe3f-40df-a19f-38d8f00801e9.mov


invincibleJai (approved):
LGTM

Issue #160, opened 2023-03-01
Keycloak-backend plugin does not implement pagination


gtema:
**Describe the bug**
Keycloak catalog plugin apparently does not implement pagination. It leads to only first 100 (or whatever is used as server default) users are being imported into Backstage

**To Reproduce**
1. enable KC plugin
2. have more then 100 users in KC
3. See only first 100 users are imported into backstage

**Expected behavior**
All users imported

PR #259, opened 2023-04-17
feat(keycloak): add support for user and group pagination

schultzp2020:
## Description

- Removes pagination limit of returning the first 100 users.
- Removes pagination limit of returning the first 100 groups.

Pagination is ignored if you specify [`max`](https://www.keycloak.org/docs-api/15.0/rest-api/index.html#_users_resource) with a [negative value](https://github.com/keycloak/keycloak/blob/main/server-spi-private/src/main/java/org/keycloak/utils/StreamsUtil.java#L64).

## Fixes

Fixes https://github.com/janus-idp/backstage-plugins/issues/160


sabre1041 (approved):
LGTM

gorkem (approved):
lgtm

christophe-f (approved):
LGTM.

plugins/keycloak-backend/README.md
```diff
@@ -13,7 +13,7 @@ The following capabilities are included within this plugin:
 
 Communication between Backstage and Keycloak is facilitated through the Keycloak API. Support is available for using username/password or client credentials.
 
-The following table describes the parameters that can configured in the `app-config.yaml` file to enable the plugin:
+The following table describes the parameters that can configured under `catalog.providers.keycloakOrg.default` in the `app-config.yaml` file to enable the plugin:
```
tumido, line 16:
```suggestion
The following table describes the parameters that can configured under `catalog.providers.keycloakOrg.<ENVIRONMENT_NAME>` in the `app-config.yaml` file to enable the plugin:
```

plugins/keycloak-backend/README.md
```diff
@@ -54,33 +61,43 @@ catalog:
         clientSecret: ${KEYCLOAK_CLIENTSECRET}
 ```
 
+Optionally, you can override the default parameters by adding the basic configuration to `app-config.yaml`.
+
+```yaml
+keycloak:
+  userQuerySize: 500
+  groupQuerySize: 250
+```
```
tumido, line 70: I don't like that we're introducing another place where the keycloak is configured. Why can't the `userQuerySize` and `groupQuerySize` live together with the rest of the config in `catalog.providers.keycloakOrg.<env>`

plugins/keycloak-backend/README.md
```diff
@@ -54,33 +61,43 @@ catalog:
...
+  env: PluginEnvironment,
+): Promise<Router> {
+  const builder = await CatalogBuilder.create(env);
+
+  builder.addEntityProvider(
+    KeycloakOrgEntityProvider.fromConfig(env.config, {
+      id: 'development',
+      logger: env.logger,
+      userQuerySize: env.config.getOptionalNumber('keycloak.userQuerySize'),
+      groupQuerySize: env.config.getOptionalNumber('keycloak.groupQuerySize'),
```
tumido, line 88: why is this not consumed from `env.config` within the provider and is passed here separately? The config object is already available to the provider via `KeycloakOrgEntityProvider.fromConfig(env.config, ...`

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -66,11 +72,17 @@ export const parseUser = (
   spec: {
     profile: {
       email: user.email,
-      displayName: [user.firstName, user.lastName].filter(Boolean).join(' '),
+      displayName:
+        user.firstName || user.lastName
+          ? `${user.firstName} ${user.lastName}`
+          : user.username,
```
tumido, line 78: How is this related to the merit of this PR? Besides it introduces a regression. If only one of `firstName` or `lastName` is defined, the display name starts/ends with a leading space.

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -66,11 +72,17 @@ export const parseUser = (
...
     },
-    memberOf: groups
-      .filter(g => g.members?.includes(user.username!))
-      .map(g => g.name!),
+    memberOf: groups.reduce((acc, g) => {
+      if (g.members?.includes(user.username!)) {
+        acc.push(g.name!);
+      }
+      return acc;
+    }, [] as string[]),
```
tumido, line 85: How is this related to the merit of this PR? It just makes the code less readable. Performance improvements are really unmeasurable here.

tumido (changes requested):
See comments inline. I think this should not create another new configuration point for the plugin. It should be part of the provider config. Besides, the newly introduced additional config values are missing from the config schema.

plugins/keycloak-backend/src/lib/read.ts
schultzp2020: @christophe-f and I found another bug. Currently, if a user does not specify a `firstName` and `lastName`, the field will remain empty; therefore, backstage will not inject the entity into the catalog. I will refactor the snippet to not have the regression; otherwise, I can create another pr if needed.

plugins/keycloak-backend/src/lib/read.ts
schultzp2020: I will remove this change.

plugins/keycloak-backend/README.md
```diff
@@ -76,55 +78,68 @@ When using client credentials, the access type must be set to `confidential` and
...
+        )
         ...
-        }
+      }
       ```
 
    2. Add a schedule directly inside the `packages/backend/src/plugins/catalog.ts` file:
 
-      ```diff
+      ```ts
```
tumido, line 103: How is this helpful change? User needs to know which lines to add. Not all are added to enable this plugin. Please respect consistent formatting across our and upstream plugins.

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -85,20 +91,61 @@ export function* traverseGroups(
   }
 }
 
+export async function getEntities<T extends Users | Groups>(
+  entities: T,
```
tumido, line 95: This is clever, I like it.

tumido (changes requested):
The new parameters are still missing from the schema. Please update the `config.d.ts`.

plugins/keycloak-backend/README.md
schultzp2020: The codeblock was a `ts` codeblock before the most recent keycloak [pr](https://github.com/janus-idp/backstage-plugins/commit/dcf345f8f0ca27f85cb4bf158a1d223cb706bb60#diff-74b70d63dd0c250dcfdd7cacf4639fc82a1a66098d9fc2a269bb98bf89790f9fL59). I can change this back to a `diff` codeblock.

schultzp2020:
@tumido @christophe-f I think this ready to merge.

plugins/keycloak-backend/README.md
tumido: Yes, it was. However, the block contained `+ ` start of lines, it was a `diff` from the beginning. Please change to `diff`.

tumido (approved):
/lgtm
/approve

PR #268, opened 2023-04-18
test(ocm-backend): Migrate tests from nock to msw

SamoKopecky:
Migrate tests for `kubernetes.ts` and `router.ts` to use `msw` instead of `nock` since `msw` is used backstage wide for mocking. We can now also utilize the defined handlers that are used in the `ocm-backend` dev setup.


tumido (approved):
Love it ❤️ This is much simpler 👍

PR #273, opened 2023-04-19
test(ocm-backend): Add missing cluster to a router test

SamoKopecky:
SSIA

The workflow on the latest `keycloak` commit probably needs to be rerun since it failed because of this. Sorry about that.


tumido:
sorry, didn't noticed this one until I've merged #278 🤦 

I'm sorry.. 🤦

Issue #274, opened 2023-04-19
Keycloak plugin does not inject entities into backstage when missing properties


schultzp2020:
**Describe the bug**

Keycloak plugin does not inject entities into backstage when missing properties

**To Reproduce**

1. Create a user without a `firstName` and `lastName`
2. Start Backstage
3. Look at the User/Group entity catalog

**Expected behavior**

Keycloak should still inject entities into Backstage when missing properties

**Screenshots**

**Additional context**

We cannot reasonably expect all users of this plugin to have correctly setup Keycloak.

schultzp2020:
@navinyashende for some reason GH is not letting me assign the issue to you.

PR #275, opened 2023-04-19
feat(keycloak): stricter keycloak entity parsing

schultzp2020:
## Description

- Remove non null assertions in parsing to fix https://github.com/janus-idp/backstage-plugins/issues/274

## Fixes

Fixes https://github.com/janus-idp/backstage-plugins/issues/274

### DIFF
--- a/plugins/keycloak-backend/src/lib/read.ts
+++ b/plugins/keycloak-backend/src/lib/read.ts
@@ -29,50 +29,99 @@ interface GroupRepresentationWithParent extends GroupRepresentation {
 export const parseGroup = (
   keycloakGroup: GroupRepresentationWithParent,
   realm: string,
-): GroupEntity => ({
-  apiVersion: 'backstage.io/v1beta1',
-  kind: 'Group',
-  metadata: {
-    name: keycloakGroup.name!,
-    annotations: {
-      [KEYCLOAK_ID_ANNOTATION]: keycloakGroup.id!,
-      [KEYCLOAK_REALM_ANNOTATION]: realm,
+): GroupEntity => {
+  if (!keycloakGroup.id) {
+    throw new Error(`Group id is required: ${JSON.stringify(keycloakGroup)}`);
+  }
+
+  if (!keycloakGroup.name) {
+    throw new Error(`Group name is required for group id: ${keycloakGroup.id}`);
+  }
+
+  return {
+    apiVersion: 'backstage.io/v1beta1',
+    kind: 'Group',
+    metadata: {
+      name: keycloakGroup.name,
+      annotations: {
+        [KEYCLOAK_ID_ANNOTATION]: keycloakGroup.id,
+        [KEYCLOAK_REALM_ANNOTATION]: realm,
+      },
     },
-  },
-  spec: {
-    type: 'group',
-    profile: {
-      displayName: keycloakGroup.name!,
+    spec: {
+      type: 'group',
+      profile: {
+        displayName: keycloakGroup.name,
+      },
+      children:
+        keycloakGroup.subGroups?.reduce((acc, g) => {
+          if (!g.name) {
+            throw new Error(`Group name is required: ${JSON.stringify(g)}`);
+          }
+
+          acc.push(g.name);
+          return acc;
+        }, [] as string[]) || [],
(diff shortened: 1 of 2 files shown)


plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -29,50 +29,99 @@ interface GroupRepresentationWithParent extends GroupRepresentation {
...
+              return str;
+            }
+
+            const trimmed = word.trim();
+            if (trimmed.length === 0) {
+              return str;
+            }
+
+            return str ? `${str} ${word.trim()}` : word.trim();
+          }) || user.username,
```
tumido, line 109: Please read backstage documentation on display name. It should not be defined if you can't provide better display string than `metadata.name`.

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -29,50 +29,99 @@ interface GroupRepresentationWithParent extends GroupRepresentation {
...
+            if (!word) {
+              return str;
+            }
+
+            const trimmed = word.trim();
+            if (trimmed.length === 0) {
+              return str;
+            }
+
+            return str ? `${str} ${word.trim()}` : word.trim();
```
tumido, line 108: Why to complicate things this way? What does it solve? The code is much less understandable, less readable, and definitely not maintainable. You're reducing over 2 field array, even if/elseif/else would be more comprehensive here.

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -29,50 +29,99 @@ interface GroupRepresentationWithParent extends GroupRepresentation {
...
+          }) || user.username,
+      },
+      memberOf: groups.reduce((acc, g) => {
+        // user.username is always defined
+        if (g.members?.includes(user.username!)) {
+          if (!g.name) {
+            throw new Error(`Group name is required: ${JSON.stringify(g)}`);
+          }
+
```
tumido, line 117: This can never happen. Group can't exist in Keycloak without a name. Besides, throwing error here stops all ingestion at this point. So a single membership error (That can never happen) would stop ingestion of the whole user object, instead of skipping the single membership. That is not a safe approach.

tumido (changes requested):
I'm sorry, but I don't see this PR to be helpful in any meaningful way.

1. In Keycloak, group always has a name and id. There's no way a group can exist in Keycloak without that. 
2. Having display name for user or group in Backstage is not required and in case you don't have a better string than `metadata.name` than it should not be defined. If display name is not defined, backstage uses `metadata.name` for display purposes. This 
means your display name changes are irrelevant.
3. Throwing errors from entity ingestor results in the entity being rejected. I don't think a single invalid membership error, should result into all users of that group to be not ingested to Backstage.

schultzp2020:
> I'm sorry, but I don't see this PR to be helpful in any meaningful way.

I am trying to fix this [issue](https://github.com/janus-idp/backstage-plugins/issues/274).

> 1. In Keycloak, group always has a name and id. There's no way a group can exist in Keycloak without that.

Good to know, I wish the typescript types would accurately reflect that :/. Maybe its worth checking or creating an issue on `@keycloak/keycloak-admin-client` to have more accurate types.

> 2. Having display name for user or group in Backstage is not required and in case you don't have a better string than `metadata.name` than it should not be defined. If display name is not defined, backstage uses `metadata.name` for display purposes. This means your display name changes are irrelevant.

This does not seem to be the case. When testing, the entity will not be injected into the backstage entity catalog if `displayName` is left [as is](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/src/lib/read.ts#L69).

After investigating further, I think the error comes from Backstage not liking that `displayName` always returns string. In this case, an empty string because everything is undefined. I think if we change `displayName` to be `undefined` instead of an empty string it will work. I need to do further testing to verify.

> 3. Throwing errors from entity ingestor results in the entity being rejected. I don't think a single invalid membership error, should result into all users of that group to be not ingested to Backstage.

Sounds good, I will remove the errors.

tumido:
@schultzp2020 since you're trying to fix a reproducable issue, can you please write a few testcases for it? It will help us understand the issue and prevent regressions.

Issue #286, opened 2023-04-21
[Topology] update pf react-core package


debsmita1:
**Describe the bug**
Upgrade react-core package in topology plugin

PR #287, opened 2023-04-21
[Topology] upgrade pf react-core package

debsmita1:
Fixes https://github.com/janus-idp/backstage-plugins/issues/286


divyanshiGupta:
@debsmita1 The build is failing. 

```
plugins/topology/dev/index.tsx:31:7 - error TS2420: Class 'MockKubernetesClient' incorrectly implements interface 'KubernetesApi'.
  Property 'proxy' is missing in type 'MockKubernetesClient' but required in type 'KubernetesApi'.
```

Please fix the issue and update the PR.

plugins/topology/package.json
```diff
@@ -35,7 +35,7 @@
     "@material-ui/lab": "^4.0.0-alpha.45",
     "@patternfly/patternfly": "4.224.2",
     "@patternfly/react-charts": "6.94.18",
-    "@patternfly/react-core": "4.276.6",
+    "@patternfly/react-core": "^4.276.8",
```
divyanshiGupta, line 38: @debsmita1 do we want to automatically update this package to future minor/patch versions? IMO that can be risky as sometimes it might include breaking changes. @invincibleJai wdyt?

PR #292, opened 2023-04-25
feat(k8s-actions): add create namespace action

### LINKED ISSUE #175: Custom action: Create a Kubernetes namespace
## Description
As a Platform Engineer, I want to be able to use a custom action to **create a Kubernetes namespace** in my GPT.

SamoKopecky:
Resolves #175 

- Add a new package for kubernetes actions with version `v1.0.0` and set it to public
- Add a custom scaffolder action to create a Kubernetes repository using the kubernetes client API library
- Add documentation for the created action
- Add tests for the created action

The `name` parameter that is used in the kube config values (like `cluster`, `name` ...) can be anything as far as I am aware since I think its only used for the kube config description itself and not for the API calls. 

Note: Checks for required parameters are not done in code since they part of the schema and backstage won't run the actions unless it has all of the required parameters.

### DIFF
--- a/plugins/kubernetes-actions/README.md
+++ b/plugins/kubernetes-actions/README.md
@@ -1,5 +1,125 @@
-# @janus-idp/backstage-plugin-scaffolder-backend-module-kubernetes
+# Kubernetes actions for Backstage
 
-The kubernetes module for [@backstage/plugin-scaffolder-backend](https://www.npmjs.com/package/@backstage/plugin-scaffolder-backend).
+This module provides [Backstage](https://backstage.io/) template [actions](https://backstage.io/docs/features/software-templates/builtin-actions) for [Kubernetes](https://kubernetes.io/docs/home/).
 
-_This plugin was created through the Backstage CLI_
+The following actions are currently supported in this module:
+
+- Create a kubernetes namespace
+
+## Getting started
+
+1. Install the action package in your Backstage project
+
+   ```bash
+   yarn workspace backend add @janus-idp/backstage-scaffolder-backend-module-kubernetes
+   ```
+
+2. [Register](https://backstage.io/docs/features/software-templates/writing-custom-actions#registering-custom-actions) the Kubenretes actions by modifying the `packages/backend/src/plugins/scaffolder.ts` file from your project with the following changes:
+
+   ```ts
+   import { CatalogClient } from '@backstage/catalog-client';
+   import {
+     createBuiltinActions,
+     createRouter,
+   } from '@backstage/plugin-scaffolder-backend';
+   import { ScmIntegrations } from '@backstage/integration';
+   import { Router } from 'express';
+   import type { PluginEnvironment } from '../types';
+   import { createKubernetesNamespaceAction } from '@janus-idp/backstage-scaffolder-backend-module-kubernetes';
+
+   export default async function createPlugin(
+     env: PluginEnvironment,
+   ): Promise<Router> {
+     const catalogClient = new CatalogClient({
+       discoveryApi: env.discovery,
+     });
+
+     const integrations = ScmIntegrations.fromConfig(env.config);
+
+     const builtInActions = createBuiltinActions({
+       integrations,
+       catalogClient,
+       config: env.config,
+       reader: env.reader,
+     });
+
+     const actions = [
+       ...builtInActions,
+       createKubernetesNamespaceAction(catalogClient),
+     ];
+
+     return await createRouter({
+       actions,
+       logger: env.logger,
+       config: env.config,
+       database: env.database,
+       reader: env.reader,
+       catalogClient,
+       identity: env.identity,
+     });
+   }
+   ```
+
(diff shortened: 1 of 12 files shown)


tumido (changes requested):
For a technical nit, see inline.

On a different topic:

I don't think we should expect the user to know the target cluster API URL or understand what is the local cluster. Instead, we should:
1. Either be able to consume this data from the Backstage Catalog (we can safely assume all clusters are stored as `kind: Resource` with `spec.type: kubernetes-cluster`) and offer those as a dropdown in the scaffolder form as entity references. The action should recognize it and pull the appropriate API URL annotation from the catalog.
2. Additionally we can provide value for *Other* which would conditionally allow the user to specify the API URL directly.


I know most of the above is handled on the Template side and not within the action itself, however, I think we should provide means to support these usecases - currently, it would be really challenging without changes in the action itself.

WDYT?

Additionally, I don't think you should provide `localCluster` flag since the user typically has no understanding of that (if Backstage is deployed on Kubernetes at all).

SamoKopecky:
> For a technical nit, see inline.
> 
> On a different topic:
> 
> I don't think we should expect the user to know the target cluster API URL or understand what is the local cluster. Instead, we should:
> 
>     1. Either be able to consume this data from the Backstage Catalog (we can safely assume all clusters are stored as `kind: Resource` with `spec.type: kubernetes-cluster`) and offer those as a dropdown in the scaffolder form as entity references. The action should recognize it and pull the appropriate API URL annotation from the catalog.
> 
>     2. Additionally we can provide value for _Other_ which would conditionally allow the user to specify the API URL directly.
> 
> 
> I know most of the above is handled on the Template side and not within the action itself, however, I think we should provide means to support these usecases - currently, it would be really challenging without changes in the action itself.
> 
> WDYT?
> 
> Additionally, I don't think you should provide `localCluster` flag since the user typically has no understanding of that (if Backstage is deployed on Kubernetes at all).

I think that's a great idea and much user friendly then the current solution. I'll update the action.

plugins/kubernetes-actions/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@kubernetes/client-node": "^0.18.1"
+  },
+  "devDependencies": {
+    "@backstage/backend-common": "^0.18.0",
+    "@backstage/cli": "^0.22.1",
+    "msw": "^1.2.1"
+  },
+  "files": [
+    "dist"
+  ]
```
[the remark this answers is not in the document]
SamoKopecky: Resoled in the latest commit.

SamoKopecky:
The action now supports either using a catalog cluster of an API url.

plugins/kubernetes-actions/src/actions/createKubernetesNamespace.ts
```diff
@@ -0,0 +1,183 @@
...
+export function createKubernetesNamespaceAction(catalogClient: CatalogClient) {
+  return createTemplateAction<TemplateActionParameters>({
+    id: 'kubernetes:create-namespace',
+    description: 'Creates a kubernetes namespace',
+    schema: {
+      input: {
+        type: 'object',
+        required: ['namespaceName', 'token'],
+        properties: {
+          namespaceName: {
```
tumido, line 85: can we rename this to `namespace` please?

plugins/kubernetes-actions/src/actions/createKubernetesNamespace.ts
```diff
@@ -0,0 +1,183 @@
...
+};
+
+export function createKubernetesNamespaceAction(catalogClient: CatalogClient) {
+  return createTemplateAction<TemplateActionParameters>({
+    id: 'kubernetes:create-namespace',
+    description: 'Creates a kubernetes namespace',
+    schema: {
+      input: {
+        type: 'object',
+        required: ['namespaceName', 'token'],
```
tumido, line 83: Should also contain `oneOf`:
https://json-schema.org/understanding-json-schema/reference/combining.html?highlight=oneof#oneof

```
oneOf: [{required: ['url']}, {required: ['clusterRef']} ]
```

tumido (changes requested):

SamoKopecky:
Thanks for the feedback, changes applied in the latest commits.

plugins/kubernetes-actions/README.md
```diff
@@ -0,0 +1,125 @@
...
+   ```
+
+4. Add the Kubernetes actions to your templates, see the [example](./examples/templates/01-kubernetes-template.yaml) file in this repository for complete usage examples
+
+   ```yaml
+   action: kubernetes:create-namespace
+   id: create-kubernetes-namespace
+   name: Create kubernetes namespace
+   input:
+     namespaceName: foo
```
tumido, line 76: Please update to reflect the previously discussed changes.

tumido (changes requested):
A little nit for the README. Otherwise LGTM

Issue #270, opened 2023-04-19
App Topology: Show newly added workload's specific details/resources in sidepanel


divyanshiGupta:
**Description:**

As a developer, I want to see SS,DS,CJ,J workloads specific details/resources (if any) in sidepanel similar to devconsole topology node sidepanel.

**Acceptance Criteria:**

- [x] Add specific details/resources for SS
- [x] Add specific details/resources for DS
- [x] Add specific details/resources for CJ
- [x] Add specific details/resources for J

PR #295, opened 2023-04-26
feat(topology): add specific details for all supported workloads

divyanshiGupta:
**Story:** https://github.com/janus-idp/backstage-plugins/issues/270

Added specific detail items for SS/DS/CJ/J resources as per OCP. Also added Jobs section in resources tab of CJ resource.

https://github.com/janus-idp/backstage-plugins/assets/20724543/a9750051-5e22-405e-905a-46ce01c56067

Screenshot 2023-05-17 at 3 57 58 PM

### DIFF
--- a/plugins/topology/src/components/Topology/TopologySideBar/TopologyCronJobDetails.tsx
+++ b/plugins/topology/src/components/Topology/TopologySideBar/TopologyCronJobDetails.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react';
+import { V1CronJob } from '@kubernetes/client-node';
+import TopologySideBarDetailsItem from './TopologySideBarDetailsItem';
+import TopologyWorkloadDetails from './TopologyWorkloadDetails';
+import { Timestamp, TimestampFormat } from '@patternfly/react-core';
+
+const TopologyCronJobDetails: React.FC<{ resource: V1CronJob }> = ({
+  resource,
+}) => {
+  return (
+    <TopologyWorkloadDetails resource={resource}>
+      <TopologySideBarDetailsItem label="Schedule">
+        {resource.spec?.schedule}
+      </TopologySideBarDetailsItem>
+      <TopologySideBarDetailsItem label="Concurrency policy">
+        {resource.spec?.concurrencyPolicy}
+      </TopologySideBarDetailsItem>
+      <TopologySideBarDetailsItem
+        label="Starting deadline seconds"
+        emptyText="Not configured"
+      >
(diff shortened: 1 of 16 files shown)


plugins/topology/src/components/Topology/TopologySideBar/TopologyCronJobDetails.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <TopologyWorkloadDetails resource={resource}>
+      <TopologySideBarDetailsItem label="Schedule">
+        {resource.spec?.schedule}
+      </TopologySideBarDetailsItem>
+      <TopologySideBarDetailsItem label="Concurrency policy">
+        {resource.spec?.concurrencyPolicy}
+      </TopologySideBarDetailsItem>
+      <TopologySideBarDetailsItem label="Starting deadline seconds">
+        {resource.spec?.startingDeadlineSeconds
+          ? `${resource.spec.startingDeadlineSeconds} second`
```
invincibleJai, line 20: should we handle plural as well i.e second or seconds?

plugins/topology/src/components/Topology/TopologySideBar/TopologyCronJobDetails.tsx
```diff
@@ -0,0 +1,34 @@
...
+          : 'Not configured'}
+      </TopologySideBarDetailsItem>
+      <TopologySideBarDetailsItem label="Last schedule time">
+        <Timestamp
+          date={resource.status?.lastScheduleTime}
```
invincibleJai, line 25: should we render Timestamp if lastScheduleTime is defined?

plugins/topology/src/components/Topology/TopologySideBar/TopologyResourcesTabPanel.css
```diff
@@ -0,0 +1,3 @@
+.topology-job-pod-ring svg {
```
invincibleJai, line 1: I know we are not consistent with styles across plugins but we need to decide on how to handle it like scoping to plugin level i.e `bs-topology-*` . WDYT

cc @debsmita1

plugins/topology/src/components/Topology/TopologySideBar/TopologyCronJobDetails.tsx
divyanshiGupta: It is just `second` in ODC but handled it here.

invincibleJai:
LGTM, verified the changes.

PR #321, opened 2023-05-03
feat(segment): analytics plugin for Segment

gorkem:
Initial work . Set up the analytics plugin cleanup and update from the previous work.


schultzp2020 (changes requested):
Don't forget to run `yarn prettier --write ./plugins/analytics-provider-segment/`

plugins/analytics-provider-segment/src/apis/implementations/AnalyticsApi/Segment.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.analytics = new AnalyticsBrowser();
+    this.analytics.load({ writeKey: writeKey });
+
+  }
+
+  /**
+   * Instantiate a fully configured Segment API implementation.
+   */
+  static fromConfig(config: Config, identityApi?: IdentityApi) {
+    const writeKey = config.getString('app.analytics.segment.writeKey');
```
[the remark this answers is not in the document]
gorkem: Makes sense, this is the part I did not touch from the old PR

schultzp2020 (approved):
LGTM

plugins/analytics-provider-segment/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@janus-idp/plugin-analytics-provider-segment",
```
tumido, line 2: Please change to maintain naming convention

```suggestion
  "name": "@janus-idp/backstage-plugin-analytics-provider-segment",
```

or 

```suggestion
  "name": "@janus-idp/backstage-plugin-segment",
```

plugins/analytics-provider-segment/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "version": "0.1.0",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "license": "Apache-2.0",
+  "private": true,
```
tumido, line 7: Please remove

```suggestion
```

plugins/analytics-provider-segment/src/index.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2022 The Backstage Authors
```
tumido, line 2: I think this should read in all files

```suggestion
 * Copyright 2023 The Janus IDP Authors
```

tumido (commented):
I have just a few formal suggestions, see inline. Otherwise LGTM

plugins/analytics-provider-segment/package.json
gorkem: I think the original author just followed the convention that was used with the Google Analytics plugin upstream. I am OK with the change but we need to update the README etc too

plugins/analytics-provider-segment/src/index.ts
gorkem: We are picking up some of the code from a closed PR upstream. I am not sure if we should change the original copyright rather add Janus

tumido (approved):
/lgtm
/approve

PR #324, opened 2023-05-03
chore(monorepo): replace lerna with turbo

schultzp2020:
## Description

Please explain the changes you made here.

- Adds turborepo as the monorepo tool since lerna was not used in any scripts
  - Faster script runtime because of caching and parallelism
  - We also use turborepo on the showcase
- ~~Removes the `package` folder~~
- ~~Renames the `plugins` folder as `packages~~
  - ~~Add missing dependencies as dev dependencies in `package.json` to pass tests~~
  - ~~Inspiration from the [GitLab plugins](https://github.com/immobiliare/backstage-plugin-gitlab)~~

## Which issue(s) does this PR fix

- Blocked by https://github.com/janus-idp/backstage-plugins/pull/367

- Fixes build issues in the downstream


nickboldt:
Well, it takes over an hour, but it passed!

https://orch.psi.redhat.com/pnc-web/#/projects/1631/build-configs/11003/builds/AYQSUI34TRQAA

turbo.json
```diff
@@ -0,0 +1,20 @@
...
+      "cache": false,
+      "persistent": true
+    },
+    "build": {
+      "outputs": ["dist/**"]
+    },
+    "clean": {
+      "cache": false
+    },
+    "test": {
```
RobotSail, line 14: Needs to depend on `build`
```suggestion
    "test": {
      "dependsOn": ["^build"]
    }  
```

turbo.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "start": {
+      "cache": false,
+      "persistent": true
```
RobotSail, line 6: Needs to depend on build
```suggestion
      "persistent": true,
      "dependsOn": ["^build"]
```

app-config.yaml
RobotSail, line 1: Why are we deleting this file?

package.json
```diff
@@ -6,24 +6,20 @@
     "node": "16 || 18"
   },
   "scripts": {
-    "dev": "concurrently \"yarn start\" \"yarn start-backend\"",
-    "start": "yarn workspace app start",
-    "start-backend": "yarn workspace backend start",
-    "build": "backstage-cli repo build",
+    "start": "turbo run start",
+    "build": "turbo run build",
     "tsc": "tsc",
```
RobotSail, line 11: Is this part of `build`? 

If not, is it possible to break this out so that all other plugins also have a `tsc` step, and that way we can just replace this with `turbo tsc`?

.dockerignore
RobotSail, line 1: Why is this being deleted?

RobotSail (changes requested):
Looks okay to me for the most part apart from some minor changes. 
Would like to get @tumido to look at this as well before merging

.dockerignore
schultzp2020: There are no dockerfiles

app-config.yaml
schultzp2020: `app-config.yaml` is only needed for running an instance of backstage, not for the plugins.

package.json
schultzp2020: I can look into this, the reason I never looked into this is because TS tooling can be a pain in the butt.

.dockerignore
RobotSail: Gotcha

package.json
schultzp2020: I don't think its worth changing because it actually takes longer to build on average and it adds extra boilerplate to each plugin. 

Here are my averages after some testing: 
| Before | After  |
| ------- | ------ |
| 11.5s   | 13.1s  |

RobotSail: Okay, that's fine, we can just leave it as is then.

RobotSail (commented):
Turbo file looks good to me. 
Would like to hear @tumido's thoughts on these changes as well.

tumido (changes requested):
So far I have just a brief look at the PR. Overall I love it, I would love to make this change happen and embrace life without `packages/[backend,app]`. It would simplify our lifes a lot.

However as of now I have to block this change from happening from a simple reason. Without `packages/backend` we have no way to develop catalog based plugins. We have no way to verify ingestion into catalog. This essentially blocks any meaningfull development and validation on `keycloak`, `3scale` and `ocm` plugins. 

I'll try to come up with a way how we can stand up a standalone catalog dev environments for those plugins. Without that we can't remove catalog (`packages/backend`) from the repo.

I've filed this newly discovered limitaion as #330 issue.

nickboldt:
>  I have to block this change from happening

This means that for now I have to build RHPIB from Paul's fork (this PR)... not ideal for a GA, but OK for a dev preview. We will need a full solution before we can GA, so please treat this as a SEV1 / P1 / blocker (whatever terminology you want). :D

tumido:
I'll try to resolve this next week. I know this is important to get in.

rtaniwa:
hey @tumido . i think we may need to unblock this and merge, and then possibly address the other concerns later. i am worried that we are holding up our ability to GA RHPIB and we currently have no one available to pick up issue #330 unless you have other options.

RobotSail:
@rtaniwa I am currently working on issue #330

tumido (changes requested):
Please rebase properly. This PR diff includes code changes in one of the plugins, documentation changes in others, it's a mess. Also, please leave the `app-config*.yaml` files intact

tumido (approved):
/lgtm
/approve

tumido (approved):
/lgtm
/approve

PR #376, opened 2023-05-18
Create Standalone Dev Environments for Backstage Plugins Relying on Catalog

### LINKED ISSUE #330: Create a standalone dev environment for plugins that depend on catalog
While we already have a standalone dev environments backend plugins serving their own API endpoints, we're lacking a setup that would allow us to work against a Backstage catalog. Currently the only way how to see if entities are being properly discovered and ingested is to stand up `packages/backstage` and connect the plugin to it. This is not suitable if we want to isolate each plugin into separate dev environments.

~This currently blocks #324. If merged, we have no way to develop *keycloak*, *3scale*, *ocm*, plugins.~

- [x] Keycloak backend has a standalone dev environment - solved via https://github.com/janus-idp/backstage-plugins/pull/543
- [x] #600
- [x] #601
- [x] #602
- [x] #603

RobotSail:
Signed-off-by: Oleg S <97077423+RobotSail@users.noreply.github.com>
Fixes #330

This PR aims to introduce standalone dev environments that developers can use when working with plugins that rely on the backend catalog.


RobotSail:
@tumido This PR isn't review-ready yet, but could you please tell me whether or not this is the general format you had in mind? At the moment, I basically copied what we do from `packages/backend` into `plugins/ocm-backend/dev`. Based on our conversations, I'm guessing that we will want to do the same for the other few plugins relying on the catalog.

tumido:
Yes!☺️ This is heading in the right direction. This is precisely it!🙂 In the ideal case I would like to keep it as minimal as possible...

Like... In this case `ocm-backend` doesn't need a scaffolder or auth, so if we can just mock them or disregard completely the better.🙂

Also it would be nice if we can also provide the `app-config` via something like:

https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm-backend/dev/index.ts#L20

But I'm not sure how that would complicate things so whatever is easier/faster at this point. We can iterate.🙂

Overall this is the right format. I love it!🙌

RobotSail:
So in order for `backstage-cli` to properly run the dev environments for the plugins, the packages they're in must be located under the `packages` directory. I've tried a number of solutions to get it working under `plugins`, but it doesn't seem to work at all. 
Another solution I can think of would be to make changes to the upstream `backstage-cli` behavior to allow for this. I haven't looked into this one so I'm not sure how easy or feasible this would be.

Issue #377, opened 2023-05-19
App Topology: Show Route decorator for workloads visualised in topology if Route resource exists along with updates on sidepanel


invincibleJai:
**Description**
Currently topology plugin visualize k8s resources like Deployment, Pods, Statefulsets, jobs and k8s workload, and decorator is shown only if an associated ingress resource is found with the supported workload. If OpenShift specific Route is associated with the workloads then it should be shown in route decorator and associated details should be shown in sidebar as well 

https://backstage.io/docs/features/kubernetes/configuration/#role-based-access-control

**Acceptance Criteria**

- [x] Show associated route as decorator
- [x] Show associated route in sidepanel
- [x] If a workload is associated with ingress and route both then in decorator show Route and both should be shown on sidepanel

PR #378, opened 2023-05-19
feat(topology): visualize Route resource in decorator and sidebar for supported workloads

invincibleJai:
**Fixes:** https://github.com/janus-idp/backstage-plugins/issues/377

**Description:**
Currently topology plugin visualize k8s resources like Deployment, Pods, Statefulsets, jobs and k8s workload, and decorator is shown only if an associated ingress resource is found with the supported workload. If OpenShift specific Route is associated with the workloads then it should be shown in route decorator and associated details should be shown in sidebar as well

**Screenshots:**
image

**Setup:**
- If OpenShfit Route is associated with any workload then it will be visualized in topology as decorator and sidebar.
-  Route is provided as part of the configuration 

```
---
kubernetes:
  customResources:
    - group: 'route.openshift.io'
      apiVersion: 'v1'
      plural: 'routes'
```

- and has the [`ClusterRole`](https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control) must be granted to `ServiceAccount` accessing the cluster

```
---
kind: ClusterRole
metadata:
  name: backstage-read-only
rules:
  ...
  - apiGroups:
      - route.openshift.io
    resources:
      - routes
    verbs:
      - get
      - list
```

**TODO:**

- [x]  Show associated route as decorator
- [x]  If a workload is associated with ingress and route both then in the decorator show Route and both should be shown on sidepanel
- [x]  Show associated route in sidepanel
- [x] Add/update specs


plugins/topology/src/components/Topology/TopologySideBar/TopologyResourcesTabPanel.tsx
```diff
@@ -1,21 +1,39 @@
...
+      return (
+        <>
+          <RouteListSidebar routesData={routesData} />
+          <IngressListSidebar ingressesData={ingressesData} />
+        </>
+      );
+    } else if (hasRoutesData && !hasIngressData) {
+      return <RouteListSidebar routesData={routesData} />;
+    } else if (hasIngressData && !hasRoutesData) {
+      return <IngressListSidebar ingressesData={ingressesData} />;
```
divyanshiGupta, line 33:
```suggestion
    } 
```
We dont need this as if there is no routeData then anyways we are showing ingress-list which will show the ingress resources if available otherwise empty state.

plugins/topology/src/components/Topology/TopologySideBar/Resources/IngressListSidebar.tsx
```diff
@@ -0,0 +1,52 @@
...
+            {ingressData.ingress.spec?.rules?.length && (
+              <IngressRules ingress={ingressData.ingress} />
+            )}
+          </li>
+        ))}
+    </TopologyResourcesTabPanelItem>
+  );
+};
+
+export default IngressListSidebar;
```
divyanshiGupta, line 52: Can we use this comp for the routes list as well instead of having another comp, since both the comps are mostly similar except for the resource?

divyanshiGupta (changes requested):

plugins/topology/src/components/Topology/TopologySideBar/Resources/IngressListSidebar.tsx
invincibleJai: have kept it separate as ingress has `IngressRules` as well which is rendered conditionally based on `ingressData.ingress.spec?.rules?.length` which is specific to ingress

divyanshiGupta: We don't have rules in Route? Also we can just have a boolean like showRules set to true in case of ingress. Wdyt?

invincibleJai: no route doesn’t have rules like ingress.
We can do that but that will not be much of a value add and I feel good to keep as separate unless you feel strongly otherwise.

divyanshiGupta:
@invincibleJai Please resolve conflicts.

plugins/topology/README.md
```diff
@@ -23,6 +23,37 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
 - The Kubernetes plugins including `@backstage/plugin-kubernetes` and `@backstage/plugin-kubernetes-backend` are installed and configured by following the [installation](https://backstage.io/docs/features/kubernetes/installation) and [configuration](https://backstage.io/docs/features/kubernetes/configuration) guides.
 - The Kubernetes plugin is configured and connects to the cluster using a `ServiceAccount`.
 - The [`ClusterRole`](https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control) must be granted to `ServiceAccount` accessing the cluster. If you have the Backstage Kubernetes plugin configured, then the `ClusterRole` is already granted.
+- To Visualize OpenShift Route, it should be added in `customResources` component in the [`app-config.yaml`](https://backstage.io/docs/features/kubernetes/configuration#configuring-kubernetes-clusters) file:
```
invincibleJai, line 26: @hmanwani-rh / @tumido / @serenamarie125 Have updated README as well for support of OpenShift Route PTAL

divyanshiGupta (approved):

plugins/topology/README.md
hmanwani-rh, line 26:
```suggestion
- The following must be added in`customResources` component in the [`app-config.yaml`](https://backstage.io/docs/features/kubernetes/configuration#configuring-kubernetes-clusters) file to view the OpenShift route:
```

hmanwani-rh (commented):
@invincibleJai Added a couple of suggestions. 

The rest looks good.

hmanwani-rh (approved):

Issue #306, opened 2023-04-27
Topology plugin: user docs


rkratky:
Develop user documentation for the Topology plugin.

Inspiration: https://roadie.io/docs/integrations/argocd/

PR #381, opened 2023-05-22
docs(Topology): updates user docs

hmanwani-rh:
GitHub issue: https://github.com/janus-idp/backstage-plugins/issues/306

cc @rkratky


invincibleJai (approved):
Thanks @hmanwani-rh Looks good.

have added some nit's, see if it makes sense and can be taken as follow up PR as well

tumido (approved):
/lgtm
/approve

Issue #368, opened 2023-05-17
[ocm] Update plugin to match UX review requirements


tumido:
As a result of https://github.com/janus-idp/backstage-plugins/issues/107 we should make the OCM plugin adopt following UI changes:

### Clusters page

![image](https://github.com/janus-idp/backstage-plugins/assets/7453394/4df83abe-8672-4962-86fb-6609a54fcc7c)

### ClusterInfoCard

![image](https://github.com/janus-idp/backstage-plugins/assets/7453394/da37485b-0f43-4a76-9bfd-65980ce31d8d)

## Acceptance criteria

- [x] Update `ocm-backend` to surface required data about nodes
- [x] Update *ClustersStatusPage* page to match the mockup
- [x] Remove *ClusterAllocatableResourcesCard*
- [x] Remove *ClusterStatusCard*
- [x] Update *ClusterInfoCard* to match the mockup

PR #352, opened 2023-05-13
feat(ocm-backend): add additional information to status endpoint

SamoKopecky:
Part of #368 

Add platform, node list and update information to the status endpoint for the `ocm-backend` plugin. Also add tests for new functionality.

### More detailed changes:
- Add additional information to the `ocm` API
- Add a new k8s `api` function all to get all cluster infos at once since the update and node information is present in the `managedclusterinfos` endpoint
- Add a function to parse node information from a cluster info object
- Update and platform information is parsed using already implemented funtions

The `/api/ocm/status` call in this PR:
```json
[
  {
    "name": "foo",
    "status": {
      "available": true,
      "reason": "Managed cluster is available"
    },
    "platform": "BareMetal",
    "openshiftVersion": "4.10.26",
    "nodes": [
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      }
    ],
    "update": {
      "available": true,
      "version": "4.10.51",
      "url": "https://access.redhat.com/errata/RHSA-2023:0561"
    }
  },
  {
    "name": "cluster1",
    "status": {
      "available": true,
      "reason": "Managed cluster is available"
    },
    "platform": "BareMetal",
    "openshiftVersion": "4.9.21",
    "nodes": [
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      }
    ],
    "update": {
      "available": true,
      "version": "4.10.51",
      "url": "https://access.redhat.com/errata/RHSA-2023:0561"
    }
  },
  {
    "name": "offline-cluster",
    "status": {
      "available": false,
      "reason": "Managed cluster is unavailable"
    },
    "platform": "BareMetal",
    "openshiftVersion": "4.9.21",
    "nodes": [
      {
        "status": "Unknown",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "Unknown",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
        "type": "Ready"
      },
      {
        "status": "True",
[...]

PR #386, opened 2023-05-24
feat(ocm)!: remake OCM UX

SamoKopecky:
**WIP**

Rebased to #352 

BREAKING CHANGE :boom: 

Part of #368 

Remake most of OCM UX, this includes:
- Create shared components `Status` and `Update`
- Use `patternfly` icons instead of `materialui` icons for the update indicator
- Convert `ClusterStatusPage` to a table and include shared `Status` and `Update` components, introduce the node information and add provider column
- Remove `ClusterStatusCard`
- Remove `ClusterAllocatableResourceCard`
- Remake `ClusterInfoCard`, remove links from the table, include shared `Status` component to replace the old `ClusterStatusCard` and use `Update` Component

image
image


plugins/ocm-backend/src/service/router.ts
```diff
@@ -101,11 +107,24 @@ export async function createRouter(
...
+                c.hubResourceName,
+              ),
+              status: parseClusterStatus(mc),
+              platform: getClaim(mc, 'platform.open-cluster-management.io'),
+              openshiftVersion:
+                mc.metadata!.labels?.openshiftVersion ||
+                getClaim(mc, 'version.openshift.io'),
+              nodes: parseNodeStatus(mcis.items[index]),
+              ...parseUpdateInfo(mcis.items[index]),
+            } as ClusterOverview),
```
tumido, line 126: This won't work. You can't be sure the api response for `ManagedCluster`s list and `ManagedClusterInfo`s list have always all the matching resources and in the same order. You can be easily introducing a regression in case cluster is being deleted/added into ACM.

Since kube api calls are not atomic and each of those list calls happen in different time you can get a result where `ManagedCluster`s call returns you

```
items: [cluster-A, cluster-B]
```

and `ManagedClusterInfo`s returns you

```
items [cluster-A, cluster-a, cluster-B]
```

In this case matching via `index` on `map` in not stable.

tumido (changes requested):
Can we please split this into a Backend PR (backend + common) and separate Frontend PR? (maybe move it to the https://github.com/janus-idp/backstage-plugins/pull/352 )

Also one regression introduced in the stitching code for cluster infos, see inline.

plugins/ocm-backend/src/service/router.ts
tumido: Disregard my PR, I can't read. Sorry. All this feedback should have landed in https://github.com/janus-idp/backstage-plugins/pull/352 🤦

tumido (approved):
/lgtm
/approve

PR #388, opened 2023-05-24
fix(3scale): don't expose token and use correct config

dweber019:
As far as I can see only `accessToken` is valid and should be a secret https://github.com/janus-idp/backstage-plugins/blob/main/plugins/3scale-backend/src/providers/config.ts#L26


tumido (approved):
/lgtm
/approve

/cc @fmenesesg

Issue #389, opened 2023-05-25
[Tekton] Prevent continuous re-rendering in the Tekton plugin


debsmita1:
**Describe the issue:**
The PipelineRunList page and the CI/CD tab with the latest pipelinerun visualization keeps re-rendering

![latest](https://github.com/janus-idp/backstage-plugins/assets/22490998/6a49a766-0e48-4b88-8111-6b28bba786f7)

PR #391, opened 2023-05-26
fix(tekton): handles unnecessary rerenders on tekton

invincibleJai:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/389


debsmita1:
verified the changes
LGTM

Issue #398, opened 2023-05-29
Topology sidepanel not closing on outside click


divyanshiGupta:
**Describe the bug**
Sidepanel is not closing on clicking outside of the node. 

**Expected behavior**
It should get closed on clicking anywhere outside of the node.

**Screenshots**

https://github.com/janus-idp/backstage-plugins/assets/20724543/eb904e21-1ef1-4294-a0eb-0faadbe513f5

PR #399, opened 2023-05-29
fix(topology): close workload sidepanel on outside-click

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/398

This PR fixes the issue with workload sidepanel not closing on outside-click. For the sidepanel to be closed, selected-id should be removed from the URL but currently it was only getting removed when there was no id returned by the selection event listener hence it was failing when some other type of topology node was selected for ex - the graph node or app node. Added a check to remove the selected-id if the the selected node is not of type workload.

https://github.com/janus-idp/backstage-plugins/assets/20724543/99130d89-c171-442c-8a42-ca2091c35d66


debsmita1:
verified the changes
LGTM

PR #402, opened 2023-05-30
chore(ci): enable local turbo cache

tumido:
Adds a step in `test` and `build` workflows which sets up the local turbo cache using Github Cache via https://github.com/dtinth/setup-github-actions-caching-for-turbo

Also cleans up node version parsing, since this is now supported out of the box via `actions/setup-node` via `node-version-file`.


schultzp2020 (approved):
LGTM

PR #404, opened 2023-05-31
fix(ocm): handle missing nodeList properly

tumido:
In testing the `3.0.0` in live environment I've noticed it can happen the `nodeList` is not present in the `ManagedClusterInfo`s `.status`...

In such case then OCM page crashes because it tries to do `.map()` on `undefined`.

As an example is a Jerry cluster on OperateFirst:

https://console-openshift-console.apps.moc-infra.massopen.cloud/k8s/ns/jerry/internal.open-cluster-management.io~v1beta1~ManagedClusterInfo/jerry/yaml

### DIFF
--- a/plugins/ocm-backend/src/helpers/parser.test.ts
+++ b/plugins/ocm-backend/src/helpers/parser.test.ts
@@ -275,13 +275,28 @@ describe('parseNodeStatus', () => {
     ]);
   });
 
-  it('should should return an empty array if no nodes are present', () => {
+  it('should should return an empty array if nodes are empty', () => {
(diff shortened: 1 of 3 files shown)


plugins/ocm-backend/src/helpers/parser.test.ts
```diff
@@ -275,7 +275,7 @@ describe('parseNodeStatus', () => {
     ]);
   });
 
-  it('should should return an empty array if no nodes are present', () => {
+  it('should should return an empty array if nodes are empty', () => {
```
invincibleJai, line 278: nit:

```suggestion
  it('should return an empty array if nodes are empty', () => {
```

schultzp2020 (approved):
LGTM

Issue #387, opened 2023-05-24
Add Prettier Config Sorting


RobotSail:
It would be nice to have a prettier config setup to automatically sort imports so that way we can have something like:
```ts
// react imports
// backstage imports

// third party imports

// janus idp imports

// builtin imports

// local imports
```

PR #419, opened 2023-05-31
style: reorder imports via prettier

tumido:
Resolves https://github.com/janus-idp/backstage-plugins/issues/387

Adds https://github.com/IanVS/prettier-plugin-sort-imports prettier plugin and configures it according to the issue by @RobotSail

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,16 @@
+module.exports = {
+  ...require('@spotify/prettier-config'),
+  importOrder: [
+    '^react(.*)$',
+    '',
+    '^@backstage/(.*)$',
+    '',
+    '<THIRD_PARTY_MODULES>',
+    '',
+    '^@janus-idp/(.*)$',
+    '',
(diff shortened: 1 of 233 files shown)


RobotSail (approved):
Yes, this is perfect! That's the exact plugin I was thinking of using. 
LGTM!

.prettierrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  ...require('@spotify/prettier-config'),
+  importOrder: [
+    '^react(.*)$',
+    '',
```
schultzp2020, line 5: The empty new lines should be removed

schultzp2020 (approved):

.prettierrc.js
tumido: They are added intentionally, see the docs here:

https://github.com/IanVS/prettier-plugin-sort-imports#3-add-spaces-between-import-groups

PR #424, opened 2023-06-01
style: update prettier config

schultzp2020:
## Changes

- [Arrow Function Parentheses](https://prettier.io/docs/en/options.html#arrow-function-parentheses)
- [Print Width](https://prettier.io/docs/en/options.html#print-width)
- [Single Quotes](https://prettier.io/docs/en/options.html#quotes)
- [Trailing Commas](https://prettier.io/docs/en/options.html#trailing-commas)


.prettierrc.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
   ...require('@spotify/prettier-config'),
+  arrowParens: 'always',
+  printWidth: 100,
+  singleQuote: true,
+  trailingComma: 'all',
```
tumido, line 6: These are already covered by parent `@spotify/prettier-config`

https://github.com/spotify/web-scripts/blob/master/packages/prettier-config/index.js

This brings no value.

tumido (commented):
This PR changes 4 prettier values. Out of these 4:
- 2 are set to the same value as in the parent config
- 1 is in direct contrast with the upstream parent config without any explanation on the benefit for deviating from config used in the upstream project
- 1 is additional deviation (not from parent config but from prettier defaults) without any explanation.

Can you please provide more detail on why are these changes beneficial to this project and why it would be good for us to deviate from Backstage upstream?

PR #432, opened 2023-06-06
feat(keycloak): add transformation for username

tanoggy:
- When Keycloak is configured with an external identity provider like Google, the username will be set as the associated email address.
- Backstage now enforces restrictions, allowing only alphanumeric characters for usernames.

### DIFF
--- a/plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.ts
+++ b/plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.ts
@@ -215,7 +215,23 @@ export class KeycloakOrgEntityProvider implements EntityProvider {
 
     await this.connection.applyMutation({
       type: 'full',
-      entities: [...users, ...groups].map(entity => ({
+      entities: [
+        ...users.map(obj => {
+          const existingUsernames = users.map(user => user.metadata.name);
+          let counter = 1;
+          const name = obj.metadata.name
+            .replace(/[^a-zA-Z0-9@-]/g, '-')
+            .replace(/@/g, '');
+          let username = name;
+
+          while (existingUsernames.includes(username)) {
+            username = `${name}-${counter}`;
+            counter++;
+          }
+          return { ...obj, metadata: { ...obj.metadata, name: username } };
+        }),
+        ...groups,
+      ].map(entity => ({
         locationKey: `keycloak-org-provider:${this.options.id}`,
         entity: withLocations(provider.baseUrl, provider.realm, entity),
       })),


tanoggy:
hi @yzhao583, please check my commit

tumido (changes requested):
Hi @tanoggy, thank you for your contribution! 🙂 

We should really think this through. It has great security implications. This is also a breaking change since all the already imported `Users` will stop working.

Can you please try addressing my inline comments? Your change needs to move to a different file and please remove the version change.

cc @gorkem @sabre1041 @christophe-f PTAL

tumido:
@tanoggy can you also please update your commit message and PR title to `feat(keycloak): add transformation for username` instead? The scope of this PR is not `pencil` (we don't have any pencils here) and you're not attempting any validation, instead your mutating or transforming the username 🙂

tumido (changes requested):
Hi @tanoggy,

yes, the commit message and PR title is now much better. However the rest of the conversations are not resolved at this point. Please kindly do not manually request a review + mark conversations as resolved unless you actually make the requested code changes. Please be mindful of my notifications inbox. 🙂 

As of now, the codebase is the same as before so my feedback still stands.

plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.ts
```diff
@@ -151,13 +151,13 @@ export class KeycloakOrgEntityProvider implements EntityProvider {
...
-      logger: Logger;
-      userTransformer?: UserTransformer;
-      groupTransformer?: GroupTransformer;
-    },
+      private options: {
+        id: string;
+        provider: KeycloakProviderConfig;
+        logger: Logger;
+        userTransformer?: UserTransformer;
+        groupTransformer?: GroupTransformer;
```
tumido, line 159: @riginoommen had a good point when discussing this PR offline. Instead of forming an opinion on instance-specific transformation requirements, we should provide a customization piece (additionally we can provide a sane default for the most common cases).

We already have these transformers here available, however they are never used (dead code). Instead of the transformation below, can you please propagate these two transformers to the `parseUser` and `parseGroup` functions in https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/src/lib/read.ts and let them transform the data?

tanoggy: I think we can a customization piece to env example using regex

tanoggy (commented):
Run yarn prettier

plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.ts
tumido: I don't think we should force a regex transformation on user, instead offer an optional default and allow them to actually apply `userTransformer` and `groupTransformer`. Please redo this PR according to my previous comment.

Please do NOT modifying `entities` in `src/providers/KeycloakOrgEntityProvider.ts`. Instead export a default `userTransformer`/`userTransformer` AND propagate it to `src/lib/read.ts` so it gets applied.

tumido:
Hi @tanoggy do you plan on working on this to satisfy our requirements as described in https://github.com/janus-idp/backstage-plugins/pull/432#discussion_r1242131743 or should we try to move this forward ourselves based on your work? 🙂

jani888:
Hey @tumido,

Just wanted to reach out and say we're really loving this plugin you're maintaining - it's doing a great job!  But, like some others, we're also running into this particular issue.

As a temporary fix, we've applied a patch using [patch-package](https://github.com/ds300/patch-package), which should do until the official fix is merged and we can implement a more permanent solution.

For the time being, we're using the Keycloak ID as the username. When this issue is fixed, we thought about making this change with the userTransformer, but it's a bit of a heavy lift since we'd have to rewrite the whole parseUser method just to tweak this one thing.

So, we're curious - do you have any thoughts or tips on how we could make this adjustment in a simpler way? Any advice would be much appreciated.

Also, we're more than willing to roll up our sleeves and contribute if there's any way we can be of help. Let us know what you think!

Stay awesome! 🔥

tumido:
Hi @jani888,

thanks for your kind words and support! 🙂 

I'll take a look at this and raise a new PR since the original author here seems away for the time being.

It's not about making the adjustment simpler, but rather at the right place. If we mutate entities during `applyMutation`, this happens after all the other entities were parsed. This can result in inconsistency in relations for `.spec.memberOf` in case of User entity and `.spec.parent`, `.spec.members`, `.spec.children` for Group entity. This is not trivial.

I'll try to create a PR that should hopefully handle those scenarios better and keep you posted.

PR #434, opened 2023-06-07
chore(plugins): move Plugins from `plugins` to `packages`

RobotSail:
This PR relocates all of the plugins from `plugins` to `packages`.

The reason for this PR is so that we can properly create dev environments for #330.
Since `backstage-cli` doesn't allow something to run unless it is located in the `packages` 
directory.


RobotSail:
@tumido It looks like this PR opens up a number of code-smells. Would it be possible to merge this PR and then create a follow-up PR to resolve all of the code-smell issues? My concern is that resolving them in a PR where we're already moving directories will be too much change to keep track of

package.json
```diff
@@ -15,6 +15,7 @@
     "build:plugins": "turbo run build --filter=...{./plugins/*}",
```
schultzp2020, line 15: You will need to update the build scripts. I would suggest using a scope filter. https://turbo.build/repo/docs/core-concepts/monorepos/filtering#scopes

schultzp2020 (changes requested):

tumido:
@RobotSail I'm sorry, I won't be merging this PR at this point, we would need to live with it for now.

Merging this after the feature freeze will make my life hell when backporting possible hardening and other fixes into `bundle-1.0` branch.

That being said, I'm happy to leave this to live as a PR, I'm not dismissing this change in any way, It's just not good timing. 🙂

RobotSail:
@tumido That's fine, we'll just icebox this for now. Let me know when you'd like me to look at this again.

package.json
RobotSail: Good catch, I'll update that & check for other references to the plugins directory.

PR #436, opened 2023-06-07
fix(monorepo): update test script

schultzp2020:
The plugin test scripts needs to have the optional args `--passWithNoTests --coverage` so they can be run at the root level with `yarn test`.


tumido (approved):
/lgtm
/approve

PR #437, opened 2023-06-07
test(keycloak): add tests

schultzp2020:
image

### DIFF
--- a/plugins/keycloak-backend/__fixtures__/helpers.ts
+++ b/plugins/keycloak-backend/__fixtures__/helpers.ts
@@ -1,0 +1,40 @@
+import { EntityProviderConnection } from '@backstage/plugin-catalog-backend';
+
+import KeycloakAdminClient from '@keycloak/keycloak-admin-client';
+
+import { groupMembers, groups, users } from './data';
+
+export const BASIC_VALID_CONFIG = {
+  catalog: {
+    providers: {
+      keycloakOrg: {
+        default: {
+          baseUrl: 'http://localhost:8080/auth',
+        },
+      },
+    },
+  },
+} as const;
+
+export const client = {
+  auth: jest.fn().mockResolvedValue({}),
+  users: {
+    find: jest.fn().mockResolvedValue(users),
+    count: jest.fn().mockResolvedValue(users.length),
+  },
+  groups: {
+    find: jest.fn().mockResolvedValue(groups),
+    count: jest.fn().mockResolvedValue(groups.length),
+    listMembers: jest
+      .fn()
+      .mockResolvedValueOnce(groupMembers[0])
+      .mockResolvedValueOnce(groupMembers[1])
+      .mockResolvedValueOnce(groupMembers[2])
+      .mockResolvedValueOnce(groupMembers[3]),
+  },
+} as unknown as KeycloakAdminClient;
+
+export const connection = {
+  applyMutation: jest.fn(),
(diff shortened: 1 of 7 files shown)


tanoggy:
skip test

schultzp2020:
The remaining sonarcloud bugs and hotspots are false positives.

plugins/keycloak-backend/src/lib/read.test.ts
```diff
@@ -0,0 +1,19 @@
+import KcAdminClient from '@keycloak/keycloak-admin-client';
+
+import { KeycloakAdminClientMock } from '../../__fixtures__/helpers';
+import { KeycloakProviderConfig } from './config';
+import { readKeycloakRealm } from './read';
+
+const config = {
+  realm: 'myrealm',
+} as unknown as KeycloakProviderConfig;
```
invincibleJai, line 9: nit: do we need to case as unknown? Can directly do `as KeycloakProviderConfig` or better to add required type with mock

```
const config: KeycloakProviderConfig = {
  realm: 'myrealm',
  id: 'mock_id',
  baseUrl: 'http://mock-url',
};
```

plugins/keycloak-backend/__fixtures__/helpers.ts
```diff
@@ -0,0 +1,44 @@
+import { EntityProviderConnection } from '@backstage/plugin-catalog-backend';
```
invincibleJai, line 1: @deprecated — import from @backstage/plugin-catalog-node instead

plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const keycloak = KeycloakOrgEntityProvider.fromConfig(config, {
+      id: 'development',
+      logger: getVoidLogger(),
+      schedule: 'manual',
+    });
+
+    const result = keycloak.map(async k => await k.connect(connection));
+
+    expect(result).toHaveLength(1);
```
invincibleJai, line 82: `keycloak.map(async k => await k.connect(connection));` will return list of promises `[ Promise { <pending> } ]` and length of result will always be same as keycloak

and if await for promises to be resolved it gives `[undefined]`

plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  it('should not read without a connection', async () => {
```
invincibleJai, line 85:
```suggestion
  it('should not read without a connection', () => {
```

plugins/keycloak-backend/__fixtures__/helpers.ts
schultzp2020: https://github.com/janus-idp/backstage-plugins/issues/445

plugins/keycloak-backend/src/lib/read.test.ts
schultzp2020: Sure, I'll update the object with the required types since there are only 3 of them.

invincibleJai (approved):
LGTM

tumido (approved):
/lgtm
/approve

PR #452, opened 2023-06-13
chore(monorepo): update README and remove concurrency arg

schultzp2020:
## Changes

- Updates the README to account for turbo filtering
- Removes concurrency arg from the root `package.json`


tumido (approved):
/lgtm
/approve

Issue #238, opened 2023-04-05
Migrate backend plugins to the new backend system


tumido:
**Is your feature request related to a problem? Please describe.**
Currently, Janus IDP backend plugins are using the legacy backend plugin system. We should investigate if/how to migrate to the new plugin system.

**Describe the solution you'd like**
Backend plugins support both the legacy and the new backend plugin systems.

**Describe alternatives you've considered**
If we can't support both backend plugins systems, the migration should happen only when the system is already available in a released Backstage version

**Acceptance criteria**

- [ ] OCM plugin supports both plugin systems
- [ ] Keycloak plugin supports both plugin systems
- [ ] 3Scale plugin supports both plugin systems

PR #459, opened 2023-06-14
feat(ocm-backend): add support for new backend system

SamoKopecky:
Part of #238

### DIFF
--- a/plugins/ocm-backend/src/helpers/kubernetes.ts
+++ b/plugins/ocm-backend/src/helpers/kubernetes.ts
@@ -1,3 +1,5 @@
+import { LoggerService } from '@backstage/backend-plugin-api';
+
 import {
   CustomObjectsApi,
   KubeConfig,
@@ -11,7 +13,7 @@ import { ManagedCluster, ManagedClusterInfo, OcmConfig } from '../types';
 
 export const hubApiClient = (
   clusterConfig: OcmConfig,
-  logger: Logger,
+  logger: Logger | LoggerService,
 ): CustomObjectsApi => {
   const kubeConfig = new KubeConfig();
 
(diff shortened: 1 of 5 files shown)


plugins/ocm-backend/src/helpers/kubernetes.ts
```diff
@@ -11,7 +13,7 @@ import { ManagedCluster, ManagedClusterInfo, OcmConfig } from '../types';
 
 export const hubApiClient = (
   clusterConfig: OcmConfig,
-  logger: Logger,
+  logger: Logger | LoggerService,
```
RobotSail, line 16: the `Logger` type and `LoggerService` are two different ideas, so this should remain as `Logger`. Since these are two distinct types, it will create errors at runtime if we follow this definition. `LoggerService` should somehow give us `Logger` here as happens in the main Backstage backend

plugins/ocm-backend/src/service/router.ts
```diff
@@ -45,30 +50,20 @@ import {
...
-  logger: Logger;
-  config: Config;
-}
-
-export async function createRouter(
-  options: RouterOptions,
-): Promise<express.Router> {
-  const { logger } = options;
-  const { config } = options;
+const buildRouter = (config: Config, logger: Logger | LoggerService) => {
```
RobotSail, line 53: We shouldn't be type unions here unless there is a really good reason to. This avoids complexity and prevents messy code going forward.

RobotSail (changes requested):
This PR is a good start. Only thing to be changed is that we'll need to ensure the logger type is either `Logger` or `LoggerService` instead of it being a union type.

tumido:
Closing in favor of #682

SamoKopecky:
Sorry, I saw the notification about the comments and wanted to look at this soon, thanks for reusing the PR. :slightly_smiling_face:

RobotSail:
@SamoKopecky No worries at all. Thank you for creating the PR!

Issue #395, opened 2023-05-26
App Topology: Show pod logs for associated container


invincibleJai:
**Description:** As a user I would like to see the logs for pods associated with any k8s workload

**Acceptance Criteria:**

- [x]  Should show `view logs`  in the resources tab for workloads in the pods section for sidepanel
- [x]  on click on view logs, should open model/drawer with logs for the selected pod
- [x]  in the log component show a filter to search and should show a dropdown to select any container for pod

**Screenshots:**

- OCP 

<img width="1790" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/fbd17b62-6579-432f-b8bc-af6cf2eebbe2">

- k8s plugin

<img width="1672" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/24d43a8b-737d-4b30-a3bc-52eb03099ce7">

_Check with PM, about how the logs should be visualized_

Spike result: https://github.com/janus-idp/backstage-plugins/issues/271#issuecomment-1564388387

invincibleJai:
@serenamarie125 / @christophe-f PTAL when you get time

PR #464, opened 2023-06-15
feat(topology): support for logviewer for pods based on container

invincibleJai:
**Tracks:** https://github.com/janus-idp/backstage-plugins/issues/395

**Description:** Show the logs for pods associated with any k8s workload

**Screenshots/Gif**: 

image

image

Light Theme:

image

**TODO:**

- [x]  Should show `view logs`  in the resources tab for workloads in the pods section for sidepanel
- [x] On click on view logs, should open a modal with logs for the selected pod
- [x]  In the log component show a filter to search and should show a dropdown to select any container for pod
- [x] UI refinement (handle error and edge cases scenarios)
- [x]  Add Unit tests

image

### DIFF
--- a/plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogs.test.tsx
+++ b/plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogs.test.tsx
@@ -1,0 +1,67 @@
+import React from 'react';
+
+import { render } from '@testing-library/react';
+
+import { usePodLogs } from '../../../../hooks/usePodLogs';
+import { PodLogs } from './PodLogs';
+
+jest.mock('../../../../hooks/usePodLogs', () => ({
+  usePodLogs: jest.fn(),
+}));
+
+jest.mock('@backstage/core-components', () => ({
+  ...jest.requireActual('@backstage/core-components'),
+  LogViewer: () => <div data-testid="log-viewer">Log viewer</div>,
+  DismissableBanner: () => (
+    <div data-testid="pod-log-banner">Dismissable banner</div>
+  ),
+}));
+
+describe('PodLogs', () => {
+  it('should render logviewer', () => {
+    const podScopeData = {
+      podName: 'node-ex-git-er56',
+      podNamespace: 'sample-app',
+      containerName: 'node-ex-git',
+      clusterName: 'OCP',
+    };
+    (usePodLogs as jest.Mock).mockReturnValue({
+      loading: false,
+      value: 'log data...',
+    });
+    const { queryByTestId } = render(<PodLogs podScope={podScopeData} />);
+    expect(queryByTestId('pod-log-banner')).not.toBeInTheDocument();
+    expect(queryByTestId('log-viewer')).toBeInTheDocument();
+  });
+
+  it('should render logviewer skeleton', () => {
+    const podScopeData = {
+      podName: 'node-ex-git-er56',
+      podNamespace: 'sample-app',
+      containerName: 'node-ex-git',
+      clusterName: 'OCP',
+    };
+    (usePodLogs as jest.Mock).mockReturnValue({ loading: true });
+    const { queryByTestId } = render(<PodLogs podScope={podScopeData} />);
+    expect(queryByTestId('pod-log-banner')).not.toBeInTheDocument();
+    expect(queryByTestId('log-viewer')).not.toBeInTheDocument();
+    expect(queryByTestId('logs-skeleton')).toBeInTheDocument();
+  });
+
+  it('should render DismissableBanner in case of error', () => {
+    const podScopeData = {
+      podName: 'node-ex-git-er56',
+      podNamespace: 'sample-app',
+      containerName: 'node-ex-git',
+      clusterName: 'OCP',
+    };
+    (usePodLogs as jest.Mock).mockReturnValue({
+      loading: false,
+      error: { message: 'some crash!!' },
+    });
+    const { queryByTestId } = render(<PodLogs podScope={podScopeData} />);
+    expect(queryByTestId('pod-log-banner')).toBeInTheDocument();
+    expect(queryByTestId('log-viewer')).not.toBeInTheDocument();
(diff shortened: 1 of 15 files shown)


plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogs.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    (usePodLogs as jest.Mock).mockReturnValue({
+      loading: false,
+      value: 'log data...',
+    });
+    const { queryByTestId } = render(<PodLogs podScope={podScopeData} />);
+    expect(queryByTestId('pod-log-banner')).not.toBeInTheDocument();
+    expect(queryByTestId('log-viewer')).toBeInTheDocument();
+  });
+
+  it('should render logviewer skelton', () => {
```
divyanshiGupta, line 37: nit:
```suggestion
  it('should render logviewer skeleton', () => {
```

plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogs.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { usePodLogs } from '../../../../hooks/usePodLogs';
+import { ContainerScope } from './types';
+
+type PodLogsProps = {
+  podScope: ContainerScope;
+};
+
+export const PodLogs = ({ podScope }: PodLogsProps) => {
+  const [loadingData, setLoadingData] = React.useState(true);
```
divyanshiGupta, line 16:
```suggestion
  const [loadingData, setLoadingData] = React.useState<boolean>(true);
```

plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogs.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { value, error, loading } = usePodLogs({
+    podScope: podScope,
+  });
+
+  React.useEffect(() => {
+    if (!loading && loadingData) {
+      setLoadingData(false);
+    }
+  }, [loading, loadingData]);
```
divyanshiGupta, line 25: Why can we not use `loading` itself instead of updating another variable based on its value and then using it?

plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogsDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+);
+
+type PodLogsDialogProps = {
+  podData: V1Pod;
+};
+
+export const PodLogsDialog = ({ podData }: PodLogsDialogProps) => {
+  const { clusters, selectedCluster } = React.useContext(K8sResourcesContext);
+  const classes = useStyles();
+  const [open, setOpen] = useState(false);
```
divyanshiGupta, line 48:
```suggestion
  const [open, setOpen] = useState<boolean>(false);
```

plugins/topology/src/hooks/useK8sObjectsResponse.test.ts
```diff
@@ -44,4 +44,61 @@ describe('useK8sObjectsResponse', () => {
     expect(result.current.clusters).toEqual(['minikube']);
     expect(result.current.selectedClusterErrors).toEqual([]);
   });
+
+  it('should return k8sResourcesContextData with empty clusters if it does not exists', () => {
```
divyanshiGupta, line 48: nit:
```suggestion
  it('should return k8sResourcesContextData with empty clusters if it does not exist', () => {
```

plugins/topology/src/hooks/usePodLogs.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { usePodLogs } from './usePodLogs';
+
+jest.mock('@backstage/core-plugin-api', () => ({
+  ...jest.requireActual('@backstage/core-plugin-api'),
+  useApi: jest.fn(),
+}));
+
+describe('usePodLogs', () => {
+  it('should return loading as true and value undefined initially', () => {
```
divyanshiGupta, line 13:
```suggestion
  it('should return loading as true and value as undefined initially', () => {
```

plugins/topology/src/hooks/usePodLogs.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      }),
+    );
+
+    waitForNextUpdate();
+
+    expect(result.current.loading).toEqual(true);
+    expect(result.current.value).toBeUndefined();
+  });
+
+  it('should return loading as false and log text', async () => {
```
divyanshiGupta, line 35:
```suggestion
  it('should return loading as false and value as log text', async () => {
```

divyanshiGupta (changes requested):

plugins/topology/src/components/Topology/TopologySideBar/PodLogs/PodLogs.tsx
invincibleJai: you are right it's clean to move this usePodsLogs and the caller can rely on loading but the local state is needed as am making calls in intervals (used `useAsyncRetry`) and for each call loading turns to true and then false

plugins/topology/src/hooks/usePodLogs.ts
```diff
@@ -0,0 +1,42 @@
...
+    [getLogs],
+  );
+
+  useInterval(() => retry(), intervalMs);
+
+  React.useEffect(() => {
+    if (!loading && loadingData) {
+      setLoadingData(false);
+    }
+  }, [loading, loadingData]);
```
divyanshiGupta, line 39: @invincibleJai its not advised to use the state variable and its state update function in same useEffect as it might lead to an infinite loop. I understand the value will only be updated when `loadingData` is true but still wanted to mention this here. LMK wdyt.

divyanshiGupta (changes requested):

debsmita1:
verified this locally
LGTM

Issue #486, opened 2023-06-26
When authmiddleware is setup for any backstage-plugin, project will not authenticate


pbuehl-redhat:
**Describe the bug**
Projects utilizing the quay backstage plugins will not work with authmiddleware set up

**To Reproduce**

1. Go to 'packages/backend/index.ts'
2. Set up apirouter for one of the plugins i.e. 'apiRouter.use('/quay', authMiddleware, await quay(quayEnv));'
3. run backstage instance and see that authentication will fail

**Expected behavior**
Projects will be allowed to authenticate when using authmiddleware.

tumido:
Hi @pbuehl-redhat thank you for noticing this, filing an issue and even opening a pull request. 👍  That's huge help. I think this is an important issue to fix. 🙂

tumido:
@pbuehl-redhat please ignore my comment above. I finally did my homework 🙂 

https://github.com/janus-idp/backstage-plugins/pull/498#discussion_r1273380327

All is good, your PR is good.

pbuehl-redhat:
thanks @tumido

PR #498, opened 2023-06-29
fix(plugins): add bearer token to request headers when sending requests, authmiddleware issue

pbuehl-redhat:
…iddleware issue

### DIFF
--- a/plugins/quay/src/api/index.ts
+++ b/plugins/quay/src/api/index.ts
@@ -2,6 +2,7 @@ import {
   ConfigApi,
   createApiRef,
   DiscoveryApi,
+  IdentityApi,
 } from '@backstage/core-plugin-api';
 
 import {
@@ -40,6 +41,7 @@ export const quayApiRef = createApiRef<QuayApiV1>({
 export type Options = {
   discoveryApi: DiscoveryApi;
   configApi: ConfigApi;
+  identityApi: IdentityApi;
 };
 
 export class QuayApiClient implements QuayApiV1 {
@@ -48,9 +50,12 @@ export class QuayApiClient implements QuayApiV1 {
 
   private readonly configApi: ConfigApi;
 
+  private readonly identityApi: IdentityApi;
+
   constructor(options: Options) {
     this.discoveryApi = options.discoveryApi;
     this.configApi = options.configApi;
+    this.identityApi = options.identityApi;
   }
 
   private async getBaseUrl() {
@@ -60,8 +65,12 @@ export class QuayApiClient implements QuayApiV1 {
   }
 
   private async fetcher(url: string) {
+    const credentials = await this.identityApi.getCredentials();
     const response = await fetch(url, {
-      headers: { 'Content-Type': 'application/json' },
+      headers: { 
+        'Content-Type': 'application/json',
+        'Authorization': `Bearer ${credentials.token!!}`
+      },
     });
     if (!response.ok) {
       throw new Error(
--- a/plugins/quay/src/api/index.ts
+++ b/plugins/quay/src/api/index.ts
@@ -65,11 +65,11 @@ export class QuayApiClient implements QuayApiV1 {
   }
 
   private async fetcher(url: string) {
-    const credentials = await this.identityApi.getCredentials();
+    const {token: idToken} = await this.identityApi.getCredentials();
     const response = await fetch(url, {
       headers: {
         'Content-Type': 'application/json',
-        Authorization: `Bearer ${credentials.token!!}`,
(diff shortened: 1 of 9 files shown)


plugins/quay/src/api/index.ts
```diff
@@ -60,8 +65,12 @@ export class QuayApiClient implements QuayApiV1 {
   }
 
   private async fetcher(url: string) {
+    const {token: idToken} = await this.identityApi.getCredentials();
     const response = await fetch(url, {
-      headers: { 'Content-Type': 'application/json' },
+      headers: {
+        'Content-Type': 'application/json',
+        ...( idToken && {'Authorization': `Bearer ${idToken}`})
```
invincibleJai, line 72: is this needed, as with proxy header are propagated automatically and won't be seen on the client side. This plugin I have verified and similar should be case with Jfrog as well https://backstage.io/docs/tutorials/using-backstage-proxy-within-plugin/

invincibleJai:
@pbuehl-redhat Thanks for your contribution, would recommend linking the issue on the ticket.

cc @tumido

pbuehl-redhat:
@invincibleJai Thank you! I'm not exactly sure how to link the issue, but this is the issue that it needs to be linked with https://github.com/janus-idp/backstage-plugins/issues/486

invincibleJai:
> #486

@pbuehl-redhat you can add this to the PR description to start with, i.e Fixes: https://github.com/janus-idp/backstage-plugins/issues/486

and the bug you shared talks about `quay backstage plugins` but PR changes are made to more plugins like Jfrog, OCM, and Quay

cc @fmenesesg @christophe-f @tumido  ^^

tumido (commented):
CI is complaining, can you please run `yarn prettier:fix` on this PR?

Can you please also take a look at the https://github.com/janus-idp/backstage-plugins/pull/498#discussion_r1250390232 comment?

malagant:
I'm absolutely in need of this PR to be merged. Any hints, on when this will happen?

tumido:
Hi @malagant this PR has still 1 unresolved discussion that is awaiting a reply from from the PR author. Maybe you can help there?

Can you help me understand for which plugin do you need this for the most?

I'm also afraid that since all except OCM plugin affected by this PR here rely on Backstage proxy, this change actually doesn't change anything for them. As far as I know, identity api is not used by Backstage proxy.

plugins/quay/src/api/index.ts
tumido: @pbuehl-redhat it seems to me like you're trying to implement https://backstage.io/docs/auth/service-to-service-auth#usage-in-backend-plugins for plugins which **don't** have a backend plugin counterpart.

malagant:
Hi @tumido, thanks for the really fast response. I am using the quay plugin. Wanted to showcase it but found out that backend auth of backstage is not implemented. I have a workaround: Disable backend auth. But this only fits because we don't expose anything outside the company.

Let me check, what's blocking and jump into the discussion. Would be glad to support/help.

tumido:
Interesting.... @malagant can you please share with me your backend app configuration for the quay plugin? What confuses me is that Quay is using Backstage backend proxy...

I suppose you use something like:

```ts
  apiRouter.use('/proxy', authMiddleware, await proxy(proxyEnv));
```

correct?

plugins/quay/src/api/index.ts
tumido: @invincibleJai I think I'm finding the answers myself. You can run the proxy with auth middleware. In that case you need to be able to authenticate against it.

malagant:
> Interesting.... @malagant can you please share with me your backend app configuration for the quay plugin? What confuses me is that Quay is using Backstage backend proxy...
> 
> I suppose you use something like:
> 
> ```ts
>   apiRouter.use('/proxy', authMiddleware, await proxy(proxyEnv));
> ```
> 
> correct?
```
proxy:
  # Quay
  '/quay/api':
    target: 'https://...'
    headers:
      X-Requested-With: 'XMLHttpRequest'
      Authorization: 'Bearer ${MTR_TOKEN}'
``` 

And yes you're correct.

tumido:
I that case it all makes sense to me. 🙂 Let's get this merged. 👍

plugins/quay/src/api/index.ts
tumido: However, as I've learnt now, you can make Proxy to use backend service auth as well. In that case this applies to the  plugins which use proxy as well.

```ts
  apiRouter.use('/proxy', authMiddleware, await proxy(proxyEnv));
```

tumido (approved):
/lgtm
/approve

PR #510, opened 2023-07-05
Add admin role for RBAC

PatAKnight:
# Description
Adds an Admin role to the RBAC plugin. This PR allows for admins to be defined within the `app-config` as either individual users or groups. There are also administrator permissions that have been added for checking if there are appropriate permissions before performing sensitive actions. The permissions are used within the RBAC router to handle access to route that we plan to create for the plugin. Currently there is only a single route exposed that will return 'Authorized' or 'Unauthorized'. 

# Which issue(s) does this PR fix
- Fixes [issue 357](https://github.com/janus-idp/backstage-showcase/issues/357)

# How to test changes / Special notes to the reviewer
- There needs to be some form of authentication so that backstage assigns you to a user and gives you an appropriate token. Without it, you will come up as undefined and the checks will fail.
- There will also need to be a user imported into backstage so that backstage has a user to assign you to.
- Within the `app-config` you will need to add the following under the permission section
```
permission:
  enabled: true
  admin:
    users:
      - name: <USER_WITH_ADMIN_PERMISSIONS
    groups:
      - name: <GROUP_WITH_ADMIN_PERMISSIONS
```

### DIFF
--- a/plugins/rh-rbac-backend/src/permissions.ts
+++ b/plugins/rh-rbac-backend/src/permissions.ts
@@ -1,0 +1,68 @@
+import {
+  createPermission,
+  ResourcePermission,
+} from '@backstage/plugin-permission-common';
+
+export const RESOURCE_TYPE_PERMISSION_ENTITY = 'permission-entity';
+
+/**
+ * Convenience type for permission entity
+ */
+export type PermissionEntityPermission = ResourcePermission<
+  typeof RESOURCE_TYPE_PERMISSION_ENTITY
+>;
+
+/**
+ * This permission is used to authorize actions that involve reading
+ * permission policies.
+ */
+export const permissionEntityReadPermission = createPermission({
+  name: 'permission.entity.read',
+  attributes: {
+    action: 'read',
+  },
+  resourceType: RESOURCE_TYPE_PERMISSION_ENTITY,
+});
+
+/**
+ * This permission is used to authorize the creation of new permission policies.
+ */
+export const permissionEntityCreatePermission = createPermission({
+  name: 'permission.entity.create',
+  attributes: {
+    action: 'create',
+  },
+});
+
+/**
+ * This permission is used to authorize actions that involve removing permission
+ * policies.
+ */
+export const permissionEntityDeletePermission = createPermission({
+  name: 'permission.entity.delete',
+  attributes: {
+    action: 'delete',
+  },
+  resourceType: RESOURCE_TYPE_PERMISSION_ENTITY,
+});
+
+/**
+ * This permission is used to authorize refreshing permission policies
+ */
+export const permissionEntityRefreshPermission = createPermission({
+  name: 'permission.entity.refresh',
+  attributes: {
+    action: 'update',
+  },
+  resourceType: RESOURCE_TYPE_PERMISSION_ENTITY,
+});
+
+/**
+ * List of all permissions on permission polices.
+ */
+export const permissionEntityPermissions = [
+  permissionEntityReadPermission,
+  permissionEntityCreatePermission,
+  permissionEntityDeletePermission,
+  permissionEntityRefreshPermission,
+];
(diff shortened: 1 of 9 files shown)


plugins/rh-rbac-backend/src/service/policy-builder.ts
```diff
@@ -30,13 +44,51 @@ export class PolicyBuilder {
...
+        env.logger,
+        fileAdapter,
+        env.config,
+      ),
     };
-    return createRouter(options);
+
+    const router = await createRouter(options);
+
+    const permissionsIntergrationRouter = createPermissionIntegrationRouter({
```
AndrienkoAleksandr, line 63: Small typo permissionsIntergrationRouter => permissionsIntegrationRouter

plugins/rh-rbac-backend/src/permissions.ts
```diff
@@ -0,0 +1,68 @@
...
+});
+
+/**
+ * This permission is used to authorize the creation of new permission policies.
+ */
+export const permissionEntityCreatePermission = createPermission({
+  name: 'permission.entity.create',
+  attributes: {
+    action: 'create',
+  },
```
AndrienkoAleksandr, line 34: Sorry, why we don't use `resourceType: RESOURCE_TYPE_PERMISSION_ENTITY,` here?

plugins/rh-rbac-backend/src/service/permission-policy.ts
```diff
@@ -26,14 +31,41 @@ e = some(where (p.eft == allow)) && !some(where (p.eft == deny))
 m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
 `;
 
+const useAdmins = (admins: Config[], enf: Enforcer) => {
+  admins.flatMap(async localConfig => {
+    const name = localConfig.getString('name');
+    const adminPermission = [name, 'permission-entity', 'read', 'allow'];
```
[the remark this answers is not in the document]
PatAKnight: Right now, I just added the `permission-entity` `read` Since we only have the one route exposed. I figured whenever we begin work on adding the CRUD operations, I could go in and add all of the permissions at once. However, I can add them now if you would prefer it.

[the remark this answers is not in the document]
gorkem: Yes. When I think about it, we may not need this many permissions, only read and write. Also, the current config does not allow us to configure them separately, so it is OK to add both now.

plugins/rh-rbac-backend/src/permissions.ts
```diff
@@ -0,0 +1,69 @@
...
+  attributes: {
+    action: 'delete',
+  },
+  resourceType: RESOURCE_TYPE_POLICY_ENTITY,
+});
+
+/**
+ * This permission is used to authorize refreshing permission policies
+ */
+export const policyEntityRefreshPermission = createPermission({
```
gorkem, line 53: This is a confusing permission. If this is meant for re-reading the values from storage, I do not think we need an explicit permission for it (read should be enough). If this is meant for updating the values for existing keys on storage, in that case, the name refresh is not reflecting that purpose.

plugins/rh-rbac-backend/src/service/permission-policy.ts
```diff
@@ -26,14 +31,36 @@ e = some(where (p.eft == allow)) && !some(where (p.eft == deny))
 m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
 `;
 
+const useAdmins = (admins: Config[], enf: Enforcer) => {
```
gorkem, line 34: nicely done!

plugins/rh-rbac-backend/src/permissions.ts
PatAKnight: I renamed it to policyEntityUpdatePermission to more accurately reflect that it is an update permission

gorkem (approved):
We can merge this one in.

AndrienkoAleksandr (approved):

Issue #522, opened 2023-07-13
Ansible: Populate the Backstage catalog with job templates from AAP instance


invincibleJai:
**Description:** 

As a user, I would like to populate the Backstage catalog with job templates from the AAP instance

**Acceptance Criteria**

- [x] Create a backend plugin for AAP
- [x] User should be able to configure AAP instance with URL, token, etc
- [x] AAP plugin will populate the software catalog from the configured AAP instance

**Additional Info:**
Can take inspiration from 3scale-backend plugin and sample catalog can be seen below

```
apiVersion: backstage.io/v1alpha1
kind: Resource
metadata:
  name: sampleJobTemplate
  title: Ansible Automation Job Template
  description: Job Templates allow for a set of parameters to be defined to run an Ansible Job.
  links:
    - url: https://docs.ansible.com/automation-controller/latest/html/userguide/job_templates.html
      title: Documentation
      icon: web
spec:
  type: ansible-job-template
  lifecycle: production
  owner: janus-authors
  system: janus-idp
```

jeffcpullen:
Not sure if this implied, but we would want to make sure this includes workflow job templates as well.

invincibleJai:
> Not sure if this implied, but we would want to make sure this includes workflow job templates as well.

for this ticket job templates are picked from API `/api/v2/job_templates`. Sorry am not very thorough with AAP, are workflow job templates not supported with the same API? how do we get those? 

cc @cooktheryan

cooktheryan:
I did not realize that the job_templates and wf_templates were separated. They are in the same space in the UI

@invincibleJai /api/v2/workflow_job_templates/ is for workflow templates. Would you like me to file an RFE for those?

PR #527, opened 2023-07-18
feat(aap): aap backend plugin to support job templates in catalog provider

invincibleJai:
# Tracks:
 
https://github.com/janus-idp/backstage-plugins/issues/522

# Description:

populate the Backstage catalog with job templates from the AAP instance

# Screenshots/Gif:

**Light Theme:**

image

image

**Dark Theme:**

image

image

# Tests

image


plugins/aap-backend/config.d.ts
```diff
@@ -0,0 +1,18 @@
...
+  authorization: string;
+  system?: string;
+  owner?: string;
+  schedule?: TaskScheduleDefinitionConfig;
+}
+
+export interface Config {
+  catalog?: {
+    providers?: {
+      aapResourceEntity?: { [key: string]: AapConfig };
```
tumido, line 15: Can we rename this? If we ever want to extend this to something else than Resource, this name will become confusing. No other catalog provider is named like this.

Would a simple aap be enough? WDYT?

```suggestion
      aap?: { [key: string]: AapConfig };
```

plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    return {
+      kind: 'Resource',
+      apiVersion: 'backstage.io/v1alpha1',
+      metadata: {
+        annotations: {
+          [ANNOTATION_LOCATION]: this.getProviderName(),
+          [ANNOTATION_ORIGIN_LOCATION]: this.getProviderName(),
+        },
+        name: `${jobTemplateName}`,
```
tumido, line 150: I'm no AAP expert, but can we expect this to be unique? Can't you have different templates (maybe in different namespaces) with the same name? If this doesn't apply here, we should be fine. It just struck me as a potential issue.

plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  private constructor(
+    config: AapConfig,
+    logger: Logger,
+    taskRunner: TaskRunner,
+  ) {
+    this.env = config.id;
+    this.baseUrl = config.baseUrl;
+    this.authorization = config.authorization;
+    this.owner = config.owner || 'AAP';
```
tumido, line 72: Owner is a backstage entity reference. `AAP` is not a valid value.

https://backstage.io/docs/features/software-catalog/descriptor-format/#specowner-required-2

plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
```diff
@@ -0,0 +1,170 @@
...
+          {
+            url: `${location}`,
+            title: 'Job Template',
+          },
+        ],
+      },
+      spec: {
+        type: `${template.type}`,
+        owner: `${this.owner}`,
+        ...(this.system && { system: `${this.system}` }),
```
tumido, line 166:
```suggestion
        owner: this.owner,
        system: this.system,
```

This should be fine. Empty values should be discarded/pruned.

tumido (changes requested):
I have few suggestions inline. PTAL 🙂

plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
invincibleJai: Tried that but it's not working, so good to check and update

```
backend:start: 2023-07-19T06:24:08.907Z catalog warn Processor BuiltinKindsEntityProcessor threw an error while validating the entity resource:default/demo_job_template; caused by TypeError: /spec/system must NOT have fewer than 1 characters - limit: 1 type=plugin entity=resource:default/demo_job_template location=AapResourceEntityProvider:dev
```

plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
invincibleJai: yeah i need to transform it , can't use the name as is

plugins/aap-backend/package.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "dependencies": {
+    "@backstage/backend-common": "^0.18.5",
+    "@backstage/backend-tasks": "^0.5.2",
+    "@backstage/catalog-model": "^1.3.0",
+    "@backstage/config": "^1.0.7",
+    "@backstage/plugin-catalog-node": "^1.3.7",
+    "winston": "^3.2.1",
+    "node-fetch": "^2.6.7",
+    "yn": "^4.0.0"
```
debsmita1, line 32: Is this used anywhere ?

plugins/aap-backend/src/providers/AapResourceEntityProvider.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      AapResourceEntityProvider.fromConfig(config, {
+        logger: getVoidLogger(),
+      }),
+    ).toThrow(
+      'No schedule provided neither via code nor config for AapResourceEntityProvider:dev.',
+    );
+  });
+
+  it('should return a single provider if one is configured', () => {
+    // const config = new ConfigReader(BASIC_VALID_CONFIG_2);
```
debsmita1, line 84: remove this line

plugins/aap-backend/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "backstage": {
+    "role": "backend-plugin"
+  },
+  "scripts": {
+    "start": "backstage-cli package start",
```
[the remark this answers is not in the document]
invincibleJai: this is backend plugin and can see this in script for upstream and Janus backend plugin both so kept it

plugins/aap-backend/package.json
invincibleJai: updated redundant dependencies

debsmita1:
tested this locally
LGTM

plugins/aap-backend/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@backstage/catalog-model": "^1.3.0",
+    "@backstage/config": "^1.0.7",
+    "@backstage/plugin-catalog-node": "^1.3.7",
+    "winston": "^3.2.1"
+  },
+  "devDependencies": {
+    "@backstage/cli": "0.22.7",
+    "@types/supertest": "2.0.12",
+    "supertest": "6.3.3",
+    "msw": "1.2.2"
```
tumido, line 36: Since #530 is merged now, can you please bump these as well?

https://github.com/janus-idp/backstage-plugins/blob/main/packages/backend/package.json

tumido (changes requested):
I think this needs a package/dependency bump since #530 is merged. Can you please rebase and update those deps?

tumido (approved):
/lgtm
/approve

Issue #480, opened 2023-06-22
App Topology: Code decorator support (Git URL)


divyanshiGupta:
## Description
As a developer, 
- I want to to be able to quick access the GH repo of my workload from the topology view.

## Acceptance Criteria
Add the decorator in the bottom right corner of nodes in the topology view 
- Clicking on the code decorator should go to the git repo in context (in a new tab)

Issue #250, opened 2023-04-12
App Topology: Code decorator support 


serenamarie125:
## Description
As a developer, 
- I want to to be able to quick access the GH repo of my workload from the topology view.
- If  OpenShift DevSpaces is installed, I want quick access to launch OpenShift Dev Spaces with my code in context from the topology view

## Acceptance Criteria
Add the decorator in the bottom right corner of nodes in the topology view 
- If OpenShift DevSpaces is installed, clicking on the decorator should launch OpenShift Dev Spaces with the code in context
- If OpenShift DevSpaces is not installed, clicking on the code decorator should go to the git repo in context (in a new tab)

PR #535, opened 2023-07-19
feat(topology): add edit code node decorator

divyanshiGupta:
**Stories:** https://github.com/janus-idp/backstage-plugins/issues/480, https://github.com/janus-idp/backstage-plugins/issues/250

This PR adds code decorator on topology node for git and devspace url

Demo:

With Git repo URL:

https://github.com/janus-idp/backstage-plugins/assets/20724543/ae66cd32-3711-4acf-a943-22b9217de6a8

With Che URL:

https://github.com/janus-idp/backstage-plugins/assets/20724543/1285994e-c64a-46bc-ba37-d9b34aff7e28

### DIFF
--- a/plugins/topology/README.md
+++ b/plugins/topology/README.md
@@ -81,13 +81,68 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
 
   ```
 
+- The following configuration must be added in`customResources` component in the [`app-config.yaml`](https://backstage.io/docs/features/kubernetes/configuration#configuring-kubernetes-clusters) file to view the edit code decorator:
+
+  ```yaml
+   kubernetes:
+     ...
+     customResources:
+        - group: 'org.eclipse.che'
+          apiVersion: 'v2'
+          plural: 'checlusters'
+  ```
+
+  Also, ensure that the `CheCluster` is granted a [`ClusterRole`](https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control) as shown in the following example code:
+
+  ```yaml
+    ...
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: ClusterRole
+    metadata:
+      name: backstage-read-only
+    rules:
+      ...
+      - apiGroups:
+          - org.eclipse.che
+        resources:
+          - checlusters
+        verbs:
+          - get
+          - list
+
+  ```
+
+- The following annotations are added to workload resources, such as Deployments to navigate to the Git repository of the associated application using the edit code decorator:
+
+  ```yaml title="deployment.yaml"
+    ...
+    labels:
+      app.openshift.io/vcs-uri: <GIT_REPO_URL>
+  ```
+
+  You can also add the following annotation to navigate to a specific branch:
+
+  ```yaml title="deployment.yaml"
+      ...
+      labels:
+        app.openshift.io/vcs-ref: <GIT_REPO_BRANCH>
+  ```
+
+  > Note: If Red Hat OpenShift Dev Spaces is [installed and configured](https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.7/html/administration_guide/installing-devspaces) and Git URL annotations are also present on the workload YAML file, then clicking on the edit code decorator redirects you to the Red Hat OpenShift Dev Spaces instance.
+
+  > Note: When you deploy the application using the OCP Git import flows, then you do not need to add the labels as import flows add the labels to the workload YAML file. Otherwise, you need to add the labels manually to the workload YAML file.
+
+  The labels are not similar to `backstage.io/edit-url` annotations as it points to the catalog entity metadata source file and is applied to Backstage catalog entity metadata YAML file, but not Kubernetes resources.
+
+  > Tip: You can also add the `app.openshift.io/edit-url` annotation with the edit URL that you want to access using the decorator.
+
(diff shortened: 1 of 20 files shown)


debsmita1:
can see this warning when  topology view loads

https://github.com/janus-idp/backstage-plugins/assets/22490998/d2f0215b-7ef1-4a3e-b2d0-2965d3193ce2

divyanshiGupta:
@hmanwani-rh I have updated the README. PTAL.

plugins/topology/README.md
```diff
@@ -81,6 +81,57 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
...
+      name: backstage-read-only
+    rules:
+      ...
+      - apiGroups:
+          - org.eclipse.che
+        resources:
+          - checlusters
+        verbs:
+          - get
+          - list
```
tumido, line 111: Can you please add this to the Cluster Role as well?

https://github.com/janus-idp/backstage-plugins/blob/main/plugins/topology/manifests/clusterrole.yaml

plugins/topology/README.md
```diff
@@ -81,6 +81,57 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
...
+
+  ```
+
+- The following annotation(s) are added to workload resources such as Deployments to navigate to the git repo via the edit code decorator.
+
+  ```yaml title="deployment.yaml"
+    ...
+    labels:
+      app.openshift.io/vcs-uri: <GIT_REPO_URL>
```
tumido, line 120: Why do you need an additional url on the Kuberentes resource, if Backstage already provides a standard one for the catalog entry?

https://backstage.io/docs/features/software-catalog/well-known-annotations/#backstageioview-url-backstageioedit-url

I think this is just confusing the user.

plugins/topology/src/utils/resource-utils.ts
```diff
@@ -288,3 +288,11 @@ export const getUrlForResource = (
...
   );
 };
+
+export const getCheCluster = (resources: K8sResponseData) => {
+  if (!resources.checlusters?.data?.length) return undefined;
+
+  return resources.checlusters.data.find(
+    cc => cc.metadata?.namespace === 'openshift-devspaces',
+  );
+};
```
tumido, line 298:
```
[].find(...) === undefined
```

Find on empty data already returns `undefined`. This whole function can be simplified to:

```suggestion
export const getCheCluster = (resources: K8sResponseData) => 
  resources.checlusters?.data?.find(
    cc => cc.metadata?.namespace === 'openshift-devspaces',
  );
```

tumido (changes requested):

plugins/topology/README.md
tumido: Yes, I like that idea. I think we should put something along these lines to the readme:

1. The annotations are already populated by the OCP import - so user doesn't have to do anything in that case.
2. In cases when OCP deploy was not via git import, they may need to do it manually and they should not confuse it with `backstage.io/edit-url` annotation since that points to something else and is applied to a completely different entity (backstage catalog entry, not kubernetes resource).

divyanshiGupta: @hmanwani-rh Please review the README and lmk if you have any feedback. Thanks!

hmanwani-rh (changes requested):
@divyanshiGupta I've added some suggestions below. Please take a look and let me know if you've any questions.

plugins/topology/README.md
```diff
@@ -81,6 +81,59 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
...
+
+  ```yaml
+   kubernetes:
+     ...
+     customResources:
+        - group: 'org.eclipse.che'
+          apiVersion: 'v2'
+          plural: 'checlusters'
+  ```
+
```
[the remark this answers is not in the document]
divyanshiGupta: @hmanwani-rh So the edit code decorator will redirect the user to Red Hat OpenShift Dev Spaces instance only when Red Hat OpenShift Dev Spaces operator is installed and configured on OCP cluster. If its not installed then it will redirect the user to the application git repo. I have added this point as a note.

hmanwani-rh (approved):

plugins/topology/README.md
```diff
@@ -81,6 +81,61 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
...
+  You can also add the following annotation to navigate to a specific branch:
+
+  ```yaml title="deployment.yaml"
+      ...
+      labels:
+        app.openshift.io/vcs-ref: <GIT_REPO_BRANCH>
+  ```
+
+  > Note: If Red Hat OpenShift Dev Spaces is installed and configured and Git URL annotations are also present on the workload YAML file, then clicking on the edit code decorator redirects you to the Red Hat OpenShift Dev Spaces instance.
```
debsmita1, line 131: Should we also add the link to the instructions for installing and configuring Openshift Dev Spaces ?

plugins/topology/src/components/Graph/decorators/EditDecorator.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const editUrl = editURL || getEditURL(vcsURI, vcsRef, cheURL);
+  const repoIcon = (
+    <RouteDecoratorIcon
+      routeURL={editUrl}
+      radius={radius}
+      cheEnabled={cheEnabled}
+    />
+  );
+
+  if (!repoIcon) {
```
debsmita1, line 39: Will this ever be true ?

divyanshiGupta: Yes in case the url is invalid

debsmita1:
LGTM !

tumido (approved):
/lgtm
/approve

Issue #511, opened 2023-07-06
Keycloak - more flexibility at import time


paoloantinori:
**Is your feature request related to a problem? Please describe.**
I've configured successfully the Backstage Plugin to communicate with RH SSO 7.6.2.GA .
In the KC realm we are using we have enabled Github as an IdP provider.
Unfortunately, the default mapping uses Github emails as userid.
And an e-mail is not an acceptable identifier for Backstage:

```
2023-07-05T16:48:11.901Z catalog warn Policy check failed for user:default/paolo.antinori@gmail.com; caused by Error: "metadata.name" is not valid; expected a string that is sequences of [a-zA-Z0-9] separated by any of [-_.], at most 63 characters in total but found "paolo.antinori@gmail.com". To learn more about catalog file format, visit: https://github.com/backstage/backstage/blob/master/docs/architecture-decisions/adr002-default-catalog-file-format.md type=plugin entity=user:default/paolo.antinori@gmail.com
```

**Describe the solution you'd like**
Assuming that using an external IdP in Keycloak is a pretty common use-case, I'm looking for ways to bypass this issue.
Ideally the Keycloak Backstage plugin could be extended with some pre-defined strategy to clean data to accept only Backstage compatible values. Even better if I could provide directly a javascript lambda, where I can specify my own custom cleaning logic.

**Describe alternatives you've considered**
As a mitigation, I think that KC should be able to be instructed to use a different convention, but I think it might still be valuable to have this feature present also on the Backstage plugin side.
I'm right now asking the KC engineers what would be the correct way to mitigate this problem on the Keycloak side, since all my attempts failed so far.

paoloantinori:
This is the potential bug on Keycloak preventing to mitigate this issue on its side: https://github.com/keycloak/keycloak/issues/11195

tumido:
Hi @paoloantinori yes, I think we want to enable this behavior.

We want users to be able to use `userTransformer` and `groupTransformer` functions that allow them to mutate the objects before their ingestion into Catalog.

This should satisfy similar needs as described for example here in LDAP plugin docs:

https://backstage.io/docs/integrations/ldap/org/#customize-the-provider

Do you think such a feature can help you with this problem? Or are you looking for a different solution?

There's a PR from the community introducing something like this, however, it still requires a lot of changes to reduce hardcoded behavior and offer the necessary customizations we strive for: https://github.com/janus-idp/backstage-plugins/pull/432

PR #542, opened 2023-07-19
feat(keycloak): add userTransformer and groupTransformer

tumido:
Replaces: #432 

Resolves: #511

Proper implementation of `userTransformer` and `groupTransformer`.

Loosely inspired by LDAP plugin's transformers https://backstage.io/docs/integrations/ldap/org/#customize-the-provider

This allows users to specify custom functions when initializing the entity provider. These functions will run for each ingested entity providing context about the:
- The resulting entity which is the output of the default parser
- The original data from Keycloak API
- Realm name information
- For users - additional group data

Proposed workflow:
1. We (pre)process the `Group`s first
   1. We parse API response to the entity via the default parser
   2. We apply `groupTransformer`
2. We process the `User`s
   1. We parse API response to the entity via the default parser
   2. We apply `userTransformer`
4. We (post)process the `Group`s again to update memberships and group hierarchy to address any name changes in groups or users entities.

cc @paoloantinori @tanoggy @jani888 let me know what you think 🙂

### DIFF
--- a/plugins/keycloak-backend/README.md
+++ b/plugins/keycloak-backend/README.md
@@ -130,6 +130,65 @@ yarn workspace backend add @janus-idp/backstage-plugin-keycloak-backend
            # highlight-add-end
    ```
 
+4. Optional: provide a transformer function for user/group to mutate the entity before their ingestion into catalog. Extend `packages/backend/src/plugins/catalog.ts` with custom `userTransformer` and `groupTransformer` functions:
+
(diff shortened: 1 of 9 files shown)


jani888 (approved):
Nice job, this pull request is looking solid! Thanks for the speedy work, really appreciate it. 🔥 💯

schwesig (approved):
/lgtm
/approve
checked and feedback to and from Tom

PR #543, opened 2023-07-19
chore(keycloak): provide a standalone dev environment

tumido:
Depends on #530 (I wasn't able to properly resolve types for Logger without an upgrade)

Part of: https://github.com/janus-idp/backstage-plugins/issues/330

For the changes meant to be introduced in this PR please see a8c95e004850752066e3276b3bc2d628cb1f4b04 commit.

This PR introduces a standalone dev environment for Keycloak plugin. It starts in parallel:
- Keycloak plugin as a standalone backend server with catalog API available
- Keycloak instance as a local container

It automatically imports the realm consumed by Keycloak plugin

### DIFF
--- a/plugins/keycloak-backend/package.json
+++ b/plugins/keycloak-backend/package.json
@@ -14,7 +14,9 @@
     "role": "backend-plugin"
   },
   "scripts": {
-    "start": "backstage-cli package start",
+    "start": "open http://localhost:8080/admin/master/console/#/janus-realm && open http://localhost:7007/catalog/entities && turbo run start:plugin start:keycloak",
+    "start:plugin": "backstage-cli package start",
+    "start:keycloak": "podman run -p 8080:8080 -e 'KEYCLOAK_ADMIN=admin' -e 'KEYCLOAK_ADMIN_PASSWORD=admin' -v ./__fixtures__/keycloak-realm.json:/opt/keycloak/data/import/keycloak-realm.json quay.io/keycloak/keycloak:22.0.1 start-dev --import-realm",
     "build": "backstage-cli package build",
     "lint": "backstage-cli package lint",
     "test": "backstage-cli package test --passWithNoTests --coverage",
@@ -34,11 +36,17 @@
     "winston": "^3.2.1"
   },
   "devDependencies": {
+    "@backstage/backend-app-api": "^0.4.5",
     "@backstage/cli": "0.22.9",
+    "@backstage/plugin-auth-node": "^0.2.16",
+    "@backstage/plugin-catalog-backend": "^1.11.0",
+    "@backstage/plugin-permission-common": "^0.7.7",
+    "@backstage/plugin-permission-node": "^0.7.10",
     "@types/lodash": "4.14.195",
     "@types/supertest": "2.0.12",
     "@types/uuid": "9.0.1",
     "deepmerge": "4.3.1",
+    "express": "^4.18.2",
     "msw": "1.2.1",
     "supertest": "6.3.3"
   },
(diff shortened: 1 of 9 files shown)


tumido:
@schultzp2020 @RobotSail I've finally figured out the standalone backend dev environments that include catalog providers! Check this out! (yes, I'm proud of myself, haha) 🎉

plugins/keycloak-backend/package.json
```diff
@@ -14,7 +14,9 @@
     "role": "backend-plugin"
   },
   "scripts": {
-    "start": "backstage-cli package start",
+    "start": "concurrently -n 'keycloak,backstage' 'yarn start:keycloak' 'yarn start:dev' 'open http://localhost:8080/admin/master/console/#/janus-realm' 'open http://localhost:7007/catalog/entities'",
+    "start:dev": "backstage-cli package start",
+    "start:keycloak": "podman run -p 8080:8080 -e 'KEYCLOAK_ADMIN=admin' -e 'KEYCLOAK_ADMIN_PASSWORD=admin' -v ./__fixtures__/keycloak-realm.json:/opt/keycloak/data/import/keycloak-realm.json quay.io/keycloak/keycloak:22.0.1 start-dev --import-realm",
```
schultzp2020, line 19: Is it possible to use turbo instead of concurrently here? I think it should be possible.

tumido: @schultzp2020 can you help me understand how it should work? I'm clueless atm. This is a single workspace specific task, that needs to spawn 2 commands instead of one. How is that achievable via turbo and what would be the benefit in this case? 🤷

schultzp2020, line 19:
```suggestion
    "start": "turbo run start:backstage start:fixtures",
    "start:backstage": "backstage-cli package start",
    "start:fixtures": "podman run -p 8080:8080 -e 'KEYCLOAK_ADMIN=admin' -e 'KEYCLOAK_ADMIN_PASSWORD=admin' -v ./__fixtures__/keycloak-realm.json:/opt/keycloak/data/import/keycloak-realm.json quay.io/keycloak/keycloak:22.0.1 start-dev --import-realm",
```

You'll also need to add the following to `turbo.json`.

```json
{
  ...
  "pipeline": {
    "start": {
      "cache": false,
      "persistent": true,
      "dependsOn": ["^build"]
    },
    "start:backstage": {
      "cache": false,
      "persistent": true
    },
    "start:fixtures": {
      "cache": false,
      "persistent": true
    },
    ...
  }
}
```

I changed the scripts to `start:backstage` and `start:fixtures` to be more generic so other plugins can use this template. The main benefits to this would be an easier template to copy and less dependencies.

tumido: Thank you! 👍 Nice job. I'll apply this change.

tumido: Nice... I've managed to find an even less invasive solution. 🙂 PTAL

tumido: I'll do it in a separate PR 🙂

schultzp2020 (approved):
LGTM

Zaperex:
When I tried running the standalone dev environment using `yarn --cwd ./plugins/keycloak-backend start` from the project root directory, I get the following error:
```bash
yarn run v1.22.19
$ open http://localhost:8080/admin/master/console/#/janus-realm && open http://localhost:7007/catalog/entities && turbo run start:plugin start:keycloak
/bin/sh: line 1: open: command not found
error Command failed with exit code 127.
```
FYI, I'm using Fedora 38.

gashcrumb:
@Zaperex is there an issue for that?  I agree this start command won't work for linux.  I'd go so far as to suggest removing the open commands and instead improve the documentation around them in the CONTRIBUTING.md, as the `&&` in the command will also be an issue for some folks as well.

Zaperex:
@gashcrumb there isn't an issue yet, but I can make one.

Issue #521, opened 2023-07-13
Stabilise release toolkit


tumido:
`semantic-release` is not the best suitable tool for monorepos. With a growing number of packages released from this repo, we're hitting issues with NPM registry rate limiting.

Related:
- https://github.com/semantic-release/semantic-release/issues/2313 
- https://github.com/janus-idp/backstage-plugins/issues?q=+is%3Aissue+label%3Asemantic-release

Ideally, we should migrate to a tool that allows similar/same level of confidence and release cadence without additional manual work or overhead.

Based on the list of compatible release tools here:

https://turbo.build/repo/docs/handbook/publishing-packages/versioning-and-publishing

I chose [`intuit/auto`](https://intuit.github.io/auto/index) with [`conventional-commits` plugin](https://github.com/intuit/auto/tree/main/plugins/conventional-commits). Huge benefit of this solution is that we don't have to change anything for the developers, no need to write manual change sets, etc.

tumido:
After a bit of exploration and trials in a test repo I don't think `auto` provides us suitable/acceptable results. Key limitations:

- True monorepo support requires lerna, I wasn't able to make it work without it
- Without lerna, it has troubles handling package dependencies between released packages
- Requires workspace root `package.json` version bumps

I've tried looking into the semantic-release problems once more, and I may have a workaround to stabilize the situation for the time being.

I think the rate limit issues come from:
https://github.com/dhoulb/multi-semantic-release#npm-invalid-npm-token

So we should be fine with applying the suggested workaround.

PR #557, opened 2023-07-24
chore(ci): workaround semantic-release/npm rate limits

tumido:
Resolves: https://github.com/janus-idp/backstage-plugins/issues/521

Instead of release toolkit migration, which didn't yield sufficient results, I've tried working around the semantic-release issues instead.

This PR applies https://github.com/semrel-extra/npm workaround to the mysterious NPM rate limits.. Let's see if it helps.


tumido:
Self merging this to resolve releasing issues.

PR #562, opened 2023-07-25
chore(kiali): fix dependencies and test arguments

tumido:
SSIA, should make Kiali use the same Backstage deps as rest of the repo and should make the tests pass.


schwesig (approved):
/lgtm
/approve
tested it together with Tom

PR #573, opened 2023-07-26
feat(kiali): move from node-fetch to axios

aljesusg:
Move node-fetch to axios.

Related with https://github.com/janus-idp/backstage-plugins/pull/460


tumido:
Oh.. sorry for the trouble @aljesusg . If I knew this is so much trouble and pain, I wouldn't ask and would happily leave `node-fetch@v2` as is. Next time, if you see me requesting something that is not worth investing your time in, just say no. With a reasonable explanation like you've provided in the https://github.com/janus-idp/backstage-plugins/pull/460#issuecomment-1651159333 comment, we can definitely find some middle ground. 🙂 

Solely the argument about ES modules would be enough in this case, since Backstage doesn't support that. 🤷‍♂️ 😄

Issue #547, opened 2023-07-20
App Topology: Show workload runtime icon in topology node


divyanshiGupta:
## Description
As a developer, 
- I want to to be able to see the associated runtime icon in the topology workload node

## Acceptance Criteria

- Show runtime icons for each topology workload node (similar to OCP)
- Add a `Display options` dropdown with a checkbox option to toggle between pod no and icon (similar to OCP)

<img width="170" alt="Screenshot 2023-07-26 at 6 05 47 PM" src="https://github.com/janus-idp/backstage-plugins/assets/20724543/66dbf16d-83fa-4909-a27a-dbf0a175114c">

PR #615, opened 2023-07-31
feat(topology): add runtime icon to topology node

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/547

**GIF:**

https://github.com/janus-idp/backstage-plugins/assets/22490998/5636d664-bc00-4ebc-8383-711a7c9fb7a4


plugins/topology/README.md
```diff
@@ -128,6 +128,14 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
...
   ```
 
+- The following label is added to workload resources, such as Deployments to display runtime icon in the topology nodes:
+
+  ```yaml title="deployment.yaml"
+    ...
+    labels:
+      app.openshift.io/runtime: <RUNTIME_NAME>
+  ```
+
```
divyanshiGupta, line 138: @debsmita1 I think we should also mention if this label is not present then this label `app.kubernetes.io/name` is used to get the icon so that the user can use either of these two.

divyanshiGupta (changes requested):

divyanshiGupta:
cc: @hmanwani-rh

divyanshiGupta:
@debsmita1 node doesnt show pod status when runtime icon is displayed

divyanshiGupta:
@debsmita1 Clicking on `pod count` checkbox doesnt change it pod count immediately. It takes more than one click and also takes a lot of time. Also look like their is some other issues with rendering as well as sometimes empty state is being shown.


https://github.com/janus-idp/backstage-plugins/assets/20724543/4d8d1234-5271-4524-9817-d8b376e32950

debsmita1:
> @debsmita1 Clicking on `pod count` checkbox doesnt change it pod count immediately. It takes more than one click and also takes a lot of time. Also look like their is some other issues with rendering as well as sometimes empty state is being shown.
> 
>  Screen.Recording.2023-08-09.at.4.07.51.PM.mov

Fixed it 

https://github.com/janus-idp/backstage-plugins/assets/22490998/5a0d975a-a9af-4d94-a1ef-9eeab713495f

PTAL

plugins/topology/README.md
```diff
@@ -128,6 +128,22 @@ The Topology plugin enables you to visualize the workloads such as Deployment, J
...
+  As another option, you can include the label below to display the runtime icon:
```
hmanwani-rh, line 139:
```suggestion
  As another option, you can include the following label to display the runtime icon:
```

hmanwani-rh (approved):
@debsmita1 Added a minor suggestion below. Rest looks good to me. 
thank you

divyanshiGupta:
@debsmita1  can we have a default-icon that will be shown when the runtime-icon is not available. For ex in minukube cluster this is how it is looking currently as it resources 
Screenshot 2023-08-10 at 5 34 42 PM
doesn’t have runtime-icon label.

debsmita1:
> @debsmita1 can we have a default-icon that will be shown when the runtime-icon is not available. For ex in minukube cluster this is how it is looking currently as it resources doesn’t have runtime-icon label.
> 
> Screenshot 2023-08-10 at 5 34 42 PM

It is left empty if there's no runtime icon available. 
Screenshot 2023-08-10 at 7 09 03 PM

divyanshiGupta (approved):
LGTM

plugins/topology/src/components/Graph/WorkloadNode.tsx
```diff
@@ -172,6 +176,13 @@ const InnerWorkloadNode = observer(
       ? getNodeDecorators(element, decorators, cx, cy, radius, decoratorRadius)
       : null;
 
+    const iconImageUrl =
+      workloadData.builderImage ??
+      getImageForIconClass(workloadData.builderImage);
```
invincibleJai, line 181: this looks confusing to me 🤔

invincibleJai (approved):
LGTM

Issue #581, opened 2023-07-27
Refactor Tekton plugin so it uses `plugin/shared-react` instead of custom code


tumido:
Part of epic: https://github.com/janus-idp/backstage-plugins/issues/580

This mainly affects files for:
- `PipelineRun`
- `pipeline-utils`
- `taskRun`
- `tekton-utils`
- `computedStatus`
- `CamelCaseWrap`

Reference:
https://sonarcloud.io/component_measures?metric=duplicated_lines&view=list&id=janus-idp_backstage-plugins

https://sonarcloud.io/component_measures?metric=duplicated_files&view=list&id=janus-idp_backstage-plugins

PR #623, opened 2023-08-07
fix(tekton): use shared-react package

debsmita1:
Resolves:
https://github.com/janus-idp/backstage-plugins/issues/581


invincibleJai (approved):
LGTM

PR #607, opened 2023-07-28
fix(kiali): fix code smells

aljesusg:
Close https://github.com/janus-idp/backstage-plugins/issues/585

PR #630, opened 2023-08-10
feat(kiali): show kiali information in header

aljesusg:
Require https://github.com/janus-idp/backstage-plugins/pull/607

header
panel


tumido (approved):
/lgtm
/approve

PR #631, opened 2023-08-10
fix(shared-react): fix code smells

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/595

**Solution description:**
- fixed code smells
- added common utils
- added more tests


plugins/shared-react/src/components/common/CamelCaseWrap.tsx
```diff
@@ -21,7 +21,7 @@ export const CamelCaseWrap = ({ value, dataTest }: CamelCaseWrapProps) => {
   const rendered = (
     <span data-test={dataTest}>
       {words?.map((word, i) => (
-        <React.Fragment key={i}>
+        <React.Fragment key={word}>
```
invincibleJai, line 24: will the word be unique here? as the key should be unique right

PR #632, opened 2023-08-10
fix(tekton): fix code smells

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/584


invincibleJai:
@debsmita1 sonarCloud is complaining about code duplication

debsmita1:
> @debsmita1 sonarCloud is complaining about code duplication

The `debounce.ts` file now exists in `topology` and `shared-react`. I will raise a separate PR to remove the debounce file from the topology plugin

invincibleJai (approved):
LGTM

PR #618, opened 2023-08-02
feat(nexus): add nexus repository manager plugin

schultzp2020:
image

fixes: https://github.com/janus-idp/backstage-plugins/issues/567

PR #633, opened 2023-08-10
feat(shared-react): add date and unit conversion utils

schultzp2020:
Adds the date and unit conversion utils required for https://github.com/janus-idp/backstage-plugins/pull/618.


plugins/shared-react/src/utils/pipeline/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './pipeline';
+export * from './task-run';
```
schultzp2020, line 2: I removed the `-utils` suffix as it is redundant because its under the `utils` folder.

plugins/shared-react/package.json
```diff
@@ -23,7 +23,9 @@
   },
   "dependencies": {
     "@kubernetes/client-node": "^0.18.1",
-    "classnames": "2.3.2"
+    "classnames": "2.3.2",
+    "date-fns": "^2.30.0",
+    "mathjs": "^11.9.1"
```
invincibleJai, line 28: shall we have `^` here or use fixed version?

plugins/shared-react/src/utils/date/format-date.ts
```diff
@@ -0,0 +1,9 @@
+import { format } from 'date-fns';
+
+export function formatDate(date: string | undefined) {
```
invincibleJai, line 3: nit: good to be consistent and use arrow functions and have unit test for it

do we need to have date type as `string | undefined` or have it as just `string`? as the caller should pass a value and for runtime, if it's not there then we should do an early return as you are doing?

plugins/shared-react/src/utils/pipeline/index.ts
invincibleJai: now probably we can do this change but once it's published then it may break plugins(topology / tetkon) that are consuming it, once they update `shared-react` package

cc @debsmita1 / @divyanshiGupta ^^

invincibleJai: FYI @debsmita1 / @divyanshiGupta

plugins/shared-react/src/utils/date/format-date.ts
schultzp2020: > nit: good to be consistent and use arrow functions and have unit test for it

https://github.com/janus-idp/backstage-plugins/pull/618#discussion_r1290698458
Ill add a unit test.

> do we need to have date type as string | undefined or have it as just string? as the caller should pass a value and for runtime, if it's not there then we should do an early return as you are doing?

I did it this way for simplicity because one of the values in the nexus plugin is `string | undefined`

plugins/shared-react/package.json
schultzp2020: Only the dev dependencies need to be fixed. e.g. `3scale-backend` [`package.json`](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/3scale-backend/package.json)

invincibleJai (approved):
LGTM

PR #637, opened 2023-08-14
fix(shared-react): export hooks

debsmita1:
part of PR https://github.com/janus-idp/backstage-plugins/pull/631/files


debsmita1:
The code duplication in the hooks folder will be handled in PR https://github.com/janus-idp/backstage-plugins/pull/632

invincibleJai (approved):
LGTM

PR #641, opened 2023-08-14
chore(segment): remove deprecated plugin to reduce code smells

### LINKED ISSUE #590: Reduce code smells in Segment plugin
Part of: #583

Investigate code smells in the Tekton plugin and fix them via a PR or explained why they should be dismissed as irrelevant.

See this plugin folder and respective code smells column here:
https://sonarcloud.io/code?id=janus-idp_backstage-plugins&selected=janus-idp_backstage-plugins%3Aplugins

Zaperex:
### What issue does this PR resolve?
Fixes #590


tumido (changes requested):
I don't think we should remove the plugin without a replacement. The Segment plugin still relies on Backstage core APIs and should remain registered as a plugin.

Zaperex:
Ok I guess we can just ignore this code smell then.

Issue #617, opened 2023-08-02
Chore(JFrog Artifactory): remove `<component-name>.defaultProps`


schultzp2020:
## Work

Replace `<component-name>.defaultProps` with [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) because `defaultProps` has been [deprecated](https://github.com/jsx-eslint/eslint-plugin-react/issues/2396).

anilabhabaral:
Please assign @schultzp2020

PR #649, opened 2023-08-17
chore(JFrog Artifactory): remove <component-name>.defaultProps

anilabhabaral:
Resolves: [#617](https://github.com/janus-idp/backstage-plugins/issues/617)

### DIFF
--- a/plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
+++ b/plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
@@ -22,15 +22,15 @@ const useLocalStyles = makeStyles({
   },
 });
 
-export function JfrogArtifactoryRepository(props: RepositoryProps) {
+export function JfrogArtifactoryRepository({ image }: RepositoryProps) {
   const jfrogArtifactoryClient = useApi(jfrogArtifactoryApiRef);
   const classes = useStyles();
   const localClasses = useLocalStyles();
   const [edges, setEdges] = useState<Edge[]>([]);
-  const title = `Jfrog Artifactory repository: ${props.image}`;
+  const titleprop = `Jfrog Artifactory repository: ${image}`;
 
   const { loading } = useAsync(async () => {
-    const tagsResponse = await jfrogArtifactoryClient.getTags(props.image);
+    const tagsResponse = await jfrogArtifactoryClient.getTags(image);
 
     setEdges(tagsResponse.data.versions.edges);
 
@@ -65,7 +65,7 @@ export function JfrogArtifactoryRepository(props: RepositoryProps) {
   return (
     <div style={{ border: '1px solid #ddd' }}>
       <Table
-        title={title}
+        title={titleprop}
         options={{ paging: true, padding: 'dense' }}
         data={data}
         columns={columns}
@@ -80,11 +80,6 @@ export function JfrogArtifactoryRepository(props: RepositoryProps) {
   );
 }
 
-JfrogArtifactoryRepository.defaultProps = {
-  title: 'Docker Images',
-};
 interface RepositoryProps {
-  widget: boolean;
   image: string;
-  title: string;
 }
--- a/plugins/jfrog-artifactory/src/components/JfrogArtifactoryDashboardPage/JfrogArtifactoryDashboardPage.tsx
+++ b/plugins/jfrog-artifactory/src/components/JfrogArtifactoryDashboardPage/JfrogArtifactoryDashboardPage.tsx
@@ -9,5 +9,5 @@ export const JfrogArtifactoryDashboardPage = () => {
   const { entity } = useEntity();
   const { imageName } = useJfrogArtifactoryAppData({ entity });
 
-  return <JfrogArtifactoryRepository image={imageName} widget={false} />;
+  return <JfrogArtifactoryRepository image={imageName} />;
 };


invincibleJai:
Jobs are failing, can see it here https://github.com/janus-idp/backstage-plugins/actions/runs/5899055632/job/16097159967?pr=649

anilabhabaral:
@invincibleJai I will check and make the changes accordingly.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
```diff
@@ -80,11 +82,8 @@ export function JfrogArtifactoryRepository(props: RepositoryProps) {
...
 
-JfrogArtifactoryRepository.defaultProps = {
-  title: 'Docker Images',
-};
 interface RepositoryProps {
-  widget: boolean;
-  image: string;
-  title: string;
+  widget?: boolean;
+  image?: string;
```
invincibleJai, line 87: Will this is optional?

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
```diff
@@ -22,15 +22,17 @@ const useLocalStyles = makeStyles({
...
+) {
   const jfrogArtifactoryClient = useApi(jfrogArtifactoryApiRef);
   const classes = useStyles();
   const localClasses = useLocalStyles();
   const [edges, setEdges] = useState<Edge[]>([]);
   const title = `Jfrog Artifactory repository: ${props.image}`;
 
   const { loading } = useAsync(async () => {
-    const tagsResponse = await jfrogArtifactoryClient.getTags(props.image);
+    const tagsResponse = await jfrogArtifactoryClient.getTags(props.image!);
```
invincibleJai, line 35: this may not be needed.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
```diff
@@ -22,15 +22,17 @@ const useLocalStyles = makeStyles({
   },
 });
 
-export function JfrogArtifactoryRepository(props: RepositoryProps) {
+export function JfrogArtifactoryRepository(
+  props: RepositoryProps = { title: 'Docker Images' },
+) {
```
invincibleJai, line 27: how about below?

```suggestion
export function JfrogArtifactoryRepository(
 {widget , image, title = 'Docker images'}: RepositoryProps,
) {
```

invincibleJai:
@anilabhabaral Thanks, am not sure if you tested the plugin with changes good to test it too.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
anilabhabaral: This looks better but if i made the following changes then i also need to change the below marked line:
```
export function JfrogArtifactoryRepository({widget,image,title = 'Docker Images'}: RepositoryProps,) {
  const jfrogArtifactoryClient = useApi(jfrogArtifactoryApiRef);
  const classes = useStyles();
  const localClasses = useLocalStyles();
  const [edges, setEdges] = useState<Edge[]>([]);


  const title = `Jfrog Artifactory repository: ${props.image}`; <=== need to remove this line due to Duplicate identifier 'title' or we need to remove the const like "title = `Jfrog Artifactory repository: ${image}`;". But as per my understanding we should not remove constant as it may cause issue in runtime.

  const { loading } = useAsync(async () => {


    const tagsResponse = await jfrogArtifactoryClient.getTags(image); <=== here 'props.image' needs to replace by 'image'

    setEdges(tagsResponse.data.versions.edges);

    return tagsResponse;
  });
```

Please let me know if my understanding is correct or not.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
anilabhabaral: If i don't give the `!` then some tests are getting failed due to some undefined string value.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
anilabhabaral: I make the three interface values optional because I only need to pass title as default value.
in this format:
```
export function JfrogArtifactoryRepository(
  props: RepositoryProps = { title: 'Docker Images' },
) {
...
```

if i don't make those 3 optional it was causing some placeholder error as the three value of the interface needs to be passed with in {}. 
But i think my approach of making all interface properties optional is not right.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
invincibleJai: yeah you would need to use values directly over `props.*`

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
invincibleJai: with the above change that might not be needed, i think as it will be the same as before and you only pass the default value to the title

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
invincibleJai: follow this https://github.com/janus-idp/backstage-plugins/pull/649/files#r1302454640 and this change will not be needed so keep it as required



```
export function JfrogArtifactoryRepository(
  props: RepositoryProps = { title: 'Docker Images' },
) {
...
```

moreover, the above won't work as you are expecting it to, as if props will be defined then `{ title: 'Docker Images' }` will never be assigned. But what we want is props should pick the image and widget from what's coming as prop and if the title is undefined then pass the default value.

anilabhabaral:
@invincibleJai I made the suggested changes please check.

plugins/jfrog-artifactory/src/components/JfrogArtifactoryRepository/JfrogArtifactoryRepository.tsx
```diff
@@ -80,11 +80,8 @@ export function JfrogArtifactoryRepository(props: RepositoryProps) {
...
 }
 
-JfrogArtifactoryRepository.defaultProps = {
-  title: 'Docker Images',
-};
 interface RepositoryProps {
   widget: boolean;
   image: string;
-  title: string;
+  title?: string;
```
invincibleJai, line 86: as the component needs only an image, so can this type of definition just have that?

anilabhabaral: Here i see widget and image is used in JfrogArtifactoryDashboardPage: https://github.com/janus-idp/backstage-plugins/blob/359d7da547651a8686b4f2b754ec11f0b75b6829/plugins/jfrog-artifactory/src/components/JfrogArtifactoryDashboardPage/JfrogArtifactoryDashboardPage.tsx#L12C11-L12C37

I think we cannot remove widget but as per my understanding we can remove the title from definition.

invincibleJai: yeh, it is passed but not used in component `JfrogArtifactoryRepository` so can be removed from https://github.com/janus-idp/backstage-plugins/blob/359d7da547651a8686b4f2b754ec11f0b75b6829/plugins/jfrog-artifactory/src/components/JfrogArtifactoryDashboardPage/JfrogArtifactoryDashboardPage.tsx#L12C11-L12C37 and then from types as well.

anilabhabaral: Sure, i will make the suggested changes.

invincibleJai:
LGTM

PR #653, opened 2023-08-21
fix(aap): fixes codesmells for aap plugin

invincibleJai:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/589


tumido (approved):
/lgtm
/approve

PR #662, opened 2023-08-23
fix(aap): fix config loader issue on backstage v1.17

invincibleJai:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/658

### DIFF
--- a/plugins/aap-backend/config.d.ts
+++ b/plugins/aap-backend/config.d.ts
@@ -1,18 +1,18 @@
 import { TaskScheduleDefinitionConfig } from '@backstage/backend-tasks';
 
-interface AapConfig {
-  baseUrl: string;
-  /** @visibility secret */
-  authorization: string;
-  system?: string;
-  owner?: string;
-  schedule?: TaskScheduleDefinitionConfig;
-}
-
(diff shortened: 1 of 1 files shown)


tumido (approved):
/lgtm
/approve

plugins/aap-backend/config.d.ts
```diff
@@ -1,13 +1,4 @@
...
-
-interface AapConfig {
-  baseUrl: string;
-  /** @visibility secret */
-  authorization: string;
-  system?: string;
-  owner?: string;
-  schedule?: TaskScheduleDefinitionConfig;
-}
+import { AapConfig } from './src/types';
```
tumido, line 1: Hold on a minute... I'm not sure this works as intended... I'll need to take a look into the build process.

tumido: How does this get references when there's no `src` folder in the built NPM package and `config.d.ts` is not transpiled into JS? It seems this reference doesn't exist/never gets imported.

invincibleJai: Have updated it have inline

Issue #582, opened 2023-07-27
Move `icons`, `debounce` from Tekton and Topology plugins to `plugin/shared-react`


tumido:
Part of epic: https://github.com/janus-idp/backstage-plugins/issues/580

This mainly affects files for:
- `icons`
- `debounce`

Reference:
https://sonarcloud.io/component_measures?metric=duplicated_lines&view=list&id=janus-idp_backstage-plugins

https://sonarcloud.io/component_measures?metric=duplicated_files&view=list&id=janus-idp_backstage-plugins

tumido:
Icons were not moved, only `debounce`. Reopening.

https://github.com/janus-idp/backstage-plugins/blob/main/plugins/topology/src/common/components/icons.tsx

https://github.com/janus-idp/backstage-plugins/blob/main/plugins/tekton/src/components/common/icons.tsx

divyanshiGupta:
Logged a ticket https://github.com/janus-idp/backstage-plugins/issues/716 for refactoring `icons` to use MUI icons based on this converstaion - https://github.com/janus-idp/backstage-plugins/pull/715#discussion_r1317244432.

PR #636, opened 2023-08-12
chore(shared-react): updated plugins to use shared-react/utils

navinyashende:
@schultzp2020 @invincibleJai Could you please review?

Closes #599 

@schultzp2020 Probably another duplicate :smile:

PR #663, opened 2023-08-23
chore(shared-react): move debounce to shared-react

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/582

This PR moves removes duplicate code for debounce and useDeepCompareMemoize files from topology and tekton plugins.


plugins/shared-react/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "@kubernetes/client-node": "^0.18.1",
+    "@patternfly/react-icons": "^4.93.6",
```
debsmita1, line 27: We should avoid creating a dependency on patternfly in this package

debsmita1:
Could you create separate PRs to handle the changes in different plugins ?

plugins/shared-react/package.json
divyanshiGupta: In that case we might want to keep icons in the plugins itself. Wdyt?  @debsmita1

invincibleJai:
> Could you create separate PRs to handle the changes in different plugins ?

that might not be needed Check this comment from @tumido https://github.com/janus-idp/backstage-plugins/pull/636#discussion_r1297295428

plugins/shared-react/package.json
invincibleJai: can we use icons from MUI? over pf/react-icons? 

if yes we can keep it here else it's not ideal to bring in pf dependency in a shared package unless it's really needed

divyanshiGupta: I will add the similar icons from MUI then.

invincibleJai (approved):
LGTM

tumido:
@invincibleJai @debsmita1 since this PR was a `chore:`, it did not trigger release of the topology plugin. Please mind that this change will be introduced in the next release (with next change to topology plugin).

PR #668, opened 2023-08-23
docs: update tekton and topology documentation with pod log permissions

### LINKED ISSUE #667: Tekton and Topology: Prerequisites in documentation insufficient to view logs
## What needs to be done
The current documentation for the Tekton and Topology plugin requires the user to create a Service Account for the Kubernetes plugin. It references the [Kubernetes RBAC section](https://backstage.io/docs/features/kubernetes/configuration/#role-based-access-control) in the backstage documentation, which does not mention the permissions required to display the pod logs.
image

To remedy this, the user would also need to provide the following permission (specifically the `pods/log` permission) to the Service Account's `ClusterRole`:
```yaml
kind: ClusterRole
metadata:
  name: view-pod-logs
rules:
  # Base for Kubernetes plugin
  - apiGroups:
      - ""
    resources:
      - pods
      - pods/log
     verbs:
      - get
      - list
```

Zaperex:
Fixes #667

### DIFF
--- a/plugins/tekton/README.md
+++ b/plugins/tekton/README.md
@@ -23,9 +23,11 @@ The Tekton plugin enables you to visualize the `PipelineRun` resources available
          plural: 'taskruns'
   ```
 - The Kubernetes plugin is configured and connects to the cluster using a `ServiceAccount`.
-- The [`ClusterRole`](https://backstage.io/docs/features/kubernetes/configuration#role-based-access-control) must be granted for custom resources (PipelineRuns and TaskRuns) to `ServiceAccount` accessing the cluster. If you have the Backstage Kubernetes Plugin configured, then the `ClusterRole` is already granted.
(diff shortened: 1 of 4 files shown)


plugins/tekton/README.md
```diff
@@ -34,6 +36,14 @@ The Tekton plugin enables you to visualize the `PipelineRun` resources available
     metadata:
       name: backstage-read-only
     rules:
+      - apiGroups:
+          - ""
+        resources:
+          - pods/log
```
invincibleJai, line 42: is this needed, I don't think there is a resource `pod/log`.

cc @debsmita1

Zaperex: I wasn't able to get the logs to load without adding `pods/log` into the `ClusterRole`

invincibleJai: ah yeah my bad, this would be needed. 👍

invincibleJai:
@Zaperex Thanks for the PR, The Semantic job is failing it should be `docs: *` i think

debsmita1:
cc @hmanwani-rh

PR #673, opened 2023-08-25
fix(kiali): upgrade patternfly

aljesusg:
Require https://github.com/janus-idp/backstage-plugins/pull/630

- Upgrade patternfly library to version 5
- Improve documentation
- add Dev script


plugins/kiali/DEVELOPMENT.md
```diff
@@ -2,32 +2,66 @@
 
 ## Setup
 
+### Launch script and see what happen
```
tumido, line 5: Hi @aljesusg, we generally take a different approach to dev environments, allowing plugins to be developed independently. Can you please see if your changes/script can be adjusted to this approach?

We don't want plugins to change or even rely on `packages/` folder with the intention to later remove this folder entirely.

Instead we seek to implement standalone dev environments for individual plugins. For frontend we use following patterns for dev environments:

- With static data: https://github.com/janus-idp/backstage-plugins/blob/main/plugins/tekton/dev/index.tsx
- With data fetched from backend plugin: https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm/dev/index.tsx
- With data fetched via proxy: https://github.com/janus-idp/backstage-plugins/blob/main/plugins/quay/dev/index.tsx


For backend plugin it can be implemented as a dev environment:
- With data fetched from a standalone local/remote deployment like here https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/dev/index.ts

Would that be something that you can implement or you'd be interested in?

tumido (commented):
See my comment inline. It's not strictly blocking - we can use it as a future enhancement and migrate to it in future PRs, wdyt?

plugins/kiali/DEVELOPMENT.md
aljesusg: well. I'll take a look on this, The point is that some workmates are going to help me in kiali and I want to ake it easy for them

aljesusg: well @tumido I removed that to not block this and next pr and I'll try to make the dev environment with your instructions Thanks !

tumido (approved):
/lgtm
/approve

Issue #652, opened 2023-08-18
Ansible plugin should list workflow templates


cooktheryan:
**Is your feature request related to a problem? Please describe.**
Currently with the PR for the Ansible plugin we show Ansible Job templates. We should extend this functionality to include showing workflow templates. The api for that is /api/v2/workflow_job_templates/

**Describe the solution you'd like**
We should include the visualization to view workflow templates using a specific labeling of workflow template vs job templates which would have the job template label

**Describe alternatives you've considered**
N/A

**Additional context**
@invincibleJai and #522

invincibleJai:
Thanks @cooktheryan for creating the ticket, currently we show type as `job_template` , for worfklow template should we list these as type `workflow_template`?  and did you mean this when you say `We should include the visualization to view workflow templates using a specific labeling of workflow template vs job templates which would have the job template label`

FYI job_template type:

<img width="1568" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/d5728330-e13e-479e-8b25-b62d8a3b002a">

cc @Kasturi1820

cooktheryan:
I agree with you we should show the label type of job_template vs workflow_template where applicable. You were correct with what I was trying to say with the visualization

invincibleJai:
Here is PR for it https://github.com/janus-idp/backstage-plugins/pull/674

PR #674, opened 2023-08-25
feat(aap): support workflow job templates in catalog provider

invincibleJai:
Tracks: https://github.com/janus-idp/backstage-plugins/issues/652

**Description:**
Adds support workflow job templates in catalog provider and show separate type for job template and software job template as provided in API

image


plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
```diff
@@ -114,9 +117,23 @@ export class AapResourceEntityProvider implements EntityProvider {
...
+      this.baseUrl,
+      this.authorization,
+    ).catch(err => {
+      console.error('Failed to fetch AAP workflow job templates', err);
+    });
+
+    const templates = [
+      ...(jobTemplates || []),
+      ...(workflowJobTemplates || []),
+    ];
```
tumido, line 134: Can we fetch them in parallel instead of in sequential order? There's no point of async functions if we `await` each call in sequence 🙂 

Something like `Promise.all` should do.

invincibleJai: `Promise.all` is rejected if any of the elements are rejected but here my intent is even if the promise for JobTemplates fails but if WorkflowJobTemplate passes then the catalog should show the entities for that 🤔

invincibleJai: Thanks @tumido, have updated it and made use of `Promise.allSettled`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

tumido (approved):
/lgtm
/approve

PR #675, opened 2023-08-25
feat(kiali): add namespace selector

aljesusg:
Require: https://github.com/janus-idp/backstage-plugins/pull/673

Add a namespace Selector

image


aljesusg:
> @invincibleJai @divyanshiGupta @debsmita1 do we have any recommendation/opinion on the namespace selector design?
> 
> otherwise LGTM 👍

Let me explain this.

In kiali we work with namespaces,  (In fact we have a namespace selector in the standalone console.) By default user will see all namespaces related with the entity, but maybe the user want to see only the information of one namespace. (This will have more sense with the next feature: Graph )

invincibleJai:
> > @invincibleJai @divyanshiGupta @debsmita1 do we have any recommendation/opinion on the namespace selector design?
> > otherwise LGTM 👍
> 
> @tumido The namespace selector design looks fine to me based on the requirements. Also the selector is a backstage core component so there should not be any issues with design consistency. Tagging our UX team (@parvathyvr) here so that they can also add their input/feedback on the design.

Yeah as @divyanshiGupta suggested for now we can get this merged and follow up with UXD if there are suggestions/Improvements we can always take it as separate PR

tumido:
Thank you both @invincibleJai @divyanshiGupta 🙂 👍

yuvalgalanti:
@tumido @invincibleJai It looks excellent and it's a backstage component so I don't see any issue, thanks for sharing.

Screenshot 2023-08-30 at 16 17 24

PR #682, opened 2023-08-25
feat(ocm-backend): add support for new backend system

RobotSail:
Reupload of PR #459

Part of #238

Co-authored-by: Oleg S <97077423+RobotSail@users.noreply.github.com>
Signed-off-by: SamoKopecky <skopecky@redhat.com>


tumido (approved):
/lgtm
/approve

Issue #592, opened 2023-07-27
Reduce code smells in Kuberentes-action plugin


tumido:

anilabhabaral:
I would like to take this. Please assign @tumido

PR #686, opened 2023-08-29
fix(kubernetes-actions): fix code smells

anilabhabaral:
Resolves: [#592](https://github.com/janus-idp/backstage-plugins/issues/592)


plugins/kubernetes-actions/src/actions/createKubernetesNamespace.ts
```diff
@@ -68,7 +68,7 @@ const getUrlFromClusterRef = async (
 const validateUrl = (url: string | undefined) => {
   try {
     // eslint-disable-next-line no-new
-    new URL(url!);
+    new URL(url as string);
```
tumido, line 71: Can you please create a test case for this? https://github.com/janus-idp/backstage-plugins/blob/main/plugins/kubernetes-actions/src/actions/createKubernetesNamespace.test.ts

I'm not sure casting this to string is the best option. What about:
```suggestion
    new URL(url || "");
```

or even better, in the signature:

```diff
- const validateUrl = (url: string | undefined) => {
+ const validateUrl = (url: string | undefined = "") => {
```

tumido (changes requested):

plugins/kubernetes-actions/src/actions/createKubernetesNamespace.ts
anilabhabaral: Thanks for the suggeation @tumido. I believe `const validateUrl = (url: string | undefined = "") => {` would be best for this use case as it is by default assigning the empty string to undefined and we don't need to check/set the type of `url`.

anilabhabaral: I will also write the tests for empty string `""` and the `undefended` value to clarify the above.

tumido (approved):
/lgtm
/approve

anilabhabaral:
Thank You @tumido for your help!

PR #689, opened 2023-08-30
chore(segment): remove @deprecated comment in segment plugin

### LINKED ISSUE #590: Reduce code smells in Segment plugin

Zaperex:
Fixes #590

### DIFF
--- a/plugins/analytics-provider-segment/src/plugin.ts
+++ b/plugins/analytics-provider-segment/src/plugin.ts
@@ -16,10 +16,9 @@
 import { createPlugin } from '@backstage/core-plugin-api';
 
 /**
- * @deprecated Importing and including this plugin in an app has no effect.
- * This will be removed in a future release.
  *
- * @public
+ * @public Importing and including this plugin in an app has no effect.
+ * This will be removed in a future release.
  */
 export const analyticsModuleSegment = createPlugin({
   id: 'analytics-provider-segment',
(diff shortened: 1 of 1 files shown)


plugins/analytics-provider-segment/src/plugin.ts
```diff
@@ -16,10 +16,9 @@
 import { createPlugin } from '@backstage/core-plugin-api';
 
 /**
- * @deprecated Importing and including this plugin in an app has no effect.
- * This will be removed in a future release.
  *
- * @public
+ * @public Importing and including this plugin in an app has no effect.
+ * This will be removed in a future release.
```
tumido, line 21: This won't be removed in future release 🙂

tumido (changes requested):

plugins/analytics-provider-segment/src/plugin.ts
Zaperex: my bad :sweat_smile:

tumido (approved):
/lgtm
/approve

PR #693, opened 2023-08-31
fix(keycloak): fix keycloak user displayname

tumido:
Resolves: #274


tumido:
cc @schultzp2020

invincibleJai (approved):
LGTM

PR #696, opened 2023-08-31
docs(jfrog): update README.md

Zaperex:
Update the Jfrog documentation to not use ` instead of ' in the annotation example (in case user copy pastes it)
Relates to https://github.com/janus-idp/backstage-plugins/issues/559


invincibleJai (approved):
LGTM

PR #698, opened 2023-08-31
chore: updating vulnerabilities found in snyk scan

acornett21:
- these vulnerabilities were found during a scan with snyk


plugins/topology/package.json
```diff
@@ -41,7 +41,7 @@
     "@patternfly/react-icons": "^4.93.6",
     "@patternfly/react-styles": "^4.92.6",
     "@patternfly/react-tokens": "4.94.6",
-    "@patternfly/react-topology": "4.93.6",
+    "@patternfly/react-topology": "5.0.0",
```
tumido, line 44: This is a breaking change. We are handling this via https://github.com/janus-idp/backstage-plugins/pull/715

tumido (changes requested):
Thank you @acornett21, we should enable Snyk scans for ourselves, shouldn't we? 😄

There's always some app and scanner to add... 🤷 🙂

PR #706, opened 2023-09-01
chore(jfrog): update dev environment to not crash on load

### LINKED ISSUE #647: Jfrog-artifactory plugin standalone dev environment breaks on load
Jfrog-artifactory plugin standalone dev environment breaks on load

Jfrog-artifactory plugin: https://github.com/janus-idp/backstage-plugins/tree/main/plugins/jfrog-artifactory

Zaperex:
## Description
Updates the JFrog dev environment to not crash on load

## Which issue(s) does this PR fix
Fixes #647 

## Special Notes to Reviewers
This PR doesn't populate the JFrog dev environment with any data as I currently don't have an instance setup to provide obtain the mock data with, so the table will be empty.

### DIFF
--- a/plugins/jfrog-artifactory/dev/index.tsx
+++ b/plugins/jfrog-artifactory/dev/index.tsx
@@ -1,13 +1,36 @@
 import React from 'react';
 
+import { Entity } from '@backstage/catalog-model';
 import { createDevApp } from '@backstage/dev-utils';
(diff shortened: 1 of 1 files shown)


invincibleJai (approved):
LGTM

Thanks @Zaperex, let's add a ticket to add mock data as well (which can be a follow-up PR).

Issue #712, opened 2023-09-05
Nexus: Support Docker Manifest 2 Schema 1


rmartine-ias:
**Is your feature request related to a problem? Please describe.**
The nexus plugin does not currently support the (deprecated) Docker [manifest v2 schema 1 schema](https://docs.docker.com/registry/spec/manifest-v2-1/). This leads to errors from assuming certain fields exist.

In our case, this is because the request for the manifest does not seem to be sent with an explicit `Accept: application/vnd.docker.distribution.manifest.v2+json` header, but graceful fallback would still be useful. 

**Describe the solution you'd like**
Support for said schema.

**Describe alternatives you've considered**
Telling everyone to upgrade.

**Additional context**
<img width="845" alt="Screenshot 2023-09-05 at 2 47 08 PM" src="https://github.com/janus-idp/backstage-plugins/assets/107639398/c5764777-2769-4a4e-961a-4845d6cd1cc0">

PR #713, opened 2023-09-05
feat(nexus): support docker manifest 2 schema 1

rmartine-ias:
This adds support for the (deprecated) Docker [manifest v2 schema 1](https://docs.docker.com/registry/spec/manifest-v2-1/).

It may be more correct to display "unknown" for image size in this case, as the actual size cannot be calculated. If so, let me know and I will make the change.

Our nexus gives `0` for the asset size, so I assumed showing 0 for the total size makes it clear that it was unable to be calculated.

Closes https://github.com/janus-idp/backstage-plugins/issues/712

### DIFF
--- a/plugins/nexus-repository-manager/src/__fixtures__/repository/docker/v2/janus-idp/backstage-showcase/manifests/sha-33dfe6b.json
+++ b/plugins/nexus-repository-manager/src/__fixtures__/repository/docker/v2/janus-idp/backstage-showcase/manifests/sha-33dfe6b.json
@@ -1,41 +1,25 @@
 {
-  "schemaVersion": 2,
-  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
-  "config": {
-    "mediaType": "application/vnd.docker.container.image.v1+json",
-    "size": 17214,
-    "digest": "sha256:b98e210da4230ea59323687f2b6678e12f00a8c82a9433f8235a02e50a06c640"
-  },
-  "layers": [
+  "schemaVersion": 1,
+  "name": "janus-idp/backstage-showcase",
+  "tag": "sha-33dfe6b",
+  "fsLayers": [
     {
-      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
-      "size": 36581161,
-      "digest": "sha256:7890eb22610600843a22de84c96fab3f2d428d19e164a529d775ebbb22cc2f3e"
+      "blobSum": "sha256:7890eb22610600843a22de84c96fab3f2d428d19e164a529d775ebbb22cc2f3e"
     },
     {
-      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
-      "size": 36267603,
-      "digest": "sha256:cfdca8bd8795bb3e2c17030868e88434c52d55b7727a10187c9c0b7d0884daf0"
+      "blobSum": "sha256:cfdca8bd8795bb3e2c17030868e88434c52d55b7727a10187c9c0b7d0884daf0"
     },
     {
-      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
-      "size": 33954115,
-      "digest": "sha256:d47076810fbfe12fe7ffed77b6e7e314d8c2edec4d191cbf5e2297feed9bfd80"
+      "blobSum": "sha256:d47076810fbfe12fe7ffed77b6e7e314d8c2edec4d191cbf5e2297feed9bfd80"
     },
     {
-      "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
-      "size": 32,
-      "digest": "sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1"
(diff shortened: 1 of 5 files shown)


plugins/nexus-repository-manager/src/utils/get-file-size/get-file-size.test.ts
```diff
@@ -14,8 +14,9 @@ describe('getFileSize', () => {
     };
     const rawAssets = [
       {
-        schemaVersion: 2,
-        mediaType: '',
+        schemaVersion: 2 as const,
```
tplass-ias, line 17: maybe the `it()` should get a new description text also

plugins/nexus-repository-manager/src/__fixtures__/repository/docker/v2/janus-idp/backstage-showcase/manifests/sha-33dfe6b.json
```diff
@@ -1,41 +1,25 @@
 {
-  "schemaVersion": 2,
-  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
-  "config": {
-    "mediaType": "application/vnd.docker.container.image.v1+json",
-    "size": 17214,
-    "digest": "sha256:b98e210da4230ea59323687f2b6678e12f00a8c82a9433f8235a02e50a06c640"
-  },
-  "layers": [
+  "schemaVersion": 1,
```
tplass-ias, line 2: should these fixtures files have been duplicated and modified? I'm unsure where they are actually used 😅

rmartine-ias: These are used in [the dev app](https://github.com/rmartine-ias/janus-backstage-plugins/blob/docker-schema-1/plugins/nexus-repository-manager/dev/index.tsx#L26), and the [tests](https://github.com/rmartine-ias/janus-backstage-plugins/blob/docker-schema-1/plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.test.ts#L75). I wanted a faster feedback loop for testing, so I could get the error in the dev app. This is not necessarily still needed, the type system _should_ catch any more errors. Think I should revert this part?

tplass-ias: thanks, that makes sense. I think let's see if the maintainers have an opinion here

tumido: I think it makes sense to have this included as it is present in the PR. As @rmartine-ias wrote, the fast dev feedback is really nice thing to have. 🙂

plugins/nexus-repository-manager/src/types.ts
```diff
@@ -9,20 +9,41 @@ export type Annotation = {
...
 };
 
-export type RawAsset = {
-  schemaVersion: number;
-  mediaType: string;
-  config: Config;
-  layers: Layer[];
+export type RawAsset = RawAssetSchema1 | RawAssetSchema2;
+
+// https://docs.docker.com/registry/spec/manifest-v2-1/
```
tumido, line 14: Small nit: Can you please change this to:

```suggestion
/** @see {@link https://docs.docker.com/registry/spec/manifest-v2-1/|Image Manifest Version 2, Schema 1} */
```

This would allow people to get an IDE hint (in IDEs that support JSDoc) along with the type. 🙂

#### Before
image



#### After
image

plugins/nexus-repository-manager/src/utils/get-file-size/get-file-size.test.ts
tumido: I agree with @tplass-ias here, can you please update the testcase description here 👍

plugins/nexus-repository-manager/src/types.ts
```diff
@@ -9,20 +9,41 @@ export type Annotation = {
...
+
+export type LayerSchema1 = {
+  blobSum: string;
+};
+
+export type HistorySchema1 = {
+  v1Compatibility: string;
+};
+
+// https://docs.docker.com/registry/spec/manifest-v2-2/
```
tumido, line 32: Same nit as above.

tumido (commented):
Hi @rmartine-ias, 

very nice find! Thank you for your PR, it's so nice when people decide to contribute right away! 🙂 

It's a nice solution and your PR is in good shape. I have commented just a few nits to make it easier to consume by other devs.

The merit of the PR is fully fulfilled. 👍

plugins/nexus-repository-manager/src/types.ts
rmartine-ias: Oh, this is very nice! Will be using this in the future. Updated.

plugins/nexus-repository-manager/src/utils/get-file-size/get-file-size.test.ts
rmartine-ias: done 👍

plugins/nexus-repository-manager/src/types.ts
rmartine-ias: updated!

tumido (approved):
/lgtm
/approve

Issue #711, opened 2023-09-05
Nexus plugin is slow to load docker artifacts


rmartine-ias:
**Describe the bug**
When loading the "Build Artifacts" tab for the first time (in each browser), with a `nexus-repository-manager/docker.image-name:` annotation, the tab takes around 20 seconds to load. This is because of the requests for the docker manifests, which take 6-7 seconds each, on our Nexus. This is past the point where people will click away because it is too slow.

**To Reproduce**

1. Initialize plugin 
2. Add a `nexus-repository-manager/docker.image-name:` annotation to a catalog item
3. Click on the "Build Artifacts" tab (may need to open devtools and disable cache)
4. See time to load

**Expected behavior**

Loading to be near-instant. If possible, load times to be shared among users, rather than repeated per user. Thoughts on how to do this:
* Add a backend to the plugin, with either a request-level or data-level cache. Refresh this on some schedule
* Add config to only fetch the latest N artifacts of each type
* Find a way to not request the full manifests, or at least not block on it -- we should be able to display a list of tags without waiting for the requests to complete. I think only artifact size requires the manifest?
* Suggest adding Varnish or similar in-between backstage and nexus (we'd rather not do this, if possible)
* Two or more of the above

**Screenshots**
<img width="1405" alt="Screenshot 2023-09-05 at 12 49 03 PM" src="https://github.com/janus-idp/backstage-plugins/assets/107639398/9142d230-45a5-40b5-bea3-1a3414b2cd7d">

**Additional context**

My company was planning on creating our own Nexus plugin, but delayed long enough that y'all shipped an OSS one. Since we've budgeted time for this already, we'd like to spend it working to contribute back to the work you're doing. I'm reporting this issue to see if there's something you already have planned for this/start a discussion, so we don't duplicate work you're already doing, or go down a different route than you envisioned.

tumido:
Hey @rmartine-ias,

Thank you for filing an issue here! 🙂 

If you have more ideas or if you would like to steer this plugin in a specific direction or design, please share it! We welcome it, we are open to any suggestions. 🙂 

If you'd like, you can also join us on [Slack](https://join.slack.com/t/janus-idp/shared_invite/zt-1pxtehxom-fCFtF9rRe3vFqUiFFeAkmg) and for our [bi-weekly community calls](https://calendar.google.com/calendar/u/0?cid=amFudXMuaWRwLmNvbW11bml0eUBnbWFpbC5jb20).


Back to the issue:
[...]

PR #719, opened 2023-09-08
feat(nexus) Set accept headers to prefer docker schema 2 v2 when available

rmartine-ias:
Requesting a manifest [without certain accept headers](https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility) makes (our) Nexus think only manifest schema 1 is supported, and on-the-fly rewriting to schema 1 is slow. This PR sets the accept headers to include `application/vnd.docker.distribution.manifest.v2+json` when requesting a Docker manifest.

This uses [request assertions](https://mswjs.io/docs/recipes/request-assertions) in the tests, which is discouraged in the MSW docs. I've written tests that don't use those as well; [this commit](https://github.com/janus-idp/backstage-plugins/commit/4df14dba8106903447b5b794f322d6bf96e6f4c4) can be dropped if we want to follow the suggestions. I chose to do this because it seems much more clear to me to test "we are requesting docker manifests, and nothing else, with these headers" rather than "assuming the mock server is set up correctly, calling this function should hit an endpoint that will conditionally return different files based on the accept headers, and so getting this file instead of that one means we must've set the headers right."

Based on commits from: https://github.com/janus-idp/backstage-plugins/pull/713

See also: https://github.com/janus-idp/backstage-plugins/issues/711


plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.ts
```diff
@@ -74,11 +87,15 @@ export class NexusRepositoryManagerApiClient
     return await SearchService.search(query);
   }
 
-  private async fetcher(url: string) {
+  private async fetcher(
+    url: string,
+    additionalHeaders: Record<string, string> = {},
```
rmartine-ias, line 92: I am not sure if this is the right level of abstraction (should this be `accept: string = "application/json"`), the right type (maybe [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers)?), or the idiomatic way to do optional arguments (`additionalHeaders?: Record<string, string>` with `...(additionalHeaders ?? {})`).

Feedback here would be appreciated.

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.test.ts
```diff
@@ -65,28 +65,42 @@ const handlers = [
     },
   ),
 
+  // Always returns manifest v2 schema 1, regardless of the accept header, to
+  // simulate a server that does not support schema 2.
```
rmartine-ias, line 69: This has created a weird hybrid server that both does and doesn't support manifest schema 2 v2. It might be better to have two separate servers -- need to think on this.

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.ts
schultzp2020: Here is how I would update the `fetcher` method.

```ts
private async fetcher(url: string, additionalHeaders: HeadersInit = {}) {
  const { token: idToken } = await this.identityApi.getCredentials();

  const headers = new Headers(additionalHeaders);
  headers.set('Content-Type', 'application/json');
  if (idToken) {
    headers.set('Authorization', `Bearer ${idToken}`);
  }

  const response = await fetch(url, {
    headers,
  });
  if (!response.ok) {
    throw new Error(
      `failed to fetch data, status ${response.status}: ${response.statusText}`,
    );
  }
  return await response.json();
}
```

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.test.ts
rmartine-ias: Tried having two separate servers, it was too much code & complexity for a somewhat-simple test, and seemed like it wasn't forward-thinking enough for when there's support for other formats. I'd suggest revisiting this after adding support for maven artifacts (next on our list).

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.ts
```diff
@@ -23,6 +23,19 @@ const NEXUS_REPOSITORY_MANAGER_CONFIG = {
...
+ * for supporting servers to return and contains size information.
+ * @see @link{https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility|Backward compatibility}
+ */
+const DOCKER_MANIFEST_HEADERS = {
+  Accept: [
+    'application/vnd.docker.distribution.manifest.v2+json',
+    'application/vnd.docker.distribution.manifest.v1+json;q=0.9',
+    '*/*;q=0.8',
+  ].join(', '),
+};
```
schultzp2020, line 37:
```suggestion
} as const satisfies HeadersInit;
```

When possible, try to use a const assertion for true constant variables.

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.ts
```diff
@@ -99,13 +117,20 @@ export class NexusRepositoryManagerApiClient
...
 
   private async getRawAssets(component: ComponentXO) {
     const assets = await Promise.all(
       component.assets?.map(
-        async asset => await this.getRawAsset(asset.downloadUrl),
+        async asset =>
+          await this.getRawAsset(
+            asset.downloadUrl,
+            component.format === 'docker' ? DOCKER_MANIFEST_HEADERS : {},
+          ),
```
schultzp2020, line 133: Sadly I cannot make this a git suggestion; however, I would rewrite `getRawAssets` with a header helper function as the following.

```ts
function getAdditionalHeaders(format?: string): HeadersInit {
  switch (format /* NOSONAR - use switch for expandability */) {
    case 'docker':
      return DOCKER_MANIFEST_HEADERS;
    default:
      return {};
  }
}

private async getRawAssets(component: ComponentXO) {
  const additionalHeaders = getAdditionalHeaders(component.format);

  const assets = await Promise.all(
    component.assets?.map(
      async asset =>
        await this.getRawAsset(asset.downloadUrl, additionalHeaders),
      // Create a dummy promise to avoid Promise.all() from failing
    ) ?? [new Promise<null>(() => null)],
  );

  return assets;
}
```

schultzp2020 (changes requested):

schultzp2020:
@rmartine-ias you can ignore the SonarCloud Code Analysis check. I'm not too sure why it's failing when the code it doesn't like doesn't change in your PR.

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.ts
rmartine-ias: Got it, read up on the differences between `const x` and `...as const` and use of `satisfies`, thank you for the catch!

plugins/nexus-repository-manager/src/api/nexus-repository-manager-api-client/nexus-repository-manager-api-client.ts
rmartine-ias: Makes sense, I was thinking it was better to wait to see if Docker is a special case, but it does look ugly on its own. Will update.

schultzp2020 (approved):
LGTM

tumido (approved):
/lgtm
/approve

Looks good to me, let's merge it! 🙂 👍

PR #728, opened 2023-09-13
feat(rbac): implement RBAC group support

AndrienkoAleksandr:
### Referenced issue:

Fixes: janus-idp/backstage-showcase#488

 ### Demo:
 [RBAC groups support for rh-rbac-backend-plugin](https://youtu.be/2c1FLB8fJTQ)

#### Short Conclusions about Possible Implementations

To support RBAC hierarchy groups, we need to retrieve information about these groups. There are two ways:

- Retrieve all group hierarchy.
- Retrieve ancestor groups for the current user.

##### Current Implementation

In the current implementation, I used the second way: we retrieve ancestor groups for the user and evaluate permissions for requests at runtime.

##### Alternative Ways

We can retrieve all group hierarchy information at the start of the backstage instance and store this information.

There are two ways to store this information:

- We can store it in the runtime cache.
- We can store it in the rh-rbac-backend-plugin database.

But in both cases, we are facing an issue: we need to keep this information up to date. We could try to update this information using a scheduled task, but then we need to find a way to get the group hierarchy diff. This solution could have performance issues because we need to retrieve information about all groups, and information about deleted groups or changed group hierarchies will not be applied immediately. Another good alternative approach is to subscribe to the catalog API events and track all changes related to groups. However, this API has not been implemented yet (backstage/backstage/issues/8219).

# Todo:
1. discuss implementation.
2. If current solution is close to goal:
  - handle todos.
  - fix test and write new tests.

#### Test catalog

So I have catalog with groups and users: https://github.com/AndrienkoAleksandr/catalog-model
Group and users defined in the folder examples/acme

For testing purposes you can fork catalog: https://github.com/AndrienkoAleksandr/catalog-model or customize your own.

In this catalog I have a groups hierarchy for testing purpose:

acme-corp
    |
infrastructure
    |
backstage
    |
  team-a

#### Test users

Group team-a has some users. I used two of them:

- AndrienkoAleksandr - permission admin.
  For this user I used github auth provider to login.
  AndrienkoAleksandr is my github account user name.
  I set up github auth provider to match github account with catalog entity user with help of metadata.name:
  https://github.com/AndrienkoAleksandr/catalog-model/blob/60f2ffdfdb4cf4e0b35aa308b63c2909195cc8d5/examples/acme/team-a-group.yaml#L18
- Logarifm - simple user.
  For this user I used bitbucket auth provider to login.
  Logarifm is my bitbucket account user name.
  I set up bithbucket auth provider to match bitbucket account with catalog entity with help of annotation:
  https://github.com/AndrienkoAleksandr/catalog-model/blob/60f2ffdfdb4cf4e0b35aa308b63c2909195cc8d5/examples/acme/team-a-group.yaml#L31

I defined catalog using application config:

```app-config.local.yaml
...
catalog:
...
  rules:
    - allow: [Component, System, Group, User, Resource, Location, Template, API]
  locations:
    # Note: integrations.github[].apps must be correctly configured to read GitHub locations
    # Uncomment these lines to add more example data
    - type: url
      target: https://github.com/AndrienkoAleksandr/catalog-model/tree/rbac/examples/all.yaml

    # Uncomment these lines to add an example org
    - type: url
      target: https://github.com/AndrienkoAleksandr/catalog-model/tree/rbac/examples/acme-corp.yaml
      rules:
        - allow: [User, Group]
...
```

For two test users I defined auth providers for user login:

```app-config.local.yaml
...
auth:
  # see https://backstage.io/docs/auth/ to learn about auth providers
  environment: development
  providers:
    # Plugin: GitHub
    github:
      development:
        clientId: ${GITHUB_BUCKET_CLIENT_ID}
        clientSecret: ${GITHUB_BUCKET_SECRET}
    bitbucket:
      development:
        clientId: ${BIT_BUCKET_CLIENT_ID}
        clientSecret: ${BIT_BUCKET_SECRET}
...
```

Frontend login configuration patch with github resolver `usernameMatchingUserEntityName` and bitbucket resolver `usernameMatchingUserEntityAnnotation`:

```packages/backend/src/plugins/auth.ts
github: providers.github.create({
  signIn: {
-          resolver(_, ctx) {
-            const userRef = 'user:default/guest'; // Must be a full entity reference
-            return ctx.issueToken({
-              claims: {
-                sub: userRef, // The user's own identity
-                ent: [userRef], // A list of identities that the user claims ownership through
-              },
-            });
-          },
-          // resolver: providers.github.resolvers.usernameMatchingUserEntityName(),
+          resolver: providers.github.resolvers.usernameMatchingUserEntityName(),
+        },
+      }),
+      bitbucket: providers.bitbucket.create({
+        signIn: {
+          resolver:
+            providers.bitbucket.resolvers.usernameMatchingUserEntityAnnotation(),
  },
}),
...
```

Backend configuration patch:

```App.tsx
const app = createApp({
+  components: {
+    SignInPage: props => (
+      <SignInPage
+        {...props}
+        auto
+        providers={[
+          {
+            id: 'bitbucket-auth-provider',
+            title: 'Bitbucket',
+            message: 'Sign In using Bitbucket',
+            apiRef: bitbucketAuthApiRef,
+          },
+          {
+            id: 'github-auth-provider',
+            title: 'GitHub',
+            message: 'Sign in using GitHub',
+            apiRef: githubAuthApiRef,
+          },
+        ]}
+      />
+    ),
+  },
```

Enable rh-rback-backend-plugin plugin in the application configuration and
define your permission admin user/s with help of entity reference
(https://backstage.io/docs/features/software-catalog/references/#string-references):

```app-config.local.yaml
...
permission:
  enabled: true
  rbac:
    # Configuration with list predefined permission policies in casbin format.
    # policies-csv-file: /some/path/rbac-policy.csv
    database:
      enabled: true
    admin:
      users:
        - name: user:default/andrienkoaleksandr
...
```

Demo script:

1. Login like permission admin using github auth.
2. Retrieve admin user token. Copy it to terminal:

export token='...token...value'

3. Logout. Login like bitbucket user. Go to the catalog tab. It should be empty.
4. Provide read catalog permission policy for team-a:

curl -X POST "http://localhost:7007/api/permission/policies" -d '{"entityReference": "group:default/team-a", "permission": "catalog-entity", "policy": "read", "effect":"allow"}' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v

5. Go to catalog tab. List catalogs should appear, because user inherits permission policies from team-a
6. Select group filter. Select team-a group. Try to refresh button. Backstage should fail.
7. Provide refresh catalog entity permission, but for user:

curl -X POST "http://localhost:7007/api/permission/policies" -d '{"entityReference": "user:default/logarifm", "permission": "catalog-entity", "policy": "update", "effect":"allow"}' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v

8. Check the refresh button. It should work.
9. Check unregister catalog entity option. It should be disabled.
10. Provide catalog entity unregister permission policy, but for group backstage:

curl -X POST "http://localhost:7007/api/permission/policies" -d '{"entityReference": "group:default/backstage", "permission": "catalog-entity", "policy": "delete", "effect":"allow"}' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v

11. Check unregister catalog entity option. It should work.
12. Go to github: https://github.com/AndrienkoAleksandr/catalog-model/blob/rbac/examples/acme/team-a-group.yaml#L12 . Change team-a parent from backstage to infrastructure. Commit changes.
13. Go to github: https://github.com/AndrienkoAleksandr/catalog-model/blob/rbac/examples/acme/backstage-group.yaml#L13. Remove team-a children.
14. Back to backstage instance, check in the catalogs group team-a and backstage. Since some period of time changes from github should appear in the
backstage.
15. Check: bitbucket user should lose unregister catalog entity permission policy.


gorkem (changes requested):
Can you check if this can be implemented with a custom casbin `RoleManager` more cleanly?

Issue #731, opened 2023-09-14
The color of the PipelineRun status in the Topology Plugin is grey in color


debsmita1:
**Describe the bug**
The color of the pipeline run status on the pipelinerun node decorator and on the side-panel is grey in color for Failed and Successful statuses 

**Expected behavior**
The Failed status should show Red exclamation icon
The Successful status should show Green check icon

**Screenshots**
Topology Sidepanel resources
<img width="316" alt="Screenshot 2023-09-13 at 6 11 31 PM" src="https://github.com/janus-idp/backstage-plugins/assets/22490998/20d85aa0-f366-4aed-89c3-6c5633b4ab13">

Topology PipelineRun node decorator
<img width="316" alt="Screenshot 2023-09-13 at 6 11 42 PM" src="https://github.com/janus-idp/backstage-plugins/assets/22490998/9ffbc580-5e4e-4986-81fc-b458fc45d6d9">

**Additional context**

PR #732, opened 2023-09-14
fix(topology): fix pipelinerun status icon color

debsmita1:
Resolves:
https://github.com/janus-idp/backstage-plugins/issues/731

Screenshot:
Screenshot 2023-09-14 at 1 42 33 PM

Screenshot 2023-09-14 at 1 41 55 PM


plugins/topology/src/common/components/Status.tsx
```diff
@@ -63,13 +65,20 @@ const Status = ({ status, noTooltip, iconOnly }: StatusProps) => {
       return (
         <StatusIconAndText
           {...statusProps}
-          icon={<RedExclamationCircleIcon />}
+          icon={
+            <ExclamationCircleIcon className="bs-topology-icons__red-exclamation-icon" />
```
divyanshiGupta, line 69: @debsmita1 since you are using pf v5 color palette for icons in tekton PR - https://github.com/janus-idp/backstage-plugins/pull/734/files#diff-3b7c01df880a9b9228daac5520b8a879b90a16e9ca5aab39dc751f23716344acR102 so to keep it consistent we should use pf v5 color palette here as well.

divyanshiGupta (changes requested):

plugins/topology/src/common/components/Status.tsx
debsmita1: Yes, these classes are using the pf5 color palette

```
Status.css


.bs-topology-status__green-check-icon {
  fill: var(--pf-v5-global--palette--green-500) !important;
}

.bs-topology-status__red-exclamation-icon {
  fill: var(--pf-v5-global--danger-color--100) !important;
}

divyanshiGupta:
LGTM

divyanshiGupta (approved):

Issue #733, opened 2023-09-14
The color of the PipelineRun status in the Tekton Plugin is grey in color 


debsmita1:
**Describe the bug**
The color of the pipeline run status on the pipelinerun visualization is grey in color for Failed and Successful statuses 

**Expected behavior**
The Failed status should show Red exclamation icon with text
The Successful status should show Green check icon with text

**Screenshots**
<img width="509" alt="Screenshot 2023-09-14 at 1 46 19 PM" src="https://github.com/janus-idp/backstage-plugins/assets/22490998/e95b4389-d77c-4c54-a1fd-0b0470a6f46a">

**Additional context**

PR #734, opened 2023-09-14
fix(tekton): fix pipelinerun status icon color

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/733

**Screenshot:**
Screenshot 2023-09-14 at 7 12 42 PM


plugins/tekton/src/components/common/Status.tsx
```diff
@@ -2,7 +2,10 @@ import React from 'react';
 
 import {
   BanIcon,
+  CheckCircleIcon,
```
invincibleJai, line 5: why we have two `Status` and `StatusIcon`, both look similar?

plugins/tekton/src/components/common/Status.tsx
```diff
@@ -100,7 +97,11 @@ export const Status = ({
       return (
         <StatusIconAndText
           {...statusProps}
-          icon={<YellowExclamationTriangleIcon />}
+          icon={
+            <ExclamationTriangleIcon
+              style={{ color: 'var(--pf-v5-global--warning-color--100)' }}
```
divyanshiGupta, line 102: @debsmita1 can we create separate icon classes and use them as needed in `Status` and `StatusIcon` similar to topology so that in future if we want to change the styles they can be applied to all the places these icons are being used? Also another thing I noticed is that we use `fill` property in topology but you are using `color` here, any reason for this inconsistency or can we use same property to apply icon colors in both plugins?

divyanshiGupta (changes requested):

plugins/tekton/src/components/common/Status.tsx
debsmita1: The `Status` component was for displaying status icon and text and the `StatusIcon` component was to display only the icon. However, I have created a common component. PTAL

invincibleJai:
@debsmita1 pls rebase it

debsmita1:
@invincibleJai @divyanshiGupta PTAL

plugins/topology/src/components/Graph/decorators/PipelineRunDecorator.tsx
```diff
@@ -37,7 +37,7 @@ export const PipelineRunDecorator = ({
   );
 
   const status = pipelineRunStatus(latestPipelineRun);
-  const statusIcon = <Status status={status} iconOnly noTooltip />;
+  const statusIcon = status && <Status status={status} iconOnly />;
```
invincibleJai, line 40: do we need check for  `status &&`? Status component does handle it right

invincibleJai:
@debsmita1 can you add/update unit tests as well?

plugins/shared-react/src/components/common/StatusIconAndText.tsx
```diff
@@ -23,28 +21,37 @@ export const StatusIconAndText = ({
...
-        })}
-      {!iconOnly && <CamelCaseWrap value={title} dataTest="status-text" />}
+      {React.cloneElement(icon, {
+        className: classNames(
+          spin && 'fa-spin',
+          icon.props.className,
+          'bs-shared-icon-and-text__icon bs-shared-icon-flex-child',
+        ),
+      })}
+      <CamelCaseWrap value={title} dataTest="status-text" />
```
invincibleJai, line 54: nit:

```suggestion
      <CamelCaseWrap value={title} data-testid=="status-text" />
```

plugins/tekton/src/components/common/Status.tsx
```diff
@@ -1,78 +1,93 @@
...
-  noTooltip,
   className,
-}: React.PropsWithChildren<StatusProps>) => {
+  height,
+  width,
+  spin,
+}: React.PropsWithChildren<StatusProps>): React.ReactElement => {
   const statusProps = {
-    title: title || status,
+    title: status ?? '',
```
invincibleJai, line 57: This looks confusing, why did we remove the title prop and pass just the status to the title?

plugins/topology/src/components/Graph/decorators/PipelineRunDecorator.tsx
```diff
@@ -37,7 +37,7 @@ export const PipelineRunDecorator = ({
   );
 
   const status = pipelineRunStatus(latestPipelineRun);
-  const statusIcon = <Status status={status} iconOnly noTooltip />;
+  const statusIcon = <Status status={status as string} iconOnly />;
```
invincibleJai, line 40: this shouldn't be needed i.e in component `Status` status prop is required one and from above you might get `null` so you needed to typecast

plugins/tekton/src/components/common/Status.tsx
debsmita1: Because we are not passing any value to the title prop anywhere when calling Status

plugins/shared-react/src/components/common/StatusIconAndText.tsx
invincibleJai: ah i see it now , it's passed just as prop to children

invincibleJai (approved):
LGTM

PR #747, opened 2023-09-19
feat(cli): add frontend dynamic plugins base build config

Hyperkid123:


tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "@backstage/cli/config/tsconfig.json",
+  "include": ["packages/*/src", "plugins/*/src"],
+  "compilerOptions": {
+    "outDir": "dist-types",
+    "rootDir": ".",
+    "useUnknownInCatchVariables": false
+  }
+}
```
tumido, line 9: Why are we introducing a root level `tsconfig.json` if we have a per-package one?

tumido (commented):
I have one question inline, otherwise lgtm.

tsconfig.json
Hyperkid123: I am not sure yet. I was running into issues in the CLI where it was complaining about missing root config. I've added it temporarily to unblock the rest of the work. I'll get back to it once I am done with the configurations and experiments.

Hyperkid123: OK, there was a ts-node process where you can configure a config directly. I've removed it for now. We might have to add the config later on and reference the plugin-specific `tsconfig.json`.

Hyperkid123:
The `app-next` and dynamic plugin will be removed before merging.

tumido (approved):
/lgtm
/approve

PR #778, opened 2023-09-22
chore: simplify a tiny bit and optimize release process

tumido:
`conventional-changelog-conventionalcommits` v7 seems to be causing troubles since new  `semantic-release` require ESM and drops CommonJS. We use `semantic-release` via `multi-semantic-release` that doesn't support ESM just yet. (`@semrel-extra/npm` also doesn't support ESM at the moment)

This PR also adds a better user experience for PR title validation - a comment is posted/removed based on the CI result. We also do not enforce PR title and commit matching when there's a single commit only and in addition we've enabled support for `[WIP]` PRs, so we don't see failing CI on those anymore.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -18,7 +18,7 @@
     },
     {
       "matchPackagePatterns": ["^@backstage/"],
-      "groupName": ["Core Backstage packages"]
+      "groupName": "Core Backstage packages"
     },
     {
       "matchFileNames": ["packages/**"],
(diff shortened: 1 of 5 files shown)


.github/renovate.json
```diff
@@ -27,6 +27,14 @@
     {
       "matchFileNames": ["plugins/**"],
       "extends": [":pinOnlyDevDependencies"]
+    },
+    {
+      "matchPackagePatterns": [
+        "^@semantic-release/",
+        "^@@semrel-extra/",
```
schultzp2020, line 34: Is it supposed to be two `@` signs?

schultzp2020 (approved):
LGTM

PR #670, opened 2023-08-23
List perm policies backend method

AndrienkoAleksandr:

PR #779, opened 2023-09-22
chore: align RBAC plugin name with convention

tumido:
cc @gorkem @AndrienkoAleksandr 

Let's align the plugin package names with the rest of the plugins available in this repository and in the whole Janus community:

${\texttt{@janus-idp/{\color{green}backstage}-plugin-{\color{red}rh-}rbac-*}}$

Also:
- adds @gorkem and @AndrienkoAleksandr as `CODEOWNERS`


AndrienkoAleksandr:
Sorry, but we don't follow ADDR https://backstage.io/docs/architecture-decisions/adrs-adr011/#decision:

> x-backend: Contains the main backend code of the plugin.
> @backstage/plugin-

From point of view convention our plugin has a valid name...

AndrienkoAleksandr:
> I don't think Backstage Community ADRs apply to Janus IDP community.

Hmm. OK, after this comment I agree with @Gorkem concert about implementation: https://github.com/janus-idp/backstage-plugins/pull/670 => https://github.com/janus-idp/backstage-plugins/pull/670#discussion_r1316085412

.github/CODEOWNERS
```diff
@@ -25,4 +25,6 @@ yarn.lock                    @janus-idp/maintainers-plugins
 /plugins/quay/               @janus-idp/maintainers-plugins @fmenesesg
 /plugins/topology/           @janus-idp/maintainers-plugins @debsmita1 @divyanshiGupta
 /plugins/tekton/             @janus-idp/maintainers-plugins @debsmita1 @divyanshiGupta
+/plugins/rbac-backend/       @janus-idp/maintainers-plugins @gorkem @AndrienkoAleksandr
```
AndrienkoAleksandr, line 28: Could you apply @PatAKnight for these plugins too?

tumido:
Oh.. when it comes to plugin names, they can be really wild. The ADR you mention is followed only by core packages in `@backstage` scope. For example Roadie plugins use the same schema as Janus IDP. https://www.npmjs.com/search?q=%40roadiehq%2F

.github/CODEOWNERS
tumido: Done 🙂

AndrienkoAleksandr (approved):
I re-tested. Plugin works fine after rename.

PR #782, opened 2023-09-22
feat(cli): switch to @janus-idp/cli for new plugins and populate them with the basics

tumido:
- Introduce a new option to `janus-cli new` that allows us to skip changes to `packages/*`
- Remove `--scope` option and hard code to `@janus-idp`
- Update templates to contain metadata in their `package.json`, fix `tsconfig.json` and add `turbo.json`.
- Update dependency resolution to pin `devDependencies`.


tumido:
do not merge, rebase and update needed 🙂

tumido:
cc @davidfestal @kadel

can you please take a look?

packages/cli/templates/default-plugin/package.json.hbs
```diff
@@ -22,20 +22,21 @@
   "scripts": {
     "start": "backstage-cli package start",
     "build": "backstage-cli package build",
+    "export-dynamic": "janus-cli package export-dynamic-plugin",
     "lint": "backstage-cli package lint",
     "test": "backstage-cli package test",
     "clean": "backstage-cli package clean",
     "prepack": "backstage-cli package prepack",
     "postpack": "backstage-cli package postpack"
```
davidfestal, line 30: we will need the preversion script here, as in https://github.com/janus-idp/backstage-plugins/blob/main/plugins/3scale-backend/package.json#L21, if we want the `semantic-release` to work correctly.

davidfestal (changes requested):

packages/cli/templates/default-backend-plugin/src/dynamic/alpha.ts.hbs
```diff
@@ -0,0 +1,8 @@
+import { BackendDynamicPluginInstaller } from '@backstage/backend-plugin-manager';
+
+import { {{id}}Plugin } from '../service/router';
```
davidfestal, line 3:
```suggestion
import { {{id}}Plugin } from '../plugin';
```

packages/cli/templates/default-plugin/app-config.janus-idp.yaml
davidfestal, line 1: Do we want to add a skeleton for the typical config that allow placing this UI plugin in the frontend application (possibly with comments) ?

packages/cli/templates/default-backend-plugin/src/dynamic/alpha.ts.hbs
tumido: There is no `../plugin` file in `default-backend-plugin`.

packages/cli/templates/default-plugin/app-config.janus-idp.yaml
tumido: Not sure. What would you like to see there? 🙂

davidfestal: something like:

``` yaml
frontend:
  <generated plugin-related name>:
    # Description pointing to the showcase docs
    dynamicRoutes: []
    mountPoints: []
    routeBindings: []
    appIcons: []
    apiFactories: []
```
?

packages/cli/templates/default-backend-plugin/src/dynamic/alpha.ts.hbs
davidfestal: Same as in the `default-backend-module` isn't it: https://github.com/janus-idp/backstage-plugins/pull/782/files#diff-a1c7ed3013180570da88302938a6eb7c30d0cde2568ef89aa6ce791da224d3b6 ?

tumido: Not sure what you're referring to.

[`default-backend-module`](https://github.com/tumido/backstage-plugins/tree/update-new-plugin-tooling/packages/cli/templates/default-backend-module) has `src/module.ts` with the actual module code, therefore it imports `../module` in `src/dynamic/alpha.ts`

[`default-backend-plugin`](https://github.com/tumido/backstage-plugins/tree/update-new-plugin-tooling/packages/cli/templates/default-backend-plugin) has `src/service/router.ts` with the actual module integration, therefore it imports `../service/router` in `src/dynamic/alpha.ts`

Neither of those templates have `src/plugin.ts`, that file is traditionally used in frontend plugins only. None of the backend plugin templates has this file, here or upstream. 🤷

packages/cli/templates/default-plugin/app-config.janus-idp.yaml
tumido: Done. 👍

packages/cli/templates/default-plugin/package.json.hbs
```diff
@@ -20,22 +20,25 @@
...
+    "@backstage/core-components": "^{{versionQuery '@backstage/core-components'}}",
+    "@backstage/core-plugin-api": "^{{versionQuery '@backstage/core-plugin-api'}}",
+    "@backstage/theme": "^{{versionQuery '@backstage/theme'}}",
+    "@material-ui/core": "^{{versionQuery '@material-ui/core' '4.12.2'}}",
+    "@material-ui/icons": "^{{versionQuery '@material-ui/icons' '4.9.1'}}",
+    "@material-ui/lab": "^{{versionQuery '@material-ui/lab' '4.0.0-alpha.61'}}",
+    "react-use": "^{{versionQuery 'react-use' '17.2.4'}}"
   },
   "peerDependencies": {
     "react": "{{versionQuery 'react' '^16.13.1 || ^17.0.0'}}"
```
gashcrumb, line 44: I wonder if this would be a good time to also allow for React 18 and change this to:

`"react": "{{versionQuery 'react' '^16.13.1 || ^17.0.0 || ^18.0.0'}}"`

packages/cli/templates/default-backend-plugin/src/service/router.ts.hbs
```diff
@@ -0,0 +1,40 @@
...
+  router.use(express.json());
+
+  router.get('/health', (_, response) => {
+    logger.info('PONG!');
+    response.json({ status: 'ok' });
+  });
+  router.use(errorHandler());
+  return router;
+}
+export const {{id}}Plugin = createBackendPlugin({
```
gashcrumb, line 26: Seems like this ought to go in a `plugin.ts.hbs` which then would match up with @davidfestal's suggestion.

packages/cli/templates/default-plugin/package.json.hbs
tumido: Yes, this seems to have been also recently updated upstream. Will do.

packages/cli/templates/default-backend-plugin/src/service/router.ts.hbs
tumido: There is no `plugin.ts.hbs` in `default-backend-plugin`. Do you suggest we create it? What would be the benefit?

packages/cli/templates/default-plugin/package.json.hbs
tumido: There seems to be 2 changes to include/update since this PR was created:

https://github.com/backstage/backstage/commit/35725e2b83c47a2590297d4a41e5c2f585dd1630
https://github.com/backstage/backstage/commit/f5b09c3a496bdcf052b0f2dd2fbc985fef0510e3

I'll cherry-pick them. 🙂

davidfestal (approved):

tumido:
/override "SonarCloud Code Analysis"

The code duplication is expected. I've removed the folder from SonarCloud duplication analysis, though there's no straightforward way to retrigger a SonarCloud run without a code push.

PR #786, opened 2023-09-22
chore(ci): run setup-node only if needed

tumido:
`setup-node` has a post hook that tries to save cache after each run. During runs where we actually do not build the package, this fails since no `node_packages` are populated. This results in run failures and a lot of noise in notifications.

Moving this step inside the conditional build phase should affect the post hook and execute it only if the we actually build the asset.

cc @davidfestal 

https://github.com/janus-idp/backstage-plugins/actions/workflows/publish-backend-plugin-manager.yaml

image
image
image

### DIFF
--- a/.github/workflows/publish-backend-plugin-manager.yaml
+++ b/.github/workflows/publish-backend-plugin-manager.yaml
@@ -46,12 +46,6 @@ jobs:
           fetch-depth: 2147483647
           fetch-tags: false
 
-      - name: Setup Node.js
-        uses: actions/setup-node@v3
-        with:
-          node-version: 18
-          cache: 'yarn'
-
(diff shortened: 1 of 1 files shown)


davidfestal (changes requested):
> Moving this step inside the conditional build phase should affect the post hook and execute it only if the we actually build the asset.
> 
> cc @davidfestal
> 

[This step](https://github.com/janus-idp/backstage-plugins/pull/786/files#diff-fbb89529b2798c1bb12d0b7e2edb07b554bac446f5765009fe69b4eb35e0216aR62) checks whether the current commit has already been published to NPM and thus uses the `npm` CLI. Will it succeed if NodeJS is still not setup ?

PR #800, opened 2023-09-25
chore(ci): fix CI error in setup-node

davidfestal:
Fix CI error due to the following bug:

https://github.com/actions/setup-node/issues/772


gashcrumb (approved):
Looking at the linked issues, this does seem like a simpler workaround.

Issue #805, opened 2023-09-26
Group Member count is wrong, if your user count exceeds 100


oliverp:
### Describe the bug

Given, I have 120 users defined in a group “company”. Using the backstage-plugin-keycloak-backend plugin, my group “company” will have a maximum number of 100 users, although there are more users members of that group and even if I set the parameter ```groupQuerySize``` and ```userQuerySize``` to 200.

### Expected Behavior
My group “company” must have 120 users if I set the userQuerySize = 200 or -1. 
Therefore, the plugin must create a group with a memberOf count that is influencable by ```userQuerySize```.

### What are the steps to reproduce this bug?

1. Connect a keycloak with a data situation with a group having more than 100 members
2. Let the plugin sync the data from a keycloak to backstage

### Versions of software used and environment
Current 1.5.4

PR #806, opened 2023-09-26
fix(keycloak): respect fetch size for members

oliverp:
## Referenced issue:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/805

## Short Conclusions about Possible Implementations
We need to pass a value for the max parameter, already present in the used keycloak api. The fetch size must be propagated, otherwise you will end up fetching only a maximum of 100 users as members of a group, causing wrong data to be used.

There are two ways:
- define a new parameter for member fetch size and pass it to the API call
- re-use the userQuerySize since a member is a user anyway

## Current Implementation
There is no possibility to fetch more than 100. It is uses the default in the keycloak server.

## Changes
We pass the userQuerySize into the fetch of members call per group to retrieve the correct data.

### DIFF
--- a/.github/workflows/publish-backend-plugin-manager.yaml
+++ b/.github/workflows/publish-backend-plugin-manager.yaml
@@ -46,14 +46,6 @@ jobs:
           fetch-depth: 2147483647
           fetch-tags: false
 
-      - name: Setup Node.js
-        uses: actions/setup-node@v3
-        env:
-          FORCE_COLOR: 0
-        with:
-          node-version: 18
-          cache: 'yarn'
-
       - name: Get sha of the latest backend-plugin-manager commit
         id: get_commit_sha
         run: |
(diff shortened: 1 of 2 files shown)


oliverp:
The failing ```turbo run test``` works fine locally. Is there something missing, why the pipeline fails? Can i improve the patch?

oliverp:
I fixed the wrong access. Now it should work as expected.

invincibleJai (approved):
LGTM

Thanks @oliverp for your contributions.

oliverp:
When will the patch be shipped?

invincibleJai:
@oliverp can you update commit and title to match plugin name as it's keycloak here

```
fix(keycloak): respect fetch size for members
```

invincibleJai:
> @oliverp can you update commit and title to match plugin name as it's keycloak here
> 
> ```
> fix(keycloak): respect fetch size for members
> ```

can refer to https://github.com/janus-idp/backstage-plugins#requirements for more details. As you would like it to be published soon so will recommend keeping one commit with semantic convention and updating PR title.

PR #808, opened 2023-09-27
feat(rbac): implement REST method to list all plugin permission policies

AndrienkoAleksandr:
### Referenced issue:
Fixes janus-idp/backstage-showcase#431

### Hot to test it

```
export token=$token

# List plugin policies
curl http://localhost:7007/api/permission/plugins/policies  -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v

# List plugin condition rules:
curl http://localhost:7007/api/permission/plugins/condition-rules  -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
```


plugins/rbac-backend/src/service/backend-plugin-ids-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import * as lockfile from '@yarnpkg/lockfile';
+
+import fs from 'fs';
+
+// BackendPluginIDsProvider should retrieve list backend plugin from
+// the root yarn.lock.
+export class BackendPluginIDsProvider {
+  private readonly orgPrefixes = ['@backstage/', '@janus-idp/'];
+  private backendSuffixRegExp = new RegExp('-backend@.*$');
```
tumido, line 11: I don't see how this can return a complete list of backend plugins within the whole instance. This considers only `@backstage/` or `@janus-idp/` packages and assumes the package name to be always suffixed with `-backend`.

By just a quick skim through the https://github.com/janus-idp/backstage-showcase/blob/main/packages/backend/package.json I can tell this will skip packages like `@immobiliarelabs/backstage-plugin-gitlab-backend`.

And that's just a simple example. Since this RBAC plugin is meant to work with any other plugins out there, I think you need a better mechanism for plugin identification.

In addition to that, the whole premise of parting `yarn.lock` is wrong, since:

1. Users are not required to use yarn (though it's very common)
2. More importantly this will skip/omit plugins installed via dynamic plugins.

The only way to determine if a package is used as a backend plugin is via its plugin instance registration.

@davidfestal do you know of any good solution here?

PatAKnight: Right now, we are considering just an approach for finding the static plugins with the plan to address dynamic plugins at a later point. 

Would parsing through the `package.json` be a better approach potentially? We may be over looking it, but we have not found a mechanism where backstage keeps a list of plugins that are registered.

tumido: I'm not sure if a hacky solution that covers 80% cases is the way... What if we try a simplistic approach first?

```yaml
# app-config.yaml
rbac:
  features:
   - catalog
   - whatever
   ...
```

or as JS Array in the constructor? That should be sufficient for the static plugins that require manual TS configuration anyways... For dynamic plugins, we should have an API later on.

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -47,6 +55,9 @@ export class PolicesServer {
     private readonly permissions: PermissionEvaluator,
     private readonly options: RouterOptions,
     private readonly enforcer: Enforcer,
+    private readonly config: Config,
+    private readonly logger: Logger,
+    private readonly discovery: PluginEndpointDiscovery,
```
davidfestal, line 60: Why don't you add a `pluginIDRetriever` parameter in this constructor, which woukd be a function type returning a list of strings (or an interface if you prefer).
This argument would be passed to the PolicyBuilder by the backend app, and from the policy builder down to here.

That will allow you decoupling the retrieval of IDs from the rest of the logic.
You could even provide a default implementation of this argument (with a  default value).
And for the showcase, we would build this list by using the fixed list of static plugins + the BackendPluginsManager.

AndrienkoAleksandr: @davidfestal Thanks, done.

plugins/rbac-backend/src/service/backend-plugin-ids-provider.ts
AndrienkoAleksandr: I provided application configuration and constructor parameter also. I think it will  be enough for static plugins. Expecially, because rbac-backend plugin should be provided only for showcase.

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -47,6 +55,10 @@ export class PolicesServer {
...
+    private readonly pluginIdProvider: { getPluginIds: () => string[] },
```
davidfestal, line 61: If you don't name the type here (through a named interface), why not using simply:
``` typescript
private readonly pluginIdProvider: () => string[],
```
?

plugins/rbac-backend/src/service/policy-builder.ts
```diff
@@ -65,11 +66,27 @@ export class PolicyBuilder {
       policy: await RBACPermissionPolicy.build(env.logger, env.config, enf),
     };
 
+    const pluginIds = env.config.getOptionalStringArray(
+      'permission.rbac.pluginsWithPermission',
+    );
+    if (pluginIds) {
+      // use plugin ids from application configuration.
+      env.pluginIdProvider = {
```
davidfestal, line 74: Do you override the IDs returned by the  `pluginIdProvider` passed in argument, but the pluginIds provided in the config ?
Would it make sense to merge them, so that the config allows adding some, but not removing the ones passed in the code ?

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -298,6 +310,51 @@ export class PolicesServer {
...
+      const endpoints = await pec.get();
+      let perms: Permission[] = [];
+      for (const endpoint of endpoints) {
+        const wellKnownURL = `${endpoint}/.well-known/backstage/permissions/metadata`;
+        try {
+          const permResp = await urlReader.readUrl(wellKnownURL);
+          const permMetaDataRaw = (await permResp.buffer()).toString();
+          const permMetaData = JSON.parse(permMetaDataRaw);
+          if (permMetaData) {
+            perms = [...perms, ...permMetaData.permissions];
```
PatAKnight, line 346: Do we want to expand this to also return the conditional rules along with the permissions?

plugins/rbac-backend/src/service/policy-builder.ts
AndrienkoAleksandr: Ok. Done.

plugins/rbac-backend/src/service/policies-rest-api.ts
AndrienkoAleksandr: I would like interface more, than anonymous function.

plugins/rbac-backend/src/service/policy-builder.ts
```diff
@@ -71,11 +78,25 @@ export class PolicyBuilder {
...
     };
 
+    const pluginIdsConfig = env.config.getOptionalStringArray(
+      'permission.rbac.pluginsWithPermission',
+    );
+    if (pluginIdsConfig) {
+      const pluginIds = pluginIdProvider.getPluginIds();
+      pluginIdProvider.getPluginIds = () => {
+        return [...pluginIdsConfig, ...pluginIds];
+      };
```
PatAKnight, line 88: This would just combine the two. Would it be better to filter to ensure that a plugin isn't in both lists. I.e. a user defines the catalog plugin in `app-config`.

davidfestal (approved):

gorkem (approved):

PR #812, opened 2023-09-28
chore(ci): correctly manage upstream tag case

davidfestal:
Correctly manage upstream tag case when publishing the `backend-plugin-manager` from a `workflow-dispatch` event with a tag git_ref. In this case the commit taken in account should be the current commit of the release.

Signed-off-by: David Festal <dfestal@redhat.com>

### DIFF
--- a/.github/workflows/publish-backend-plugin-manager.yaml
+++ b/.github/workflows/publish-backend-plugin-manager.yaml
@@ -71,11 +71,8 @@ jobs:
 
       - name: Get sha of the latest backend-plugin-manager commit
         id: get_commit_sha
-        env:
-          GIT_REF: ${{ steps.get_checkout_ref.outputs.GIT_REF }}
         run: |
-          echo "GIT_REF=$GIT_REF"
-          if [[ "$GIT_REF" == "master" ]]; then
+          if [[ ${{ steps.get_checkout_ref.outputs.GIT_REF }} == "master" ]]; then
             echo "COMMIT_ID=$(git rev-list --max-count=1 HEAD -- packages/backend-plugin-manager)" >> "$GITHUB_OUTPUT"
           else
             echo "COMMIT_ID=$(git rev-list --max-count=1 HEAD)" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/publish-backend-plugin-manager.yaml
+++ b/.github/workflows/publish-backend-plugin-manager.yaml
@@ -50,21 +50,36 @@ jobs:
           else
             GIT_REF=master
           fi
-          echo "GIT_REF=$GIT_REF" >> $GITHUB_ENV
+          echo "GIT_REF=$GIT_REF" >> $GITHUB_OUTPUT
 
-      - name: Checkout
+      - name: Checkout from master
+        if: steps.get_checkout_ref.outputs.GIT_REF == 'master'
         uses: actions/checkout@v4
         with:
           repository: backstage/backstage
-          ref: ${{ steps.get_checkout_ref.outputs.GIT_REF }}
+          ref: master
           # fetch all history, but only for the current ref
           fetch-depth: 2147483647
           fetch-tags: false
 
+      - name: Checkout from a git ref
+        if: steps.get_checkout_ref.outputs.GIT_REF != 'master'
+        uses: actions/checkout@v4
+        with:
+          repository: backstage/backstage
+          ref: '${{ steps.get_checkout_ref.outputs.GIT_REF }}'
+
       - name: Get sha of the latest backend-plugin-manager commit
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-backend-plugin-manager.yaml
```diff
@@ -50,21 +50,36 @@ jobs:
...
         uses: actions/checkout@v4
         with:
           repository: backstage/backstage
-          ref: ${{ steps.get_checkout_ref.outputs.GIT_REF }}
+          ref: master
           # fetch all history, but only for the current ref
           fetch-depth: 2147483647
           fetch-tags: false
 
+      - name: Checkout from a git ref
```
nickboldt, line 65: instead of an "if master checkout from master, if other checkout from other" ... why not just ALWAYS checkout from `steps.get_checkout_ref.outputs.GIT_REF` ? Seems like on line 53 you set a value to either the `master` branch or the input branch.

.github/workflows/publish-backend-plugin-manager.yaml
```diff
@@ -50,21 +50,36 @@ jobs:
...
+          ref: '${{ steps.get_checkout_ref.outputs.GIT_REF }}'
+
       - name: Get sha of the latest backend-plugin-manager commit
         id: get_commit_sha
+        env:
+          GIT_REF: ${{ steps.get_checkout_ref.outputs.GIT_REF }}
         run: |
-          echo "COMMIT_ID=$(git rev-list --max-count=1 HEAD -- packages/backend-plugin-manager)" >> "$GITHUB_OUTPUT"
+          echo "GIT_REF=$GIT_REF"
+          if [[ "$GIT_REF" == "master" ]]; then
```
nickboldt, line 78: you could simply do 
```
if [[ ${{ steps.get_checkout_ref.outputs.GIT_REF }} == "master" ]]; then
```

nickboldt (approved):

.github/workflows/publish-backend-plugin-manager.yaml
davidfestal: In the case of master, we also want to specify special values for `fetch-depth` and `fetch-tags`, which do not make sense for the git_ref case.

.github/workflows/publish-backend-plugin-manager.yaml
davidfestal: fixed in https://github.com/janus-idp/backstage-plugins/pull/812/commits/fb940dbceedd5e13e2fc2e57e82732d4def4a6c5

davidfestal:
@nickboldt I fixed one of you comments and answered the second one.

.github/workflows/publish-backend-plugin-manager.yaml
```diff
@@ -50,21 +50,33 @@ jobs:
...
-      - name: Checkout
+      - name: Checkout from master
+        if: steps.get_checkout_ref.outputs.GIT_REF == 'master'
         uses: actions/checkout@v4
         with:
           repository: backstage/backstage
-          ref: ${{ steps.get_checkout_ref.outputs.GIT_REF }}
+          ref: master
           # fetch all history, but only for the current ref
           fetch-depth: 2147483647
```
nickboldt, line 62: Seems weird to need the full history. why not just the last 20 commits ?

nickboldt:
Tests/builds are failing:

```
error Couldn't find any versions for "@janus-idp/backstage-plugin-rbac-backend" that matches "^0.1.0"
```
-- https://github.com/janus-idp/backstage-plugins/actions/runs/6379904117/job/17313285517?pr=812
-- https://github.com/janus-idp/backstage-plugins/actions/runs/6379904117/job/17313285240?pr=812

.github/workflows/publish-backend-plugin-manager.yaml
davidfestal: Because we want to be able to search for the last commit that impacted the package `backend-plugin-manager`. I just didn't want to use an arbitrary limit that might return a wrong commit.
OTOH we only checkout one branch, so that's not that long.

davidfestal:
> Tests/builds are failing:
> 
> ```
> error Couldn't find any versions for "@janus-idp/backstage-plugin-rbac-backend" that matches "^0.1.0"
> ```
> 
> -- https://github.com/janus-idp/backstage-plugins/actions/runs/6379904117/job/17313285517?pr=812 -- https://github.com/janus-idp/backstage-plugins/actions/runs/6379904117/job/17313285240?pr=812

Yeah, that doesn't seem to be related to this change at all.

nickboldt:
tests might pass after this https://github.com/janus-idp/backstage-plugins/pull/818

PR #819, opened 2023-10-02
feat: add the `export-dynamic-plugin` command, with an implementation for the backend case

davidfestal:
Add the `export-dynamic-plugin` command in the `janus-idp` CLI for the backend case.


packages/cli/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@janus-idp/cli",
+  "name": "@janus-idp/backstage-cli",
```
tumido, line 2: Why do you rename the package?

packages/cli/package.json
```diff
@@ -1,7 +1,7 @@
...
   "description": "CLI for developing Backstage plugins and apps",
-  "version": "1.0.5",
+  "version": "1.0.6-next.0",
```
tumido, line 4: This is versioned automatically by the release toolkit and released on merge

packages/cli/package.json
```diff
@@ -161,7 +161,7 @@
     "@types/minimatch": "5.1.2",
     "@types/mock-fs": "4.13.1",
     "@types/node": "18.17.17",
-    "@types/npm-packlist": "7.0.0",
+    "@types/npm-packlist": "3.0.0",
```
tumido, line 164: Any reason for such a severe downgrade?

packages/cli/src/commands/index.ts
```diff
@@ -152,6 +152,34 @@ export function registerScriptCommand(program: Command) {
...
+    )
+    .option(
+      '--clean',
+      'Remove the dynamic plugin output before exporting again.',
+    )
+    .option(
+      '--dev',
+      'Allow testing/debugging a backend plugin dynamic loading locally. This installs the dynamic plugin content (symlink) into the dynamic plugins root folder configured in the app config. This also creates a link from the dynamic plugin content to the plugin package `src` folder, to enable the use of source maps.',
+    )
+    .action(lazy(() => import('./export-dynamic-plugin').then(m => m.command)));
```
tumido, line 181: I feel these options are relevant to backend and can mostly never be used on the frontend plugins. I wonder if we can make it obvious in any way.

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+        if (typeof expr === 'string') {
+          return str === expr;
+        }
+        return expr.test(str);
+      }
+
+      pkgToCustomize.name = `${pkgToCustomize.name}-dynamic`;
+      (pkgToCustomize as any).bundleDependencies = true;
+      pkgToCustomize.scripts = undefined;
+      pkgToCustomize.types = undefined;
```
tumido, line 186: Out of curiosity, is there a reason to explicitly remove these?

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+    include: mergeWithOutput,
+    exclude: mergeWithOutput.length !== 0 ? undefined : /.*/,
+  };
+  const moveToPeerDependencies: (string | RegExp)[] = [/@backstage\//];
+
+  if (opts.sharedPackage !== undefined) {
+    for (const sharedPkg of opts.sharedPackage as string[]) {
+      moveToPeerDependencies.push(sharedPkg);
+    }
+  }
```
tumido, line 110: Can't this be simply:

```suggestion
  const moveToPeerDependencies: (string | RegExp)[] = [/@backstage\//, ...(opts.sharedPackage || [])]
```

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+        f => !f.startsWith('dist-dynamic/'),
+      );
+
+      for (const dep in dependenciesToAdd) {
+        if (!Object.prototype.hasOwnProperty.call(dependenciesToAdd, dep)) {
+          continue;
+        }
+        if (pkgToCustomize.dependencies === undefined) {
+          pkgToCustomize.dependencies = {};
+        }
```
tumido, line 198: Why is this always reset to empty object? For each item in `dependenciesToAdd` this potentially resets the dependencies. Shouldn't this be outside the for loop?

Also, can't you use:
```suggestion
        pkgToCustomize.dependencies ||= {};
```

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+        if (existingVersion !== dependenciesToAdd[dep]) {
+          if (!semver.satisfies(existingVersion, dependenciesToAdd[dep])) {
+            throw new Error(
+              `The version of a dependency ('${dep}') of an embedded module conflicts with main module dependencies: '${dependenciesToAdd[dep]}', '${existingVersion}'`,
+            );
+          }
+        }
+      }
+      if (pkgToCustomize.dependencies) {
```
tumido, line 212: This is always true since the reset above assigns a `{}` here and `Boolean({}) === true`

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+            continue;
+          }
+
+          for (const toMove of moveToPeerDependencies) {
+            if (test(dep, toMove)) {
+              console.log(`Moving '${dep}' to peerDependencies`);
+
+              if (pkgToCustomize.peerDependencies === undefined) {
+                pkgToCustomize.peerDependencies = {};
+              }
```
tumido, line 248: Same as above

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+              pkgToCustomize.peerDependencies[dep] =
+                pkgToCustomize.dependencies[dep];
+              delete pkgToCustomize.dependencies[dep];
+              break;
+            }
+          }
+        }
+      }
+      pkgToCustomize.devDependencies = {};
```
tumido, line 257: Why do we reset this? `devDependencies` are not bundled anyways. If you're trying to avoid `devDependencies` during `yarn install`, simply run `yarn install --production`

https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production-true-false

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,320 @@
...
+    // Create an empty yarn.lock to make it clear that the exported dynamic plugin is
+    // NOT part of a workspace.
+    await fs.ensureFile(yarnLock);
+  }
+
+  if (opts.install) {
+    let yarnInstall = 'yarn install';
+    if (yarnLockExists) {
+      yarnInstall += ' --frozen-lockfile';
+    }
```
tumido, line 274:
```suggestion
    const yarnInstall = `yarn install${yarnLockExists ? " --frozen-lockfile" : ""}`
```

packages/cli/package.json
davidfestal: Yes, there were some compilation problems, with some methods having changed signatures.

packages/cli/package.json
davidfestal: OK. So what should I do here ? Just rollback this change ?

packages/cli/package.json
davidfestal: Just a mistake (at some point, to test I had renamed it to `dfatwork-pkgs/backstage-cli` to push it to my personal NPMJS org). Let me change this back

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: Fixed

packages/cli/src/commands/index.ts
davidfestal: I added the ` (backend plugin only)` mention at the end of the description for most of them.
Only `--shared-package` and `--clean` don't have this mention, since I assume they could be meaningful for frontend plugins as well:
- `--clean` can effectively completely remove the `dist-scalprum` folder before doing the export. 
- `--shared-package` may be useful to allow mentioning some packages that you want to be shared in addition to those you already consider so in the generated module federation settings you use)

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: This package has no sources, and no scripts in it. It is already derived from the source package. I don't see how it makes sense to add `devDependencies`.

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: This package has no sources, and no action that is to be done on sources (such as lint, build, etc ...). It is already derived from the source package. I don't see how it makes sense to keep scripts or even types in the corresponding package.json.

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: It is reset to empty object only if it is undefined. So it will be only once afaict.
I do it in the loop so that if `dependenciesToAdd` is empty, we don't change `pkgToCustomize.dependencies` at all, because we won't need to add any new dependency in this (possibly undefined) list.

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: It is not (uness I'm misreading). It might be undefined if the `dependenciesToAdd` list was empty.

packages/cli/package.json
davidfestal: Fixed

packages/cli/package.json
davidfestal: I rolled back the change

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: > Also, can't you use:

Fixed. Thanks for the suggestion.

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: I added back `devDependencies` in the derived package, as discussed.

packages/cli/src/commands/export-dynamic-plugin/index.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2022 The Backstage Authors
```
nickboldt, line 2: new file, old date.

packages/cli/src/lib/builder/embedPlugin.ts
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2020 The Backstage Authors
```
nickboldt, line 2: dated from before the pandemic?

packages/cli/src/commands/export-dynamic-plugin/command.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (role === 'backend-plugin' || role === 'backend-plugin-module') {
+    return backend(roleInfo, opts);
+  }
+
+  if (role === 'frontend-plugin') {
+    return frontend(roleInfo, opts);
+  }
+
+  throw new Error(
+    'Only packages with the "backend-plugin", "backend-plugin-module" or "frontend-plugin" roles can be exported as dynamic backend plugins',
```
nickboldt, line 44: how are  "backend-plugin" and  "backend-plugin-module"  different?

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,373 @@
...
+  const commonPackage = pkg.name.replace(/-backend$/, '-common');
+  if (commonPackage !== pkg.name) {
+    mergeWithOutput.push(commonPackage);
+  }
+
+  if (opts.embedPackage !== undefined) {
+    for (const pkgToEmbed of opts.embedPackage as string[]) {
+      if (pkg.dependencies === undefined || !(pkgToEmbed in pkg.dependencies)) {
+        console.log(
+          `Embeded package '${pkgToEmbed}' is not part of direct dependencies.Are you sure you want to embed it ?`,
```
nickboldt, line 89:
```suggestion
          `Embeded package '${pkgToEmbed}' is not part of direct dependencies. Are you sure you want to embed it ?`,
```

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,373 @@
...
+        ) {
+          dependenciesToAdd[dependencyName] = newDependencyVersion;
+          console.log(
+            `Several compatible versions ('${existingDependencyVersion}', '${newDependencyVersion}') of the same transitive dependency ('${dependencyName}') for embedded module ('${embeddedModule}'): keeping '${newDependencyVersion}'`,
+          );
+          return;
+        }
+
+        throw new Error(
+          `several incompatible versions ('${existingDependencyVersion}', '${newDependencyVersion}') of the same transitive dependency ('${dependencyName}') for embedded module ('${embeddedModule}')`,
```
nickboldt, line 175:
```suggestion
          `Several incompatible versions ('${existingDependencyVersion}', '${newDependencyVersion}') of the same transitive dependency ('${dependencyName}') for embedded module ('${embeddedModule}')`,
```

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,373 @@
...
+        }
+        if (existingVersion !== dependenciesToAdd[dep]) {
+          const existingMinVersion = semver.minVersion(existingVersion);
+
+          if (
+            existingMinVersion &&
+            semver.satisfies(existingMinVersion, dependenciesToAdd[dep])
+          ) {
+            console.log(
+              `The version of a dependency ('${dep}') of an embedded module differs with main module dependencies: '${dependenciesToAdd[dep]}', '${existingVersion}': keeping it though since it is compatible`,
```
nickboldt, line 244:
```suggestion
              `The version of a dependency ('${dep}') of an embedded module differs from the main module's dependencies: '${dependenciesToAdd[dep]}', '${existingVersion}': keeping it as it is compatible`,
```

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,373 @@
...
+            );
+          } else {
+            throw new Error(
+              `The version of a dependency ('${dep}') of an embedded module conflicts with main module dependencies: '${dependenciesToAdd[dep]}', '${existingVersion}'`,
```
nickboldt, line 248:
```suggestion
              `The version of a dependency ('${dep}') of an embedded module conflicts with main module dependencies: '${dependenciesToAdd[dep]}', '${existingVersion}': cannot proceed!`,
```

davidfestal: fixed

packages/cli/src/commands/export-dynamic-plugin/command.ts
davidfestal: The second one is a plugin that contributes extensions to another plugin.

packages/cli/src/commands/export-dynamic-plugin/index.ts
davidfestal: fixed

packages/cli/src/lib/builder/embedPlugin.ts
davidfestal: fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,373 @@
...
+import { paths } from '../../lib/paths';
+import { Task } from '../../lib/tasks';
+
+export async function backend(
+  roleInfo: PackageRoleInfo,
+  opts: OptionValues,
+): Promise<void> {
+  if (!fs.existsSync(paths.resolveTarget('src', 'dynamic'))) {
+    throw new Error(
+      `Package doesn't seem to support dynamic loading. It should have a ./src/dynamic folder, containing the dynamic loading entrypoints.`,
```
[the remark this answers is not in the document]
davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/backend.ts
davidfestal: In fact I removed it back again since, at least in yarn v1 `yarn --production` still brings the `devDependencies`, at least the first time it is run, and there's no way to remove the dev dependencies from the generated `yarn.lock`.

See https://github.com/yarnpkg/yarn/issues/6373#issuecomment-760068356

davidfestal:
@tumido and @nickboldt I fixed / answered all your comments. Please resolve any pending conversation if the answer is OK for you.

nickboldt (approved):
already have a plan for downstreaming this (adding more paths to cachito configuration to cache more yarn deps), so +1 to merge when you're ready.

davidfestal:
@tumido Do you think you could resolve the remaining unresolved conversations if it's OK for you ?

tumido (approved):
/lgtm
/approve

packages/cli/src/commands/export-dynamic-plugin/backend.ts
```diff
@@ -0,0 +1,383 @@
...
+  // Remove the `dist` folder of the original plugin root folder and rebuild it,
+  // since it has been compiled with dynamic-specific settings.
+  await fs.remove(paths.resolveTarget('dist'));
+  await buildPackage({
+    outputs,
+    minify: Boolean(opts.minify),
+  });
+
+  if (opts.dev) {
+    if (opts.dev) {
```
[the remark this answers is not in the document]
davidfestal: Oh, no it seems to be a typo.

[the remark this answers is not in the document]
davidfestal: Good catch !

PR #815, opened 2023-09-29
chore(tekton): update tekton to be compatible with backstage 1.18.x

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/776

Updated tekton plugin dependencies to make it compatible with 1.18.4 (latest version)

PR #832, opened 2023-10-11
chore(topology): update topology to be compatible with backstage 1.18.x

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/830

Updated topology plugin dependencies to make it compatible with 1.18.4 (latest version)


divyanshiGupta:
cc: @debsmita1 @invincibleJai

invincibleJai (approved):
LGTM

divyanshiGupta:
@tumido topology npm package hasnt been updated with this PR yet, same with this one https://github.com/janus-idp/backstage-plugins/pull/815 for tekton plugin. Am I missing something that triggers the release process or is this something else?

tumido:
It should be released in `1.15.7`, isn't it?

https://github.com/janus-idp/backstage-plugins/commits/%40janus-idp/backstage-plugin-topology%401.15.7/

I think the issues is that `chore` doesn't trigger a release... So it was caused by a mislabeled PR.

divyanshiGupta:
> It should be released in `1.15.7`, isn't it?
> 
> 
> 
> https://github.com/janus-idp/backstage-plugins/commits/%40janus-idp/backstage-plugin-topology%401.15.7/
> 
> 
> 
> I think the issue is that `chore` doesn't trigger a release... So it was caused by a mislabeled PR.

Oh, got it. Yes the new version should be 1.15.7. I see that the new versions for both plugins got released as one of the fixes PR got merged so we should be good now. Will make sure not to use chore in case of immediate releases requirement. Thanks.

Issue #804, opened 2023-09-26
[Tekton plugin]: Update user experience for tekton plugin


invincibleJai:
**Description**
As a user, I want to visualize the list of all pipelineRuns and task run status at a glance in the CI tab

<img width="1161" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/5654cfc1-e62d-4537-939f-a122caea3265">

**Acceptance Criteria**

- [x] Have appropriate filter tools for Cluster, status
- [x] Have search for table
- [x] On click of each row, it should expand and show the PLR visualization
- [x] On click on status it should show PLR logs modal as it does today
- [x] Table should have pagination
- [x] expand and collapse icon on the header to expand/collapse all visible rows at once

invincibleJai:
@ShiranHi / @divyanshiGupta / @debsmita1 PTAL and feel free to add/modify if anything is not accurate

ShiranHi:
Prototype: 
https://www.figma.com/proto/Gp6QQ0wtRO3RSCphRZIdjQ/RHDH-Theme-for-Backstage?type=design&node-id=295-127&t=wxAQOpamCxu8Ad6V-1&scaling=min-zoom&page-id=162%3A592&starting-point-node-id=230%3A474&mode=design

Figma file: 
https://www.figma.com/file/Gp6QQ0wtRO3RSCphRZIdjQ/RHDH-Theme-for-Backstage?type=design&node-id=295%3A127&mode=design&t=NkIYl4IXIw4c7Usa-1

PR #839, opened 2023-10-17
feat(tekton)!: update tekton UX

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/804

This PR updates tekton plugin UX based on new designs. It also removes support for a separate tekton tab.

Screenshots:

https://github.com/janus-idp/backstage-plugins/assets/20724543/dd157912-f7d8-46fd-8b65-5d638dec6db8

To-Do:

- [x] Update Readme
- [x] Add more unit tests


divyanshiGupta:
cc: @ShiranHi

ShiranHi:
@divyanshiGupta that looks great!
I have 3 questions:
1. Is it possible to decrease the padding between the arrow used to collapse and expand rows in the table and the first column NAME? The current gap looks quite large.
2. Could we consider including tooltips for when you hover over the "Collapse all" and "Expand all" icons?
2. How about adding an icon for each status in the filter? Let's go with the same icons we use on the table

divyanshiGupta:
> @divyanshiGupta that looks great! I have 3 questions:
> 
> 1. Is it possible to decrease the padding between the arrow used to collapse and expand rows in the table and the first column NAME? The current gap looks quite large.
> 2. Could we consider including tooltips for when you hover over the "Collapse all" and "Expand all" icons?
> 3. How about adding an icon for each status in the filter? Let's go with the same icons we use on the table

@ShiranHi thanks for the feedback. Will try to add the changes that you have suggested.

plugins/tekton/src/components/common/TableExpandCollapse.tsx
```diff
@@ -0,0 +1,41 @@
...
+});
+
+export const TableExpandCollapse = () => {
+  const classes = useStyles();
+  const { isExpanded, setIsExpanded } = React.useContext(
+    TektonResourcesContext,
+  );
+  // const [collapseDisabled, setCollapseDisabled] =
+  //   React.useState<boolean>(false);
+  // const [expandDisabled, setExpandDisabled] = React.useState<boolean>(false);
```
debsmita1, line 23: remove these comments

debsmita1 (changes requested):

debsmita1:
The PipelineRunVisualization component doesn't need its own TektonResourcesContext Provider.

debsmita1:
In the dark theme mode, the list looks like this
Screenshot 2023-10-19 at 3 39 04 PM

divyanshiGupta:
@ShiranHi I have reduced padding for expand/collapse buttons and added tooltips as well. Also header table cells have same padding for all but the gap between name and status cell seem more because the plr name occupies more space and hence has more width so the name cell aligns itself with it to look consistent hence increasing the space between the cells.


https://github.com/janus-idp/backstage-plugins/assets/20724543/ad7b4c48-2b3e-4cea-ad8d-8e5579dbe66e


@debsmita1 I have fixed the dark theme issue.

divyanshiGupta:
@ShiranHi currently the MUI core `Select` component currently doesnt support anything apart from strings. There is another component  `Automcomplete` https://v4.mui.com/components/autocomplete/  from MUI lab which supports adding icons but since this component is under lab it might be experimental and is not used anywhere else as of now in the code base. cc: @invincibleJai

ShiranHi:
@divyanshiGupta got it. And thank you for adding the tooltips, looks good to me.

plugins/tekton/src/components/PipelineRunList/PipelineRunListSearchBar.tsx
```diff
@@ -0,0 +1,83 @@
...
+  IconButton,
+  Input,
+  InputAdornment,
+  makeStyles,
+} from '@material-ui/core';
+import Clear from '@material-ui/icons/Clear';
+import Search from '@material-ui/icons/Search';
+
+type PipelineRunListSearchBarProps = {
+  pipelineRuns?: any[];
```
invincibleJai, line 15: nit: can we have type for it?

```suggestion
  pipelineRuns?: PipelineRunKind[];
```

invincibleJai (commented):
Thanks @divyanshiGupta, this looks good. Have added a few nits.

debsmita1 (changes requested):

invincibleJai:
@divyanshiGupta I see one edge case with multi-cluster and two warnings 


https://github.com/janus-idp/backstage-plugins/assets/5129024/c1a59ec5-51e0-48f5-ba53-fa7be2051a2c


image

image

divyanshiGupta:
> @divyanshiGupta I see one edge case with multi-cluster and two warnings
> 
>  tektonUX2.mov 
> image image

This issue is not reproducible for me. 

https://github.com/janus-idp/backstage-plugins/assets/20724543/5ad87ef1-9509-4d53-b316-33c3393942f7

Also I dont see the first warning. The warning regarding the key prop is from main.

invincibleJai:
@divyanshiGupta Okay, let's get this and we observe this issue we can take it up an an issue.

PR #857, opened 2023-10-18
chore: remove tracked but ignored file

gashcrumb:
Since app-config.local.yaml is checked in it isn't ignored even though the .gitignore file is set up to ignore it.  This change rectifies this problem, ensuring that someone doesn't accidentally commit some local configuration to this file.


tumido (approved):
/lgtm
/approve

Issue #864, opened 2023-10-19
(RBAC) CRUD Backend APIs for Roles


PatAKnight:
### What do you want to improve?

The RBAC plugin should support the concept of roles and provide APIs for role CRUD. All CRUD operations should be protected by the permissions as part of #510. 

Created from the spike #836 exploration

### What is the current behavior?

RBAC plugin authorizes only via groups and users.

### What will the new behavior be?

Roles can be created and stored on the Backstage database by the RBAC Admins. They will include groups and users and will not include other roles. 

### Acceptance Criteria

- CRUD for roles
- Update the CRUD for policies validation to now include roles
- Update the RoleManager to now handle role and group memberships
- Update documentation to now include roles
- Update tests

PR #867, opened 2023-10-23
feat(rbac): implement the concept of roles in rbac

PatAKnight:
## Description
This PR introduces the concept of Roles to the RBAC plugin. We are now able to create roles using backstage groups and users. Role manager has been updated to reflect roles. CRUD operations have been implemented for roles.

## Fixes
- #864 

Currently as a draft as tests still need to be added for the new functionality.

1. Update app-config.yaml to includer your user as admin
```yaml
permission:
  enabled: true
  rbac:
    admin:
      users:
        - name: user:default/<USERNAME>
```
2. Login using github auth.
3. Retrieve user token. Copy it to terminal: `export token='...token...value`

Read roles
`curl -X GET "http://localhost:7007/api/permission/roles" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v`

Read a single role
`curl -X GET "http://localhost:7007/api/permission/roles/role/default/guests" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v`

Create a single role
`curl -X POST "http://localhost:7007/api/permission/roles" -d '{ "roleMemberReferences":  [ "user:default/test" ], "roleName": "role:default/test" }' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v`

Update a single role
`curl -X PUT "http://localhost:7007/api/permission/roles/role/default/test" -d '{ "oldRole": { "roleMemberReferences":  [ "user:default/test" ], "roleName": "role:default/test" }, "newRole": { "roleMemberReferences": [ "user:default/test", "user:default/test2" ], "roleName": "role:default/test" } }' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v`

Delete a single user / group from a role
`curl -X DELETE "http://localhost:7007/api/permission/roles/role/default/test?roleMemberReferences=user:default/test" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v`
ex. delete `user:default/test` from
```
roleMemeberReferences: [user:default/test, user:default/test2],
roleName: role:default/test
```
resulting in
```
roleMemeberReferences: [user:default/test2],
roleName: role:default/test
```

Delete a single role
`curl -X DELETE "http://localhost:7007/api/permission/roles/role/default/test" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v`

### DIFF
--- a/plugins/rbac-backend/src/service/policies-rest-api.test.ts
+++ b/plugins/rbac-backend/src/service/policies-rest-api.test.ts
@@ -28,7 +28,17 @@ const mockEnforcer: Partial<Enforcer> = {
     .mockImplementation(async (..._param: string[]): Promise<boolean> => {
       return true;
     }),
+  addGroupingPolicies: jest
+    .fn()
+    .mockImplementation(async (..._param: string[]): Promise<boolean> => {
+      return true;
+    }),
   hasPolicy: jest
+    .fn()
+    .mockImplementation(async (..._param: string[]): Promise<boolean> => {
+      return false;
+    }),
+  hasGroupingPolicy: jest
     .fn()
     .mockImplementation(async (..._param: string[]): Promise<boolean> => {
       return false;
@@ -42,6 +52,13 @@ const mockEnforcer: Partial<Enforcer> = {
         return [
           ['user:default/permission_admin', 'policy-entity', 'create', 'allow'],
         ];
+      },
+    ),
+  getFilteredGroupingPolicy: jest
+    .fn()
+    .mockImplementation(
+      async (_fieldIndex: number, ..._fieldValues: string[]) => {
+        return [['user:default/permission_admin', 'role:default/rbac_admin']];
       },
     ),
 };
@@ -92,6 +109,11 @@ describe('REST policies api', () => {
 
   beforeEach(async () => {
     mockEnforcer.hasPolicy = jest
+      .fn()
+      .mockImplementation(async (..._param: string[]): Promise<boolean> => {
+        return false;
+      });
+    mockEnforcer.hasGroupingPolicy = jest
       .fn()
       .mockImplementation(async (..._param: string[]): Promise<boolean> => {
         return false;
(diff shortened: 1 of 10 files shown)


plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -227,84 +234,287 @@ export class PolicesServer {
...
+        if (!isAdded) {
+          throw new Error('Unexpected error');
+        }
+
+        response.status(200).end();
+      },
+    );
+
+    // Role CRUD
+    // TODO: Need to add the authorization piece
```
AndrienkoAleksandr, line 312: I think you can remove comment, cause you handled this part.

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -227,84 +234,287 @@ export class PolicesServer {
...
       );
 
       if (decision.result === AuthorizeResult.DENY) {
         throw new NotAllowedError(); // 403
       }
+      const roleRaw: Role = request.body;
+      const err = validateRole(roleRaw);
+      if (err) {
+        throw new InputError( // 400
+          `Invalid policy definition. Cause: ${err.message}`,
```
AndrienkoAleksandr, line 376:
```suggestion
          `Invalid role definition. Cause: ${err.message}`,
```

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -227,84 +234,287 @@ export class PolicesServer {
...
+      );
+
+      if (decision.result === AuthorizeResult.DENY) {
+        throw new NotAllowedError(); // 403
+      }
+      const roleEntityRef = this.getEntityReference(request);
+
+      const oldRoleRaw = request.body.oldRole;
+      if (!oldRoleRaw) {
         throw new InputError(`'oldPolicy' object must be present`); // 400
```
AndrienkoAleksandr, line 413:
```suggestion
        throw new InputError(`'oldRole' object must be present`); // 400
```

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -227,84 +234,287 @@ export class PolicesServer {
...
       }
-      const newPolicyRaw = req.body.newPolicy;
-      if (!newPolicyRaw) {
+      const newRoleRaw = request.body.newRole;
+      if (!newRoleRaw) {
         throw new InputError(`'newPolicy' object must be present`); // 400
```
AndrienkoAleksandr, line 417:
```suggestion
        throw new InputError(`'newRole' object must be present`); // 400
```

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -270,7 +258,7 @@ describe('BackstageRoleManager', () => {
       expect(result).toBeFalsy();
     });
 
-    // user:default/mike should inherits from group:default/team-a
+    // user:default/mike should inherits from group:default/team-a x
```
AndrienkoAleksandr, line 261: I think this is small typo:
```suggestion
    // user:default/mike should inherits from group:default/team-a
```

plugins/rbac-common/src/types.ts
```diff
@@ -4,10 +4,15 @@ export type Policy = {
   effect?: string;
 };
 
-export type EntityReferencedPolicy = Policy & {
+export type RoleBasedPolicy = Policy & {
   entityReference?: string;
 };
 
+export type Role = {
+  entityReferences: string[];
```
AndrienkoAleksandr, line 12: I think we need to provide more clear name to explain, that here is stored references to role users and groups.

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -71,11 +117,18 @@ class AncestorSearchMemo {
   }
 }
 
+function createRoleName(name: string, domain: string[]) {
+  return `${domain[0]}::${name}`;
```
AndrienkoAleksandr, line 121: Sorry, do we really use domains ?

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -238,4 +342,26 @@ export class BackstageRoleManager implements RoleManager {
       await this.findAncestorGroups(memo);
     }
   }
+
+  private createRole(name: string): Role {
```
AndrienkoAleksandr, line 346: What do you think, maybe `getOrCreateRole` name will be better?

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -238,4 +342,26 @@ export class BackstageRoleManager implements RoleManager {
...
+      return role;
+    }
+    const newRole = new Role(name);
+    this.allRoles.set(name, newRole);
+
+    return newRole;
+  }
+
+  // parse the entity to find out if it is a user / group / or role
+  private parseEntity(name: string): string {
```
AndrienkoAleksandr, line 358: What do you think, maybe parseEntityKind name will be better?

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -127,6 +194,24 @@ export class BackstageRoleManager implements RoleManager {
       return true;
     }
 
+    // Create Temporary roles to compare
+    const tempRoleName = createRoleName(name1, domain);
```
AndrienkoAleksandr, line 198: We have a code upper:
```
    if (domain.length > 0) {
      throw new Error('domain argument is not supported.');
    }
```
So I don't think that we're using domain. Also I'm not sure that we need this feature at least rigth now.

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -142,6 +227,14 @@ export class BackstageRoleManager implements RoleManager {
...
 
+      if (!memo.hasEntityRef(name2) && this.parseEntity(name2) === 'role') {
+        for (const [key, value] of this.allRoles.entries()) {
+          return (
+            value.hasRole(tempRoleName2, 1) &&
+            memo.hasEntityRef(this.parseRole(key))
+          );
+        }
+      }
       memo.hasEntityRef(name2);
```
AndrienkoAleksandr, line 238: Remove this redundant line, please. It do nothing: 
```suggestion
```

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -142,6 +227,14 @@ export class BackstageRoleManager implements RoleManager {
     if (!memo.isAcyclic()) {
       const cycles = memo.findCycles();
 
+      if (!memo.hasEntityRef(name2) && this.parseEntity(name2) === 'role') {
```
AndrienkoAleksandr, line 230: I guess that's a wrong place for this code block. `if (!memo.isAcyclic()) {` that's code block which should notify policy admin about cycle dependency - mistake in the group structure.

When you provide return with result, then you hide warning.

```
if (this.parseEntity(name2) === 'role') {
  for (const [key, value] of this.allRoles.entries()) {
    return (
        value.hasRole(tempRoleName2, 1) &&
         memo.hasEntityRef(this.parseRole(key))
    );
  }  
}
```

plugins/rbac-common/src/types.ts
PatAKnight: What do you think would be best? Maybe something like userReferences or groupReferences?

plugins/rbac-backend/src/service/role-manager.ts
PatAKnight: So would removing the return statement fix the issue? Something like this:
```
if (!memo.hasEntityRef(name2) && this.parseEntityKind(name2) === 'role') {
  for (const [key, value] of this.allRoles.entries()) {
    value.hasRole(tempRoleName2, 1);
    memo.hasEntityRef(key);
  }
}
```

plugins/rbac-common/src/types.ts
AndrienkoAleksandr: Maybe roleMemberReferences, since it can be users and group in the same time.

plugins/rbac-backend/src/service/role-manager.ts
AndrienkoAleksandr: I think in this case code do nothing, so this code can be removed. I guess in case cycle dependency it is better to return always false for has link with warning message in the log. Also you can modify error message to notify user, that role permission was also affected with some group which contain cycle dependency.

plugins/rbac-backend/src/service/policies-rest-api.test.ts
```diff
@@ -892,6 +914,523 @@ describe('REST policies api', () => {
...
+      const result = await request(app)
+        .post('/roles')
+        .send({
+          roleMemberReferences: ['user:default/permission_admin'],
+        });
+
+      expect(result.statusCode).toBe(400);
+      expect(result.body.error).toEqual({
+        name: 'InputError',
+        message: `Invalid role definition. Cause: 'entityReference' must not be empty`,
```
AndrienkoAleksandr, line 1065: Error message is not clear in this case. It is hard to figure out what's wrong from this error message.

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -240,71 +406,122 @@ export class PolicesServer {
...
+      const isAdded = await this.enforcer.addGroupingPolicies(newRole);
       if (!isAdded) {
         throw new Error('Unexpected error');
       }
 
-      resp.status(200).end();
+      response.status(200).end();
     });
 
+    router.delete(
```
AndrienkoAleksandr, line 477: I think delete method should delete role, corresponding permission policies and user, group references. What do you think, @PatAKnight ?

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -123,6 +133,52 @@ describe('BackstageRoleManager', () => {
       expect(result).toBeFalsy();
     });
 
+    // user:default/mike should not inherits from role:default/role
+    //
+    //     Hierarchy:
+    //
+    // user:default/mike -> user without group
```
AndrienkoAleksandr, line 140:
```suggestion
    // user:default/mike -> user without role
```

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -161,6 +237,26 @@ describe('BackstageRoleManager', () => {
       expect(result).toBeFalsy();
     });
 
+    // user:default/mike should not inherits from role:default/somerole
+    //
+    //     Hierarchy:
+    //
+    // group:default/not-matched-group -> role:default/somerole
+    //         |
+    // user:default/mike
```
AndrienkoAleksandr, line 246:
```suggestion
    // group:default/not-matched-group role:default/somerole
    //         |                                                                   |
    // user:default/mike                              group:default/somegroup
```

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -511,6 +929,52 @@ describe('BackstageRoleManager', () => {
...
     });
 
+    // user:default/mike should inherits role from group:default/team-a, but we have cycle dependency: team-a -> team-c.
+    // So return false on call hasLink.
+    //
+    //     Hierarchy:
+    //
+    // group:default/team-a -> role:default/team-a   group:default/team-b
+    //       ↓       ↑                                    ↓
+    // group:default/team-c -> role:default/team-c   group:default/team-d
```
AndrienkoAleksandr, line 939: You didn't use  role:default/team-c, so you can remove it from Hierarchy, or really use it with roleManager.addLink('group:default/team-c', 'role:default/team-c');

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -581,6 +1045,80 @@ describe('BackstageRoleManager', () => {
...
+        'group:default/team-a',
+      );
+      expect(result).toBeFalsy();
+      expect(loggerMock.warn).toHaveBeenCalledWith(
+        'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-c"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for group: group:default/team-a',
+      );
+
+      result = await roleManager.hasLink(
+        'user:default/tom',
+        'group:default/team-b',
```
AndrienkoAleksandr, line 1118:
```suggestion
        'role:default/team-b',
```

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -581,6 +1045,80 @@ describe('BackstageRoleManager', () => {
...
+        }
+        return { items: [] };
+      });
+
+      roleManager.addLink('group:default/team-a', 'role:default/team-a');
+      roleManager.addLink('group:default/team-b', 'role:default/team-b');
+
+      let result = await roleManager.hasLink(
+        'user:default/mike',
+        'group:default/team-a',
```
AndrienkoAleksandr, line 1109:
```suggestion
        'role:default/team-a',
```

plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -72,12 +110,19 @@ class AncestorSearchMemo {
   }
 }
 
+function createRoleName(name: string) {
```
AndrienkoAleksandr, line 113: Sorry, but this method do nothing. It's like echo. I think we can remove it.

plugins/rbac-backend/src/service/policies-rest-api.ts
PatAKnight: Right now, this will only delete a single group / user from a role. So it doesn't remove the role completely if there are other users / groups assigned to it. I figured this approach would be a good starting point.

But we can add the option to delete the entire role (all users / groups attached to a role) if there are no query params listed. Then we can go through and remove all policies that are associated with that role.

PatAKnight: I have updated it to now delete an entire role. I still left in the functionality to delete a single user / group from a role and will just document the differences for now.

gorkem (approved):
lgtm

gorkem (approved):
lgtm

PR #868, opened 2023-10-23
chore(cli): remove unused code

tumido:
First pass on CLI cleanup. Leave only the commands we want and plan to use:

- `new` will be used to bootstrap new plugins (tailored experience for Janus IDP only)
- `package export-dynamic-plugin` - exports either frontend or backend plugin as a dynamic one
- `package start|build` - for frontend core package. (Frontend app needs to be built with Scaplrum support baked in, therefore I leave this command in here)

All other commands are removed + dead code as well + (hopefully) all its NPM dependencies.


tumido:
cc @davidfestal

I've tested the dynamic backend plugin output and it yields 0 diff, so this should not break anything.

cc @Hyperkid123

davidfestal (approved):

PR #882, opened 2023-10-25
chore: add lightweight dev setup e2e tests

gashcrumb:
This change adds a simple cypress setup and tests to act as a simple
sanity check for the tekton plugin's development setup.  Cypress is
added along with an additional script to easily run this e2e test from
the command line and potentially CI.  All changes are currently only
applied to the tekton plugin.

Interestingly enough, recent updates to the tekton plugin seem to have broken the tekton test since I first wrote it that needs a further look :-)


invincibleJai:
@gashcrumb yeah tekton went through major update and the would not work now

gashcrumb:
Closing this as we're moving to Playwright and this should be reworked with that framework.

PR #886, opened 2023-10-26
fix(rbac): fix service to service requests for RBAC CRUD

AndrienkoAleksandr:
# What does this pull request do:

PermissionEvaluator( implementation ServerPermissionClient) uses TokenManager to provide service to service requests:

https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-node/src/ServerPermissionClient.ts#L86 or
https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-node/src/ServerPermissionClient.ts#L95

https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-node/src/ServerPermissionClient.ts#L117
https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-node/src/ServerPermissionClient.ts#L106

To prevent failing code on service to service request we should remove identity.getIdentity({ request }) usage from code. This stuff any way used in the permission handler implementation under the hood:

identity: https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-backend/src/service/router.ts#L166
https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-backend/src/service/router.ts#L191C26-L191C34
https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-backend/src/service/router.ts#L104
https://github.com/backstage/backstage/blob/a503b2d8fae30f09cc75ab727cc57b939f92e6df/plugins/permission-backend/src/service/router.ts#L120

# How to  test it

Generate new plugin with pluginId test: https://backstage.io/docs/plugins/backend-plugin/#creating-a-backend-plugin.  Define this test backend plugin in the backend/src/plugins folder:

```test.ts
import { Router } from 'express';

import { createRouter } from '@janus-idp/plugin-test-backend';

import { PluginEnvironment } from '../types';

  env: PluginEnvironment,
): Promise<Router> {

  return await createRouter({
    logger: env.logger,
    tokenManager: env.tokenManager,
  });
}
```

Register plugin in the backend/src/index.ts:

```
import test from './plugins/test';
...

  const testEnv = useHotMemoize(module, () => createEnv('test'));
  apiRouter.use('/test', await test(testEnv));
```

Provide TokenManager for this plugin using plugin env and route options. Use this route code to call rbac-backend api method:

```ts
import { errorHandler, TokenManager } from '@backstage/backend-common';

import express from 'express';
import Router from 'express-promise-router';
import { Logger } from 'winston';

export interface RouterOptions {
  logger: Logger;
  tokenManager: TokenManager;
}

export async function createRouter(
  options: RouterOptions,
): Promise<express.Router> {
  const { logger } = options;

  const router = Router();
  router.use(express.json());

  router.get('/health', async (_, response) => {
    const { token } = await options.tokenManager.getToken(); 
    const url = 'http://localhost:7007/api/permission/policies';

    const resp = await fetch(url, {
      method: 'GET',
      headers: {
        Authorization: `Bearer ${token}`,
      },
    });

    const body = await resp.text();
    console.log(`=== ${resp.status}. Response Body: ${JSON.stringify(body)}`);

    logger.info('PONG!');
    response.json({ code: resp.status });
  });
  router.use(errorHandler());
  return router;
}

```

Launch backstage-plugins and make request:

```
 curl http://localhost:7007/api/test/health
```
 
This REST method should call backend plugin. Backend plugin should make service to service request to rbac-backend policy api.

### DIFF
--- a/plugins/rbac-backend/src/service/policies-rest-api.ts
+++ b/plugins/rbac-backend/src/service/policies-rest-api.ts
@@ -469,14 +469,9 @@ export class PolicesServer {
     );
 
     router.get('/conditions', async (req, resp) => {
-      const decision = await this.authorize(
-        this.identity,
-        req,
-        this.permissions,
-        {
-          permission: policyEntityReadPermission,
-        },
-      );
+      const decision = await this.authorize(req, {
+        permission: policyEntityReadPermission,
+      });
 
       if (decision.result === AuthorizeResult.DENY) {
         throw new NotAllowedError(); // 403
@@ -493,14 +488,9 @@ export class PolicesServer {
     });
 
     router.post('/conditions', async (req, resp) => {
-      const decision = await this.authorize(
-        this.identity,
-        req,
-        this.permissions,
-        {
-          permission: policyEntityCreatePermission,
-        },
-      );
+      const decision = await this.authorize(req, {
+        permission: policyEntityCreatePermission,
+      });
 
       if (decision.result === AuthorizeResult.DENY) {
         throw new NotAllowedError(); // 403
@@ -515,14 +505,9 @@ export class PolicesServer {
     });
 
     router.get('/conditions/:id', async (req, resp) => {
-      const decision = await this.authorize(
-        this.identity,
-        req,
-        this.permissions,
-        {
-          permission: policyEntityReadPermission,
-        },
-      );
+      const decision = await this.authorize(req, {
+        permission: policyEntityReadPermission,
+      });
 
       if (decision.result === AuthorizeResult.DENY) {
         throw new NotAllowedError(); // 403
@@ -542,14 +527,9 @@ export class PolicesServer {
     });
 
     router.delete('/conditions/:id', async (req, resp) => {
-      const decision = await this.authorize(
-        this.identity,
-        req,
-        this.permissions,
-        {
-          permission: policyEntityDeletePermission,
-        },
-      );
+      const decision = await this.authorize(req, {
+        permission: policyEntityDeletePermission,
+      });
 
       if (decision.result === AuthorizeResult.DENY) {
         throw new NotAllowedError(); // 403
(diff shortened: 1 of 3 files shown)


gorkem (changes requested):
This is actually giving more permissions to other services than I am comfortable with. This would enable any plugin to bypass rbac admin. Can we limit the machine permissions only to view only and remove it from put/post/delete

gorkem (changes requested):
This is actually giving more permissions to other services than I am comfortable with. This would enable any plugin to bypass rbac admin. Can we limit the machine permissions only to view only and remove it from put/post/delete

AndrienkoAleksandr:
> Can we limit the machine permissions only to view only and remove it from put/post/delete

Done.

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -52,21 +53,20 @@ export class PolicesServer {
...
-    identity: IdentityApi,
     request: Request,
-    permissionEvaluator: PermissionEvaluator,
     permission: QueryPermissionRequest,
-  ) {
-    const user = await identity.getIdentity({ request });
-    if (!user) {
-      throw new NotAllowedError();
+  ): Promise<PolicyDecision> {
+    if (permission.permission !== policyEntityReadPermission) {
```
gorkem, line 59: This only works for machine-to-machine tokens now. What would happen if I were a user with read or write permission? We still need to allow users with correct test of rights.

gorkem (changes requested):

gorkem (approved):

plugins/rbac-backend/src/service/policies-rest-api.ts
AndrienkoAleksandr: During talk in the slack with Gorkem we checked and confirmed, that everything should work fine for user token too.

AndrienkoAleksandr:
I rebase pull request and fixed compilation.

PR #891, opened 2023-10-27
fix(rbac): add models folder and config.d.ts to package

PatAKnight:
This PR adds the models folder with the policy conf file and `config.d.ts` to the package.json


gorkem (approved):
lgtm

PR #902, opened 2023-11-02
fix(scalprum): share react-router since various 3rd party plugins use it

tumido:
Seems like sole package build does not properly run generate step on release. This should fix it.


tumido (commented):
Self-approving. Sorry, we need to fix the build

PR #909, opened 2023-11-06
fix(cli): add default scalprum config

Hyperkid123:
The `scalprum` config is no longer required. By default, CLI will use the `index.ts` as an exposed module called `PluginRoot`.

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/frontend.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/frontend.ts
@@ -26,12 +26,31 @@ export async function frontend(
   _: PackageRoleInfo,
   __: OptionValues,
 ): Promise<void> {
-  const { name, version, scalprum } = await fs.readJson(
-    paths.resolveTarget('package.json'),
-  );
+  const {
+    name,
+    version,
+    scalprum: scalprumExternal,
+  } = await fs.readJson(paths.resolveTarget('package.json'));
+
+  let scalprum = scalprumExternal;
   if (scalprum === undefined) {
-    throw new Error(
-      `Package doesn't seem to support dynamic loading. It should have a 'scalprum' key in 'package.json' containing the dynamic loading entrypoints.`,
+    let scalprumName;
+    if (name.includes('/')) {
+      const fragments = name.split('/');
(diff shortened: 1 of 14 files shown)


tumido:
Hi @Hyperkid123 ,

 We need the cli to be usable by other actors outside the Janus IDP community, by our users as well. Therefore we can't assume the package name to be  prefixed with `janus-idp.`.

I thought by "supporting defaults" we mean allowing a situation where the user doesn't have to specify the `module` key in their app config. And scalprum would know to use the `default` entry point from the `exportedModules`. Aka getting rid of an arbitrary name in the `exportedModules` config where there's only one module exported and therefore reducing the bloat.

I think we should still require the scalprum config to be explicitly stated. This means people want the package to be ready for consumption as a dynamic plugin. WDYT?

Hyperkid123:
@tumido I've updated the generated name to not be janus specific, but rather fully derived from package name.

plugins/quay/package.json
```diff
@@ -69,11 +69,5 @@
     "plugin"
   ],
   "homepage": "https://janus-idp.io/",
-  "bugs": "https://github.com/janus-idp/backstage-plugins/issues",
-  "scalprum": {
```
tumido, line 73: Can you please remove this from all the other plugins as well?

Hyperkid123: Sure

packages/cli/src/commands/export-dynamic-plugin/frontend.ts
```diff
@@ -26,12 +26,31 @@ export async function frontend(
...
+  let scalprum = scalprumExternal;
   if (scalprum === undefined) {
-    throw new Error(
-      `Package doesn't seem to support dynamic loading. It should have a 'scalprum' key in 'package.json' containing the dynamic loading entrypoints.`,
+    let scalprumName;
+    if (name.includes('/')) {
+      const fragments = name.split('/');
+      scalprumName = `${fragments[0].replace('@', '')}.${fragments[1]}`;
+    } else {
+      scalprum = name;
```
tumido, line 42:
```suggestion
      scalprumName = name;
```

tumido (changes requested):

tumido (approved):
/lgtm
/approve

PR #911, opened 2023-11-07
chore(deps): unlock local cli version in private packages

tumido:
SSIA

The release tooling can bump dependant versions in packages that are set to be released (`private: false`). And it doesn't really make sense to lock version in private packages for local dependencies anyways...


invincibleJai (approved):
LGTM

PR #912, opened 2023-11-07
feat(cli): generate plugin specific schema for dynamic plugins

tumido:
We need to generate a config schema for each dynamic plugin since they do not end up in the regular `node_modules` dependencies, therefore it's not crawled and discovered by the Backstage schema collect mechanism. This leads to visibility issues (frontend is unable to fetch values, backend can expose secrets in logs).

Config schema can also be stored in transitive dependencies, not within the plugin itself, so we need to collect schemas in similar way as Backstage does.

However, Backstage schema collection mechanism crawls **all** dependencies - which essentially gets all the way back to Bascktage core - that essentially means, each plugin specific schema would be a full Backstage instance schema. Therefore I had to do a heuristic filtering out the most common schema packages to hopefully reach plugin specific additions only and a reasonable schema size.

Schema is stored in `dist-scalprum/configSchema.json` and `dist-dynamic/dist/configSchema.json`.


davidfestal (approved):

davidfestal:
@tumido 
Approved, though maybe it would be great to just upgrade the `dynamic-frontend-plugin` `@janus-idp/cli` version to avoid the CI build error ?

PR #919, opened 2023-11-08
feat(ocm): export default icon for use in dynamic plugins

tumido:
Part of: https://github.com/janus-idp/backstage-showcase/issues/725

Export default OCM icon as part of the dynamic plugin assets.


invincibleJai (approved):
LGTM

Issue #913, opened 2023-11-07
rbac: rbac-backend plugin fails with  ssl database connection


AndrienkoAleksandr:
### Describe the bug

rbac-backend plugin fails with  ssl database connection

### Expected Behavior
rbac-backend plugin should  support ssl database connection.

### What are the steps to reproduce this bug?
Issue was reproduced with showcase gitops installation. Thanks @PatAKnight for but report.

kadel:
Bug has been fixed in the plugin, but we still need to upgrade the plugin in the showcase

AndrienkoAleksandr:
@kadel, thanks. https://github.com/janus-idp/backstage-showcase/pull/761

PR #923, opened 2023-11-09
fix(rbac): handle postgres ssl connection for rbac backend plugin

AndrienkoAleksandr:
# What doest this pull request do:

https://github.com/janus-idp/backstage-plugins/issues/913


PatAKnight (approved):
LGTM

PR #934, opened 2023-11-14
chore: move janus configs to `app-config.janus-idp.yaml`

davidfestal:
Rename the Janus-related plugin-specific default configuration snippet to `app-config.janus-idp.yaml`, in order to make it clear that these are the default settings when the plugin is used in the janus IDP.


tumido (approved):
/lgtm
/approve

PR #895, opened 2023-10-30
feat(rbac): display administration to authorized users

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/842

**Test setup:**
```app-config.yaml```
```
auth:
  providers:
    github:
      development:
        clientId: $GITHUB_CLIENT_ID
        clientSecret: $GITHUB_CLIENT_SECRET
        
permission:
  enabled: true
  rbac:
    admin:
      users:
        - name: user:default/<your-github-id>
        - name: group:default/admins
```

```DynamicRoot.tsx```
```
Add SignInPage component 

    if (!app.current) {
      app.current = createApp({
        apis,
        components: {
          SignInPage: props => (
            <SignInPage
              {...props}
              auto
              provider={{
                id: 'github-auth-provider',
                title: 'GitHub',
                message: 'Sign in using GitHub',
                apiRef: githubAuthApiRef,
              }}
            />
          ),
        },
```

```auth.ts```
```
      github: providers.github.create({
        signIn: {
          resolver({ result }, ctx) {
            const userRef = `user:default/${result.fullProfile.username}`;
            return ctx.issueToken({
              claims: {
                sub: userRef, // The user's own identity
                ent: [userRef], // A list of identities that the user claims ownership through
              },
            });
          },
          // resolver: providers.github.resolvers.usernameMatchingUserEntityName(),
        },
      }),
```

**_To load the Administration sidebar item dynamically_**

Test Setup

Add the `@janus-idp/backstage-plugin-rbac` package in the packages/app/package.json
Keep the app-config changes from the above set-up

app-config.yaml

```
dynamicPlugins:
  frontend:
    janus.dynamic-frontend-plugin:
      dynamicRoutes:
        …
        - path: '/rbac'
          importName: 'RbacPage'
          module: 'Administration'


```

Root.tsx

```
	export const Root = ({ children }: PropsWithChildren<{}>) => (
	…
        <ScalprumComponent
        scope="janus.dynamic-frontend-plugin"
        module="Administration"
        importName="Administration"
      />
	…
```

dynamic-modules-registry.json

```
      {
   	"module": "Administration",
   	 "scope": "janus.dynamic-frontend-plugin",
    	"manifestLocation": "http://localhost:8004/plugin-manifest.json"
  	}
```

dynamic-frontend-plugin/package.json

```
"scalprum": {
    …
    "exposedModules": {
       …
      "Administration": "./src/remotes/Administration"
    }
  }
```

[...]

Issue #843, opened 2023-10-17
 [RBAC Plugin] Show list of Roles in table with actions to delete


invincibleJai:
What do you want to improve?
Show a list of roles in the table if the administration is selected

- [ ] Show page heading and Roles tab
- [ ] show roles in the table with all the columns as in the screenshot below
- [ ] filter/sort/pagination on the table should be supported
- [ ] with actions should be able to delete. Ask for confirmation from user

Screenshots:

<img width="1208" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/be36f880-6574-448f-8ef5-5be53e5c0159">

PR #937, opened 2023-11-15
feat(rbac): List roles

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/843

**Screenshots/GIF:**
Screenshot 2023-11-15 at 5 09 39 PM

User not authorized to delete roles

Test setup:
Add this in the rbac-policy.csv
`p, user:default/<username>, policy-entity, delete, deny`

Screenshot 2023-11-15 at 5 12 11 PM

https://github.com/janus-idp/backstage-plugins/assets/22490998/a96c7b40-5e1e-47c6-b73c-215ce56b28d8

Test setup:

Follow the steps mentioned in the PR https://github.com/janus-idp/backstage-plugins/pull/895


plugins/rbac/src/components/DeleteRoleDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const rbacApi = useApi(rbacApiRef);
+
+  const deleteRole = async () => {
+    try {
+      await rbacApi.deleteRole(roleName);
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(err);
+    }
```
divyanshiGupta, line 58: @debsmita1 If an error occurs while deleting a role, IMO we should notify the users that there was some error while deleting the role and they can try again or something similar.

cc: @invincibleJai @parvathyvr

plugins/rbac/src/components/DeleteRoleDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+      color: theme.palette.grey[500],
+    },
+  }),
+);
+
+type DeleteRoleDialogProps = {
+  open: boolean;
+  closeDialog: () => void;
+  roleName: string;
+};
```
divyanshiGupta, line 42: nit: Add new line after type

plugins/rbac/src/components/RolesList.tsx
```diff
@@ -0,0 +1,131 @@
...
+      title: 'Modified by',
+      field: 'modifiedBy',
+      type: 'string',
+    },
+    {
+      title: 'Last modified',
+      field: 'lastModified',
+      type: 'string',
+    },
+  ];
```
divyanshiGupta, line 81: This can also be moved to a separate fille like `RolesListColumns`.

divyanshiGupta (changes requested):

plugins/rbac/src/components/RolesList.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const closeDialog = () => {
+    setOpen(false);
+  };
+  const useStyles = makeStyles(theme => ({
+    empty: {
+      padding: theme.spacing(2),
+      display: 'flex',
+      justifyContent: 'center',
+    },
+  }));
```
divyanshiGupta, line 33: Can you move this out of this component?

divyanshiGupta (changes requested):

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -0,0 +1,53 @@
...
+    () =>
+      !roles || roles.length === 0
+        ? []
+        : roles.reduce(
+            (acc: any, role: Role) => [
+              ...acc,
+              {
+                id: role.name,
+                name: role.name,
+                description: '-',
```
invincibleJai, line 37: shall we hardcode `-` or use `role.description` and once the backend produces the data it'll be handled? 

```suggestion
                description: role.description || '-',
```

same applies for `modifiedBy` and `lastModified`. If for some field backend will not provide data in near term then we should not show that column as well

cc @PatAKnight

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -0,0 +1,53 @@
...
+                ),
+                modifiedBy: '-',
+                lastModified: '-',
+              },
+            ],
+            [],
+          ),
+    [roles, policies],
+  );
+  useInterval(() => retry(), 3000);
```
invincibleJai, line 51: shall we have it or provide user an option to configure it as optional and if not provided default to `3000`

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -0,0 +1,43 @@
+import { isUserEntity, parseEntityRef } from '@backstage/catalog-model';
+
+import { RoleBasedPolicy } from '@janus-idp/backstage-plugin-rbac-common';
+
+export const getPermissions = (role: string, policies: RoleBasedPolicy[]) => {
```
invincibleJai, line 5: nit: return type for this util

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (!policies || policies?.length === 0) {
+    return 0;
+  }
+  return policies.filter(
+    (policy: RoleBasedPolicy) => policy.entityReference === role,
+  ).length;
+};
+
+export const getMembers = (memberReferences: string[]) => {
```
invincibleJai, line 14: nit: return type for this util

plugins/rbac/src/hooks/useRoles.ts
PatAKnight: Ideally, we aim to release the update with `modifiedBy`, `description`, and `lastModified` as soon as we receive the green light to begin working on them. However, I believe it might be a good idea to temporarily omit the columns instead of leaving it blank or with the hardcoded `-`. This way, the feature remains functional before the next release, and we can potentially avoid complaints about missing data / empty data.

invincibleJai: Thanks @PatAKnight 

@parvathyvr FYI

debsmita1:
Updated the GIF @parvathyvr @invincibleJai @divyanshiGupta

parvathyvr:
@debsmita1 some of the updates to this design are

1. Column name - Members changed to 'Users and groups'
2. Column name- Permission changed to 'Permission policies'
3. Delete icon will be shown all times- if user do not have delete access- we plan to show the delete icon disabled with a tooltip explanation. [Figma link](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=922-6877&mode=design&t=z47Hl1dlQAdvQG0v-4)
4. The delete modal-included some more context info for the user- referring to the part where we talk about the users and permission policies in that role Eg: 'The 5 users and 12 groups associated with this role will lose access to all the 3 permission policies specified in this role.' [Figma link](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=932-1999&mode=design&t=z47Hl1dlQAdvQG0v-4)

Let me know if there are any questions-Thank you!

ShiranHi:
Hey @debsmita1, it looks good, it is exciting to see this :)

I have some comments:
1. The tabs looks like they are "floating", can we align it with the header?
2. When the delete button on a row is disabled, do we implemented the tooltip that explains why?
3. The delete role modal
- It is too wide, can we have a max-width there? We can use the [Material Design guidelines](https://m3.material.io/components/dialogs/specs), max-width 560dp, minimum 280dp
- The red error should not take the whole wide of the modal
- The error text can be improved. New text: TBD
- Can the "Name" placeholder be "Role name"? We can make the text field wider if needed. If we do this change we do not need the description text
- The button should not be "ok" but "Delete", and should be main button (not a secondary button), and Red
- When user types the name in the field it turns to be Red, I think we can remove this validation and just enable the "Delete" button when the role name added correctly

debsmita1:
> Hey @debsmita1, it looks good, it is exciting to see this :)
> 
> I have some comments:
> 
> 1. The tabs looks like they are "floating", can we align it with the header?
> 2. When the delete button on a row is disabled, did we implement the tooltip that explains why?

Yes, the tooltip says "Role cannot be deleted"


> 3. The delete role modal
> 
> * It is too wide, can we have a max-width there? We can use the [Material Design guidelines](https://m3.material.io/components/dialogs/specs), max-width 560dp, minimum 280dp
> * The red error should not take the whole wide of the modal
> * The error text can be improved. New text: TBD
> * Can the "Name" placeholder be "Role name"? We can make the text field wider if needed. If we do this change we do not need the description text
> * The button should not be "ok" but "Delete", and should be main button (not a secondary button), and Red
> * When user types the name in the field it turns to be Red, I think we can remove this validation and just enable the "Delete" button when the role name added correctly


https://github.com/janus-idp/backstage-plugins/assets/22490998/c69a2d1c-31bf-42b8-b35a-ed088a005404

Screenshot 2023-11-23 at 11 49 49 PM

parvathyvr:
@debsmita1 About the inline error message -I assume we will be able to show an error message from the back end explaining why the delete didn't work.For now,in the screenshot after 'Unable to delete'...it looks incomplete. 
Apart from the error message part, i think you have covered all the updates :+1:
CC @ShiranHi do take a look and see if any updates are needed.

invincibleJai:
Thanks @debsmita1 , i see below behaviour

- No of roles doesn't update on filter

image

- On successful deletion of role i don't see any confirmation to user

plugins/rbac/src/components/RolesList.tsx
```diff
@@ -0,0 +1,97 @@
...
+    setOpen(true);
+  };
+
+  const closeDialog = () => {
+    setOpen(false);
+    retry();
+  };
+
+  React.useEffect(() => {
+    columns.push({
```
invincibleJai, line 36: why we needed this to be added here?

plugins/rbac/src/hooks/useRoles.ts
invincibleJai: @debsmita1 what's the decision on this?

plugins/rbac/src/components/DeleteRoleDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+      color: theme.palette.grey[500],
+    },
+  }),
+);
+
+type DeleteRoleDialogProps = {
+  open: boolean;
+  closeDialog: () => void;
+  roleName: string;
+  propOptions?: any;
```
divyanshiGupta, line 45: Can we have a type for this?

divyanshiGupta (changes requested):

ShiranHi:
> > * When user types the name in the field it turns to be Red, I think we can remove this validation and just enable the "Delete" button when the role name added correctly

Thanks for the updates, @debsmita1! It seems fine to me. Just noticed a couple of minor things:
1. There's an extra space before the "?" in the modal title. It should be "Delete this role?" instead of "Delete this role ?"
2. The spacing of the text in the modal seems a bit off:
- Could we add more padding between "The 1 user associated..." and the text field?
- Also, the padding below the buttons and on the right-hand side looks a bit off
Would appreciate it if you could take a look at these. Thanks!

@parvathyvr fyi

plugins/rbac/src/components/RolesList.tsx
debsmita1: updated the logic

debsmita1:
> Thanks @debsmita1 , i see below behaviour
> 
> * No of roles doesn't update on filter
> 
> image
> * On successful deletion of role i don't see any confirmation to user


Done
https://github.com/janus-idp/backstage-plugins/assets/22490998/a6348ff6-50d0-46f3-a62e-396062239962

debsmita1:
@ShiranHi @parvathyvr I have updated the GIF. PTAL

plugins/rbac/src/hooks/useRoles.ts
debsmita1: @invincibleJai  omitted the field

debsmita1:
> @debsmita1 i notice that the dismissable banner we show to indicate the successful role deletion is small and can be easily missed. Can we move towards the center of the screen like in this [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=932-1751&mode=design&t=XrFaX3zfIQazd4Z2-4)? Also see that the [backstage design system banners](https://www.figma.com/file/wlH5SQTQiIMHp5T0t7P4aU/Backstage-Design-System-(Community)?type=design&node-id=137-771&mode=design&t=EjZjfrqwdtXFAqsQ-4) can be long as shown here- these ones have a width of 861px CC @ShiranHi

Screenshot 2023-11-27 at 11 27 29 PM

plugins/rbac/src/components/ToastContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+  setToastMessage: (message: string) => void;
+};
+
+export const ToastContext = createContext<ToastContextType>({
+  toastMessage: '',
+  setToastMessage: () => {},
+});
+
+export const ToastContextProvider = (props: any) => {
+  const [showToast, setShowToast] = React.useState(false);
```
invincibleJai, line 14: this is no longer being used so we can avoid it?

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const useRoles = (pollInterval?: number) => {
+  const rbacApi = useApi(rbacApiRef);
+  const {
+    loading: rolesLoading,
+    value: roles,
+    retry,
+  } = useAsyncRetry(async () => await rbacApi.getRoles());
+
+  const { loading: policiesLoading, value: policies } = useAsync(
```
invincibleJai, line 25: should we handle `error` case as well?

invincibleJai:
Hi @debsmita1 , i see issue when role or policy call fails 

image

invincibleJai:
and should toast auto dismiss after a certain time interval? and what should be the experience for toast if is user to deletes multiple roles without dismissing the toast? currently same toast gets updated with the last deleted role name

cc @debsmita1  @parvathyvr

plugins/rbac/src/hooks/useRoles.ts
debsmita1: I am showing "No roles found" when there's an error

Screenshot 2023-11-28 at 1 30 10 PM

debsmita1:
> and should toast auto dismiss after a certain time interval? and what should be the experience for toast if is user to deletes multiple roles without dismissing the toast? currently same toast gets updated with the last deleted role name
> 
> cc @debsmita1 @parvathyvr

The Toast will auto-dismiss after 6 secs, I missed passing onClose callback. PTAL

invincibleJai (approved):
LGTM

divyanshiGupta:
@debsmita1 Everything is working as expected. I just see one warning on reloading the list page. LMK if you want to handle it in this PR or we can have another PR for this.

https://github.com/janus-idp/backstage-plugins/assets/20724543/c14fdff3-c1a9-465e-95d3-bfe048fce5d4

divyanshiGupta (approved):
LGTM

Issue #935, opened 2023-11-14
rbac: resource permission defined by permission name doesn't work


AndrienkoAleksandr:
### Describe the bug

Resource permission defined by permission name doesn't work. For example read catalog permission has in the same time resource name and resource type:

```
export const RESOURCE_TYPE_CATALOG_ENTITY = 'catalog-entity';
...
export const catalogEntityReadPermission = createPermission({
  name: 'catalog.entity.read',
  attributes: {
    action: 'read',
  },
  resourceType: RESOURCE_TYPE_CATALOG_ENTITY,
});
```

### Expected Behavior
Such kind of permissions should work.

### What are the steps to reproduce this bug?

1. Define permission policies:

```
p, role:default/team_a, catalog.entity.read, read, allow

g, user:default/andrienkoaleksandr, role:default/team_a
```

2. Execute backstage-plugins instance. User user:default/andrienkoaleksandr should ability read catalog in the UI.

### Known workaround

Define this permission by resource type:

```
p, role:default/team_a, catalog-entity, read, allow

g, user:default/andrienkoaleksandr, role:default/team_a
```

PR #940, opened 2023-11-16
fix(rbac): fix work resource permission specified by name

AndrienkoAleksandr:
# What does this pull request do:

https://github.com/janus-idp/backstage-plugins/issues/935

### DIFF
--- a/plugins/rbac-backend/src/service/role-manager.test.ts
+++ b/plugins/rbac-backend/src/service/role-manager.test.ts
@@ -41,12 +41,6 @@ describe('BackstageRoleManager', () => {
       ).toThrow('Method "syncedHasLink" not implemented.');
     });
 
-    it('should throw an error for getRoles', async () => {
-      await expect(roleManager.getRoles('name')).rejects.toThrow(
-        'Method "getRoles" not implemented.',
-      );
-    });
-
     it('should throw an error for getUsers', async () => {
       await expect(roleManager.getUsers('name')).rejects.toThrow(
         'Method "getUsers" not implemented.',
@@ -1119,6 +1113,16 @@ describe('BackstageRoleManager', () => {
         'role:default/team-b',
       );
       expect(result).toBeTruthy();
+    });
+  });
+
+  describe('getRoles test', () => {
+    it('should returns 2 roles for user', async () => {
+      roleManager.addLink('user:default/test', 'role:default/rbac_admin');
+      const roles = await roleManager.getRoles('user:default/test');
+      expect(roles.length).toBe(2);
+      expect(roles[0]).toEqual('user:default/test');
+      expect(roles[1]).toEqual('role:default/rbac_admin');
     });
   });
   function createGroupEntity(
--- a/plugins/rbac-backend/src/service/role-manager.test.ts
+++ b/plugins/rbac-backend/src/service/role-manager.test.ts
@@ -1116,15 +1116,74 @@ describe('BackstageRoleManager', () => {
     });
   });
 
-  describe('getRoles test', () => {
-    it('should returns 2 roles for user', async () => {
+  describe('getRoles returns roles per user', () => {
+    it('should returns role per user', async () => {
       roleManager.addLink('user:default/test', 'role:default/rbac_admin');
-      const roles = await roleManager.getRoles('user:default/test');
-      expect(roles.length).toBe(2);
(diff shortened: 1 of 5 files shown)


plugins/rbac-backend/src/service/role-manager.ts
```diff
@@ -240,7 +240,7 @@ export class BackstageRoleManager implements RoleManager {
    * domain is a prefix to the roles.
    */
   async getRoles(_name: string, ..._domain: string[]): Promise<string[]> {
-    throw new Error('Method "getRoles" not implemented.');
+    return Promise.resolve([...this.allRoles.keys()]);
```
PatAKnight, line 243: So, this is returning all roles within Casbin, but the comment above mentions that it returns the roles that the name inherits. Should we instead filter through `allRoles` and return the roles that a user is associated with using `hasLink`? Or will returning `allRoles` be fine for other inheritance cases?

I'm seeking clarification since I haven't seen into many inheritance scenarios come up yet, making me wonder about what potential issues may arise from returning all roles through `getRoles`.

plugins/rbac-backend/src/service/role-manager.test.ts
```diff
@@ -1121,6 +1115,16 @@ describe('BackstageRoleManager', () => {
...
   });
+
+  describe('getRoles test', () => {
+    it('should returns 2 roles for user', async () => {
+      roleManager.addLink('user:default/test', 'role:default/rbac_admin');
+      const roles = await roleManager.getRoles('user:default/test');
+      expect(roles.length).toBe(2);
+      expect(roles[0]).toEqual('user:default/test');
+      expect(roles[1]).toEqual('role:default/rbac_admin');
+    });
```
PatAKnight, line 1126: This follows up on my previous comment regarding `getRoles`. Here, we pass in a specific user to `getRoles` and expect the return of two roles. However, if we were to add additional links with `addLink`, it would retrieve all roles within Casbin, not just the two roles for the user.

Reproducible test
```
it('should returns 2 roles for user, additional links have been made', async () => {
      roleManager.addLink('user:default/test', 'role:default/rbac_admin');
      roleManager.addLink('user:default/test-two', 'role:default/rbac_admin');
      roleManager.addLink('user:default/test-three', 'role:default/rbac_admin_test_three');
      const roles = await roleManager.getRoles('user:default/test');
      expect(roles.length).toBe(2);
      expect(roles[0]).toEqual('user:default/test');
      expect(roles[1]).toEqual('role:default/rbac_admin');
    });
```
in this case we would receive 5 instead of 2.

PatAKnight (commented):
Looks pretty good so far, just had a question regarding the `getRoles` in the `RoleManager`. Also, I was able to test and it worked.

PatAKnight:
One other thing, do you mind updating this [part](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/permissions.md?plain=1#L7-L11) of the documentation. It will no longer be necessary with your changes.

plugins/rbac-backend/src/service/role-manager.ts
AndrienkoAleksandr: @PatAKnight , thanks. Really great catch. Awesome!

plugins/rbac-backend/src/service/role-manager.test.ts
AndrienkoAleksandr: Thanks, I counted you proposal in the code and tests. But I think we should return only "role" string entity references without users for method getRoles.

PatAKnight:
> One other thing, do you mind updating this [part](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/permissions.md?plain=1#L7-L11) of the documentation. It will no longer be necessary with your changes.

Actually, instead of removing it, could you add something along the lines that of:

Permission policies defined using the name of the permission will have higher priority over permission policies that are defined using the resource type.
Example:
```
p, role:default/myrole, catalog-entity, read, allow
p, role:default/myrole, catalog.entity.read, read, deny
g, user:default/myuser, role:default/myrole
```
Where 'myuser' will have a deny for reading catalog entities, because the permission name takes priority over the permission resource type

AndrienkoAleksandr:
@PatAKnight, thanks, I update rbac-backend docs with you suggestions.

PatAKnight (approved):
/lgtm

PR #938, opened 2023-11-15
fix(keycloak): don't log sensitive internal data

jerolimov:
Fixes #629

/cc @schultzp2020 @kadel

Tested two cases, once with connection refused and once with invalid credentials.

### With developing logging

```text
backend:start:backstage: 2023-11-15T12:14:32.837Z catalog info Reading Keycloak users and groups type=plugin class=KeycloakOrgEntityProvider taskId=KeycloakOrgEntityProvider:default:refresh taskInstanceId=e5833bf0-4ad4-45f1-8cba-8929b034d875
backend:start:backstage: 2023-11-15T12:14:32.839Z catalog error connect ECONNREFUSED ::1:8080 type=plugin class=KeycloakOrgEntityProvider taskId=KeycloakOrgEntityProvider:default:refresh taskInstanceId=e5833bf0-4ad4-45f1-8cba-8929b034d875 errno=-111 code=ECONNREFUSED syscall=connect address=::1 port=8080 config=[object Object] request=[object Object] response=undefined isAxiosError=true toJSON=function toJSON(){return{message:this.message,name:this.name,description:this.description,number:this.number,fileName:this.fileName,lineNumber:this.lineNumber,columnNumber:this.columnNumber,stack:this.stack,config:this.config,code:this.code,status:this.response&&this.response.status?this.response.status:null}} stack=Error: connect ECONNREFUSED ::1:8080
backend:start:backstage:     at __node_internal_captureLargerStackTrace (node:internal/errors:496:5)
backend:start:backstage:     at __node_internal_exceptionWithHostPort (node:internal/errors:671:12)
backend:start:backstage:     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1555:16)
```

### With production logging before

```json
backend:start:backstage: {
  "class": "KeycloakOrgEntityProvider",
  "level": "info",
  "message": "Reading Keycloak users and groups",
  "plugin": "catalog",
  "service": "backstage",
  "taskId": "KeycloakOrgEntityProvider:default:refresh",
  "taskInstanceId": "30a64bd9-1b44-4b45-9032-dfc27a81db3a",
  "type": "plugin"
}

backend:start:backstage: {
  "code": "ECONNREFUSED",
  "config": {
    "data": "client_id=admin-cli&grant_type=password&password=admin2&username=admin",  // Exposed credentials
    "headers": {  // Would contains secrets when the first network call would be successful
      "Accept": "application/json, text/plain, */*",
      "Content-Length": 70,
[...]

PR #945, opened 2023-11-16
fix(aap+3scale+ocm): don't log sensitive data from errors

jerolimov:
This applies a similar change as #938 to the same issues for AAP, 3Scale, and OCM providers.

Changes:
* Extract the essential error data from the error object (name, message, stack, status code).
* Updated `AapResourceEntityProvider.test.ts` where I could confirm that the auth header was logged before. 3scale and OCM don't have unit tests for the Providers to confirm the change.
* Added a message string to all `logger.error` statements. As best practice to make it easier find the right log statement and also to make it more compatible with [backstage LoggerService](https://github.com/backstage/backstage/blob/master/packages/backend-plugin-api/src/services/definitions/LoggerService.ts).


kadel:
Please don't forget to update aap and ocm plugins in showcase repo after this is merged.

tumido (approved):
/lgtm
/approve

invincibleJai (approved):
LGTM

PR #955, opened 2023-11-20
feat(segment): expose plugin also as a dynamic

tumido:
SSIA, makes segment frontend plugin dynamic

### DIFF
--- a/plugins/analytics-provider-segment/package.json
+++ b/plugins/analytics-provider-segment/package.json
@@ -10,11 +10,12 @@
     "types": "dist/index.d.ts"
   },
   "backstage": {
-    "role": "frontend-plugin-module"
+    "role": "frontend-plugin"
   },
   "scripts": {
     "start": "backstage-cli package start",
     "build": "backstage-cli package build",
+    "export-dynamic": "janus-cli package export-dynamic-plugin",
     "tsc": "tsc",
     "lint": "backstage-cli package lint",
     "test": "backstage-cli package test --passWithNoTests --coverage",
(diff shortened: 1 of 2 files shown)


plugins/analytics-provider-segment/package.json
```diff
@@ -10,11 +10,12 @@
     "types": "dist/index.d.ts"
   },
   "backstage": {
-    "role": "frontend-plugin-module"
+    "role": "frontend-plugin"
```
invincibleJai, line 13: why do we need to change role?

invincibleJai:
I observed an issue but this doesn't seem related to this work so have logged it https://github.com/janus-idp/backstage-plugins/issues/959 but is a quick fix , there is typo `visbility` to `visibility`

https://github.com/janus-idp/backstage-plugins/blob/main/plugins/analytics-provider-segment/config.d.ts#L40
https://github.com/janus-idp/backstage-plugins/blob/main/plugins/analytics-provider-segment/config.d.ts#L60

invincibleJai (approved):
LGTM

plugins/analytics-provider-segment/package.json
tumido: I think the `frontend-plugin-module` is just a leftover. It's actually always treated the same as `frontend-plugin` by Backstage CLI.

https://github.com/backstage/backstage/blob/d141d12529687ad199523cca272947b43dae6e05/packages/cli/config/eslint-factory.js#L184-L185

https://github.com/backstage/backstage/blob/d141d12529687ad199523cca272947b43dae6e05/packages/cli/config/jest.js#L56-L57

https://github.com/backstage/backstage/blob/d141d12529687ad199523cca272947b43dae6e05/packages/cli/src/lib/bundler/packageDetection.ts#L86

However in Janus CLI, we (for the sake of simplicity) never actually considered it. So dynamic plugin export actually fails because of:

https://github.com/janus-idp/backstage-plugins/blob/30cc4b3f68e30c255fe8c02c2ec18450ee8b2e9c/packages/cli/src/commands/export-dynamic-plugin/command.ts#L52

PR #956, opened 2023-11-21
feat: adds storybook

jkilzi:
[FLPATH-701
](https://issues.redhat.com/browse/FLPATH-701)

This PR adds [Storybook](https://storybook.js.org/).  A playground for testing and developing UI components in isolation.

## Usage

Two new top-level scripts where added: `start:storybook` and `build:storybook`, the later builds a static website which is useful for showcasing Storybook's stories.

```sh
yarn --cwd storybook install
yarn start:storybook [plugins/<frontend-plugin-name>]
```

### Notes:

1. If no `plugins/<plugin-name>` is passed, the process will scan for stories inside each plugin workspace.
2. To make a build of Storybook use: `yarn build:storybook [plugins/<frontend-plugin-name>]`.
   This command produces a directory called `packages/storybook/storybook-static`. This directory can be then rendered
   using a static web server.

## Resources
- [What is a story?](https://storybook.js.org/docs/react/get-started/whats-a-story)
- [How to write stories?](https://storybook.js.org/docs/react/writing-stories/introduction)


packages/storybook/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "storybook",
+  "version": "0.0.0",
+  "license": "Apache-2.0",
+  "private": true,
+  "backstage": {
+    "role": "frontend"
+  },
```
invincibleJai, line 8: this is not a plugin so might not be needed

packages/storybook/.eslintrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('@backstage/cli/config/eslint-factory')(__dirname);
```
invincibleJai, line 1: do we need to have `storybook` under packages or can live in root?

invincibleJai:
@tumido FYI ^^

packages/storybook/package.json
jkilzi: Correct, I would add that it is unnecessary since it is not even meant to be published to NPM or a similar repository. However, TTBOMK `@backstage/cli` requires this to properly identify packages.

packages/storybook/.eslintrc.js
jkilzi: It can be outside of `packages/`. That's how it's set up in https://github.com/backstage/backstage/tree/master/storybook

jkilzi: TL;DR I think for simplicity it's better to have it inside `packages/` as a private package

jkilzi:
@invincibleJai I'd like to mark the conversation as resolved, do you have anything to add?

invincibleJai:
> @invincibleJai I'd like to mark the conversation as resolved, do you have anything to add?

HI @jkilzi, I have looked and it and will try it and would be nice to have some components as part of story book as starting point and I would like some more eyes on this as this would be used by any other plugins to contribute in the future. currently team is a bit occupied so expect some delays

cc @tumido @jerolimov @gashcrumb

jkilzi:
> > @invincibleJai I'd like to mark the conversation as resolved, do you have anything to add?
> 
> HI @jkilzi, I have looked at it and will try it. Would be nice to have some components as part of the story book as a starting point and I would like some more eyes on this as this would be used by any other plugins to contribute in the future. currently team is a bit occupied so expect some delays
> 
> cc @tumido @jerolimov @gashcrumb

You can also see it in action in this PR https://github.com/caponetto/backstage-plugins/pull/24, I implemented a story for the Orchestrator plugin there

packages/storybook/package.json
tumido: I think the backstage cli is kinda dependent, we would need to make an explicit exception for this. So.. Unless we have a conflict with the linter settings enforced by backstage CLI, let's use it 🙂 Therefore let's keep the role in here, since it's just easier 🙂

tumido (approved):
/lgtm
/approve

tumido:
@invincibleJai do you want any more 👀 on this or are we fine to merge it? 🙂 I've resolved the conversations above, the justification provided by @jkilzi makes sense to me. I don't want to hold the team hostage by our release more than necessary and the code looks good to me. 🙂

Issue #961, opened 2023-11-21
Dynamic plugins versions are not bumped during release


tumido:
### Describe the bug

Current release flow:
1. builds the packages
2. propagate appropriate version bumps into `package.json`
3. publishes assets

Since we generate the dynamic assets in step 1, we never get version bumps into those dynamic plugin metadata (`plugin-manifest.json` and `dist-dynamic/package.json`). While the assets are correct, metadata is wrong. We need to align it.

### Expected Behavior

All metadata matches.

### What are the steps to reproduce this bug?

Release a package and observe version mismatch.

PR #963, opened 2023-11-21
fix: sync versions in dynamic assets and publish derived packages as additional packages

tumido:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/961

This fix is three-fold:

1. Add `dist-dynamic/package.json` and `dist-dynamic/yarn.lock` to the git-commited files during release
2. Add `postversion` hook to all packages that use `export-dynamic` to re-run the export dynamic command once the package is versioned (makes the build longer since `export-dynamic` is run twice)
3. Adds post publish `successCmd` to also publish the `dist-dynamic` as a separate package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,12 +133,28 @@
       ],
       "@semantic-release/changelog",
       "@semrel-extra/npm",
-      "@semantic-release/git",
+      [
+        "@semantic-release/git",
+        {
+          "assets": [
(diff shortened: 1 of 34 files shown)


tumido:
cc @davidfestal

davidfestal (approved):

PR #966, opened 2023-11-21
fix: do not fail release on missing dist-dynamic

tumido:
Revert push of derived package to fix the release pipeline.


tumido:
hot fix the release pipeline

PR #967, opened 2023-11-22
fix: make the `export-dynamic-plugin` command more generic for backend plugins.

davidfestal:
This PR makes the `export-dynamic-plugin` command more generic for backend plugins.
The goal is to have it usable outside of the restricted Janus context, especially on upstream Backstage backend packages.

Signed-off-by: David Festal <dfestal@redhat.com>


tumido (approved):
/lgtm
/approve

PR #970, opened 2023-11-22
chore: fix yarn.lock

tumido:
SSIA, should sync `yarn.lock`


tumido:
self mergining to make it in before segment plugin actually releases.

PR #971, opened 2023-11-23
fix(cli): do not change interop mode by default

davidfestal:
Do not change interop mode by default but provide an additional CLI argument to allow changing it, either globally or per imported package.

This PR also includes the related `./alpha` package when embedding a given package.


davidfestal:
To avoid any disruption risk this PR, which impacts the generated code of dynamic backend plugins, should only be merged after the showcase `1.0.x` branch code freeze.

Issue #854, opened 2023-10-18
[RBAC Plugin] Show overview for Role


invincibleJai:
What do you want to improve?
Show overview screen for Role (Check with @parvathyvr on this)

Screenshots:

<img width="971" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/51dd7099-c1d0-4d47-8a49-25d38207d641">

PR #972, opened 2023-11-23
feat(rbac): role overview

debsmita1:
Resolves:
https://github.com/janus-idp/backstage-plugins/issues/854

**GIF/Screenshots:**
https://github.com/janus-idp/backstage-plugins/assets/22490998/8583257f-eaae-4503-a781-0fa57c915499

Overview Page
Screenshot 2023-11-28 at 9 23 46 PM

Card with Error
Screenshot 2023-12-01 at 12 26 14 AM

**Test setup:**

Follow the steps mentioned in the PR https://github.com/janus-idp/backstage-plugins/pull/895 and the [documentation](https://github.com/janus-idp/backstage-plugins/tree/main/plugins/rbac) to install the RBAC UI plugin

Add 
```app-config.yaml
catalog:
  locations:
     ...
     - type: URL
        target: https://github.com/backstage/backstage/blob/master/packages/catalog-model/examples/acme-corp.yaml
        rules:
          - allow: [User, Group]
```

### DIFF
--- a/plugins/rbac/package.json
+++ b/plugins/rbac/package.json
@@ -28,6 +28,7 @@
     "@backstage/catalog-model": "^1.4.3",
     "@backstage/core-components": "^0.13.6",
     "@backstage/core-plugin-api": "^1.7.0",
+    "@backstage/plugin-catalog": "^1.15.1",
     "@backstage/plugin-permission-react": "^0.4.16",
     "@backstage/theme": "^0.4.3",
     "@janus-idp/backstage-plugin-rbac-common": "1.1.0",
@@ -36,10 +37,12 @@
     "@material-ui/lab": "^4.0.0-alpha.45",
     "@mui/icons-material": "5.14.11",
     "@mui/material": "^5.14.18",
-    "react-use": "^17.4.0"
+    "react-use": "^17.4.0",
+    "lodash": "^4.17.21"
   },
   "peerDependencies": {
-    "react": "^16.13.1 || ^17.0.0"
+    "react": "^16.13.1 || ^17.0.0",
+    "react-router-dom": "^6.20.0"
   },
   "devDependencies": {
     "@backstage/cli": "0.23.0",
@@ -49,6 +52,7 @@
     "@janus-idp/cli": "1.4.6",
     "@testing-library/jest-dom": "5.17.0",
     "@testing-library/react": "12.1.5",
+    "@testing-library/react-hooks": "8.0.1",
     "@testing-library/user-event": "14.5.1",
     "@types/node": "18.18.5",
     "msw": "1.3.2"
--- a/plugins/rbac/src/components/AboutCard.tsx
+++ b/plugins/rbac/src/components/AboutCard.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+
+import { MarkdownContent } from '@backstage/core-components';
+import { AboutField } from '@backstage/plugin-catalog';
+
+import {
+  Card,
+  CardContent,
+  CardHeader,
+  Grid,
+  makeStyles,
+} from '@material-ui/core';
+
+const useStyles = makeStyles({
+  gridItemCard: {
+    display: 'flex',
+    flexDirection: 'column',
+    height: 'calc(100% - 10px)', // for pages without content header
+    marginBottom: '10px',
+  },
+  fullHeightCard: {
+    display: 'flex',
+    flexDirection: 'column',
+    height: '100%',
+  },
+  gridItemCardContent: {
+    flex: 1,
+  },
+  fullHeightCardContent: {
+    flex: 1,
+  },
+  text: {
+    wordBreak: 'break-word',
+  },
+});
+
+export const AboutCard = () => {
+  const classes = useStyles();
+  const cardClass = classes.gridItemCard;
+  const cardContentClass = classes.gridItemCardContent;
+
+  return (
+    <Card className={cardClass}>
+      <CardHeader title="About" />
+      <CardContent className={cardContentClass}>
+        <Grid container>
+          <AboutField label="Description" gridSizes={{ xs: 4, sm: 8, lg: 4 }}>
+            <MarkdownContent
+              className={classes.text}
+              content="No description"
+            />
+          </AboutField>
+          <AboutField label="Modified By" gridSizes={{ xs: 4, sm: 8, lg: 4 }}>
+            <MarkdownContent
+              className={classes.text}
+              content="No information"
+            />
+          </AboutField>
+          <AboutField label="Last Modified" gridSizes={{ xs: 4, sm: 8, lg: 4 }}>
+            <MarkdownContent
+              className={classes.text}
+              content="No information"
+            />
+          </AboutField>
+        </Grid>
+      </CardContent>
+    </Card>
+  );
+};
--- a/plugins/rbac/src/hooks/useMembers.ts
+++ b/plugins/rbac/src/hooks/useMembers.ts
@@ -1,0 +1,99 @@
+import React from 'react';
+import { useAsync, useAsyncRetry } from 'react-use';
+
+import { stringifyEntityRef } from '@backstage/catalog-model';
+import { useApi } from '@backstage/core-plugin-api';
+
+import { get } from 'lodash';
+
+import { rbacApiRef } from '../api/RBACBackendClient';
+import { MembersData } from '../types';
+import { getKindNamespaceName, getMembersFromGroup } from '../utils/rbac-utils';
+
+const getErrorText = (
+  role: any,
+  members: any,
+): { message: string } | undefined => {
+  if (!Array.isArray(role) && get(role, 'statusText')) {
+    return {
+      message: `Unable to fetch role: ${get(role, 'statusText')}`,
+    };
+  } else if (!Array.isArray(members) && get(members, 'statusText')) {
+    return {
+      message: `Unable to fetch members: ${get(members, 'statusText')}`,
+    };
+  }
+  return undefined;
+};
+
+export const useMembers = (roleName: string) => {
+  const rbacApi = useApi(rbacApiRef);
+  let data: MembersData[] = [];
+  const {
+    loading: rolesLoading,
+    value: role,
+    retry,
+    error: roleError,
+  } = useAsyncRetry(async () => {
+    return await rbacApi.getRole(roleName);
+  });
+
+  const {
+    loading: membersLoading,
+    value: members,
+    error: membersError,
+  } = useAsync(async () => {
+    return await rbacApi.getMembers();
+  });
+
+  const loading = rolesLoading && membersLoading;
+
+  data = React.useMemo(
+    () =>
+      Array.isArray(role)
+        ? role[0].memberReferences.reduce((acc: MembersData[], ref) => {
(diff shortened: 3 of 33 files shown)


parvathyvr:
@debsmita1 the changes needed are:
1. Members table name to 'Users and groups'.
2. The number of 'Permission policies' we should show is the number of policies- Eg: if there is a single row and there are 'Read,Update,Delete' those are 3 policies.

plugins/rbac/package.json
```diff
@@ -25,16 +25,17 @@
...
+    "@janus-idp/backstage-plugin-rbac-common": "1.1.0",
     "@material-ui/core": "^4.9.13",
     "@material-ui/icons": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.45",
-    "react-use": "^17.4.0",
-    "@backstage/plugin-permission-react": "^0.4.16",
-    "@janus-idp/backstage-plugin-rbac-common": "1.1.0",
-    "@mui/icons-material": "5.14.11"
+    "@mui/icons-material": "5.14.11",
+    "react-use": "^17.4.0"
```
divyanshiGupta, line 38: @debsmita1 what is being updated in this file?

plugins/rbac/src/components/AboutCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const classes = useStyles();
+  const cardClass = classes.gridItemCard;
+  const cardContentClass = classes.gridItemCardContent;
+
+  return (
+    <Card className={cardClass}>
+      <CardHeader
+        title="About"
+        action={
+          <IconButton aria-label="Refresh" title="Refresh" onClick={() => {}}>
```
divyanshiGupta, line 49: @debsmita1 is it like a placeholder icon without any cta for now?

divyanshiGupta (changes requested):

ShiranHi:
Hey @debsmita1, thanks! Noticed a few things:

1. The "role: default/guests" header looks a bit off – thinking it should have a background like the other catalog pages, also the tabs here seem to be floating
2. The about card needs to be wider, keeping it to a single row of content. "Modified by" and "Last modified" should share the row with the "Description" text (see [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=977%3A7407&mode=design&t=uATaSCYoWnVSzNaB-1))
3. The edit button is missing in all cards
4. There's this empty space above each table, right before the column titles. Is that the table header? Can we remove it?

@parvathyvr FYI

nirfarkas:
Another small nit:
Cards should be aligned left to the header text (for example at the beginning of the video we can see that the cards are aligned to the left side of the "Administration" header, while at 00:03 we can see that they jumped to the right and are not aligned with the header.
image
image

debsmita1:
> Hey @debsmita1, thanks! Noticed a few things:
> 
> 1. The "role: default/guests" header looks a bit off – thinking it should have a background like the other catalog pages, also the tabs here seem to be floating
> 2. The about card needs to be wider, keeping it to a single row of content. "Modified by" and "Last modified" should share the row with the "Description" text (see [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=977%3A7407&mode=design&t=uATaSCYoWnVSzNaB-1))
> 3. The edit button is missing in all cards
> 4. There's this empty space above each table, right before the column titles. Is that the table header? Can we remove it?
> 
> @parvathyvr FYI

Updated
Screenshot 2023-11-27 at 8 34 45 PM

cc @nirfarkas

plugins/rbac/package.json
debsmita1: some libraries have been added

parvathyvr:
@debsmita1 Thanks for the updated screen. Few things missing:
1. For the Permission policies table-the number shown should be the number of policies selected.
2. For the Users and groups table, for type user we could keep it blank as in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=977-7407&mode=design&t=IROic36Q2uiUliEC-4). Because i don’t see a members card in the User overview page.

debsmita1:
> @debsmita1 Thanks for the updated screen. Few things missing:
> 
> 1. For the Permission policies table-the number shown should be the number of policies selected.
> 2. For the Users and groups table, for type user we could keep it blank as in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=977-7407&mode=design&t=IROic36Q2uiUliEC-4). Because i don’t see a members card in the User overview page.

Screenshot 2023-11-27 at 11 40 19 PM

divyanshiGupta:
@parvathyvr I was wondering if we should have a search bar in the members and permissions tables? Wdyt?

ShiranHi:
> * The edit button is missing in all cards
I think this is still missing

debsmita1:
> Details

@ShiranHi We haven't implemented Edit yet. There's a separate story to handle it

debsmita1:
> > * The edit button is missing in all cards
> 
> I think this is still missing

@ShiranHi We haven't implemented Edit yet. There's a separate story to handle it

parvathyvr:
> @parvathyvr I was wondering if we should have a search bar in the members and permissions tables? Wdyt?

@divyanshiGupta it's a good idea! I didn't add it since i haven't noticed search in the cards of Overview page! But if we can incorporate search into the table,it's a good addition. CC @debsmita1

debsmita1:
> > @parvathyvr I was wondering if we should have a search bar in the members and permissions tables? Wdyt?
> 
> @divyanshiGupta it's a good idea! I didn't add it since i haven't noticed search in the cards of Overview page! But if we can incorporate search into the table,it's a good addition. CC @debsmita1

Done

Screenshot 2023-11-28 at 9 23 46 PM

plugins/rbac/src/hooks/useMembers.ts
```diff
@@ -0,0 +1,54 @@
...
+    return await rbacApi.getRole(roleName);
+  });
+
+  const { loading: membersLoading, value: members } = useAsync(async () => {
+    return await rbacApi.getMembers();
+  });
+
+  const data = React.useMemo(
+    () =>
+      role?.[0].memberReferences.reduce((acc: MembersData[], ref) => {
```
invincibleJai, line 27: why do we need `0` here, you are quering based on the role name so should be just one role right? Is it  due to the response

plugins/rbac/src/hooks/usePermissionPolicies.ts
```diff
@@ -0,0 +1,30 @@
...
+export const usePermissionPolicies = (entityReference: string) => {
+  const rbacApi = useApi(rbacApiRef);
+  const {
+    loading: policiesLoading,
+    value: policies,
+    retry,
+  } = useAsyncRetry(async () => {
+    return await rbacApi.getPolicies();
+  });
+  const { value: permissionPolicies } = useAsync(async () => {
```
invincibleJai, line 18: would be good to check on loading here and in general good to check on error cased as well

invincibleJai (commented):
Thanks @debsmita1 looks good overall, good to add unit tests for hooks as well

plugins/rbac/src/hooks/useMembers.ts
debsmita1: yes, the response is an array with one role

plugins/rbac/src/hooks/usePermissionPolicies.ts
debsmita1: How do we want to show the user if there is failure in API call ?
Note: Each card makes different API calls cc @parvathyvr @ShiranHi

parvathyvr: > How do we want to show the user if there is failure in API call ? Note: Each card makes different API calls cc @parvathyvr @ShiranHi

@debsmita1 Does it make sense to use the [error banner](https://www.figma.com/file/wlH5SQTQiIMHp5T0t7P4aU/Backstage-Design-System-(Community)?type=design&node-id=137-772&mode=design&t=is0zKoWImNDw28Ap-4) in backstage design system for this scenario?

debsmita1: Screenshot 2023-11-29 at 10 29 49 PM

invincibleJai: @debsmita1 I know this is an edge case and won't happen ideally but should we show the error above the table under each card rather than inside of the table?

@parvathyvr WDYT?

parvathyvr: @invincibleJai i thought we could replace the 'No records to display' with the error message as @debsmita1 has implemented-But I am okay with the error being above the empty table as well- If that is a better option!
CC @ShiranHi do we have a preference/guideline about this?

invincibleJai: okay nice if you have considered this 👍

debsmita1: Updated

Screenshot 2023-11-30 at 1 52 50 PM

ShiranHi: @debsmita1 it looks good! Can we add a bit of a 16px space between the error message and the table ?
Also, it'll be great if a word will not break into two lines "grou-ps" and "poli-cies" like it is right now. 
Thanks for all the changes!

debsmita1: Screenshot 2023-12-01 at 12 26 14 AM

parvathyvr: A small note- @debsmita1 we are showing 'No members found' and 'No permissions found' on the empty table.
But ideally since the table heading are 'Users and groups' and 'Permission policies' i think it should be 'No users and groups found' and 'No permission policies found'.
But i also wonder if it's easier to show the generic message' No records found'.

nirfarkas: > The word breaks on small screens, what do you suggest here ?

I would suggest changing to 1 column per table to avoid text breaks as well as table misalignment.
Another reason to avoid 2 column composition is to reduce the cognitive load and showing two elements next to each other is usually for comparing them (which is not the case here).
So,
TL;DR
Moving to 1 column layout will help with:

- Get rid of the text breaks
- Allow easier data consumption with no distractions

debsmita1:
> @debsmita1 I see a mismatch in the number of users/groups and the policies in list page vs overview page
> 
> image image
> the reason for the user is, that the user is not registered however if we get the user in API we should show it irrespective of whether it's registered or not.
> 
> and the reason for policies in in list we are showing count of all the policies irrespective of effect allow or deny but in overview we only list allow once hence the mismatch @parvathyvr Let us know your thoughts here

Addressed this, now listing all users received in reponse, if a user doesn't exist simply showing "Entity not found" when the user is clicked.

Also, not listing a role if it has no users/ policies associated with it

invincibleJai (approved):
LGTM

Thanks @debsmita1

parvathyvr (approved):
@debsmita1 LGTM! Thank you.

PR #973, opened 2023-11-28
fix(servicenow-actions): revert `node-fetch` dep

davidfestal:
Revert the `node-fetch` dependency to the `2.7.0` version, in order to fix the incompatibility with ES-only modules (we use the `require()` call).

cf. https://bobbyhadz.com/blog/javascript-error-err-require-esm-require-of-es-module-not-supported#node-fetch-error-err_require_esm-require-of-es-module-not-supported

This fixes issue #976


tumido (approved):
/lgtm
/approve

Issue #942, opened 2023-11-16
[RBAC Plugin] Add create role CTA on roles list page


divyanshiGupta:
- Add create role CTA on roles list page
- The button should be disabled if user doesnt have policy.entity.create and catalog.entity.read permissions.
- Once role is created successfully user should be redirected to the roles list page

Issue #941, opened 2023-11-16
[RBAC Plugin] Add create role form


divyanshiGupta:
Based on designs https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=260-3188&mode=design&t=ViriOdEnoDWFlHoo-0 add a create role form using Stepper component.

It should have support for adding name and description. User should be able to add members and submit the form to create a role.

PR #974, opened 2023-11-28
feat(rbac): add support for creation of role

divyanshiGupta:
Story: https://github.com/janus-idp/backstage-plugins/issues/941, https://github.com/janus-idp/backstage-plugins/issues/942

This PR adds support for creating a role via a create role form. It adds a create button on roles list page which takes the user to the create role form. User can provide name and description and also add  members (users/groups) to the role. On successful creation user is redirected to the roles list page.

Remaining ACs will be covered in this story - https://github.com/janus-idp/backstage-plugins/issues/975

https://github.com/janus-idp/backstage-plugins/assets/20724543/32437302-e95f-4f72-9554-4c5325587a91

Error on submit:
Screenshot 2023-11-30 at 9 15 21 PM

To Do

- [x] Add validation
- [x] Add tests
- [x] Clean up css

### DIFF
--- a/plugins/rbac/dev/index.tsx
+++ b/plugins/rbac/dev/index.tsx
@@ -39,9 +39,11 @@ class MockRBACApi implements RBACAPI {
   constructor(fixtureData: Role[]) {
     this.resources = fixtureData;
   }
+
   async getRoles(): Promise<Role[]> {
     return this.resources;
   }
+
   async getPolicies(): Promise<RoleBasedPolicy[]> {
     return mockPolicies;
   }
@@ -75,6 +77,10 @@ class MockRBACApi implements RBACAPI {
     _policies: Policy[],
   ): Promise<number> {
     return 204;
+  }
+
+  async createRole(_role: Role): Promise<any> {
+    return { status: 200 };
   }
 }
 
--- a/plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
@@ -1,0 +1,116 @@
+import React from 'react';
+import { useAsync } from 'react-use';
+
+import { useApi } from '@backstage/core-plugin-api';
+
+import { LinearProgress, TextField } from '@material-ui/core';
+import Autocomplete from '@material-ui/lab/Autocomplete';
+import FormHelperText from '@mui/material/FormHelperText';
+import { FormikErrors } from 'formik';
+
+import { rbacApiRef } from '../../api/RBACBackendClient';
+import { MemberEntity } from '../../types';
+import {
+  getChildGroupsCount,
+  getMembersCount,
+  getParentGroupsCount,
+} from '../../utils/create-role-utils';
+import { MembersDropdownOption } from './MembersDropdownOption';
+import { CreateRoleFormValues, SelectedMember } from './types';
+
+type AddMembersFormProps = {
+  selectedMembers: SelectedMember[];
+  selectedMembersError?: string;
+  setFieldValue: (
+    field: string,
+    value: any,
+    shouldValidate?: boolean,
+  ) => Promise<FormikErrors<CreateRoleFormValues>> | Promise<void>;
+};
+
+export const AddMembersForm = ({
+  selectedMembers,
+  selectedMembersError,
+  setFieldValue,
+}: AddMembersFormProps) => {
+  const rbacApi = useApi(rbacApiRef);
+  const [search, setSearch] = React.useState<string>('');
+  const {
+    loading: membersLoading,
+    value: members,
+    error,
+  } = useAsync(async () => {
+    return await rbacApi.getMembers();
+  });
+
+  const getDescription = (member: MemberEntity) => {
+    const memberCount = getMembersCount(member);
+    const parentCount = getParentGroupsCount(member);
+    const childCount = getChildGroupsCount(member);
+
+    return member.kind === 'Group'
+      ? `${memberCount} members, ${parentCount} parent group, ${childCount} child groups`
+      : undefined;
+  };
+
+  const membersOptions: SelectedMember[] = members
+    ? members.map((member: MemberEntity, index) => ({
+        label: member.metadata.name,
+        description: getDescription(member),
+        etag:
+          member.metadata.etag ??
+          `${member.metadata.name}-${member.kind}-${index}`,
+        type: member.kind,
+        namespace: member.metadata.namespace,
+        members: getMembersCount(member),
+      }))
+    : [];
+
+  return (
+    <>
+      <FormHelperText>
+        Search and select users and groups to be added. Selected users and
+        groups will appear in the members table.
+      </FormHelperText>
+      <br />
+      <Autocomplete
+        options={membersOptions}
+        getOptionLabel={(option: SelectedMember) => option.label}
+        getOptionSelected={(option: SelectedMember, value: SelectedMember) =>
+          option.etag === value.etag
+        }
+        loading={membersLoading}
+        loadingText={<LinearProgress />}
+        onChange={(_e, value: SelectedMember) =>
+          setFieldValue('selectedMembers', [...selectedMembers, value])
+        }
+        disableClearable
+        getOptionDisabled={(option: SelectedMember) =>
+          !!selectedMembers.find(
+            (sm: SelectedMember) => sm.etag === option.etag,
+          )
+        }
+        renderOption={(option: SelectedMember, state) => (
+          <MembersDropdownOption option={option} state={state} />
+        )}
+        renderInput={params => (
+          <TextField
+            {...params}
+            variant="outlined"
+            label="Users and groups"
+            placeholder="Search by user name or group name"
+            error={!!selectedMembersError}
+            helperText={selectedMembersError ?? ''}
+            value={search}
+            onChange={e => setSearch(e.target.value)}
+            required
+          />
+        )}
+      />
+      <br />
+      {error?.message && (
+        <FormHelperText error={!error}>{error.message}</FormHelperText>
+      )}
+    </>
+  );
+};
--- a/plugins/rbac/src/components/CreateRole/CreateRoleForm.test.tsx
+++ b/plugins/rbac/src/components/CreateRole/CreateRoleForm.test.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+
+import { render, screen } from '@testing-library/react';
+
+import { CreateRoleForm } from './CreateRoleForm';
+
+jest.mock('react-router-dom', () => ({
+  useNavigate: jest.fn(),
+}));
+
+jest.mock('@backstage/core-plugin-api', () => ({
+  ...jest.requireActual('@backstage/core-plugin-api'),
+  useApi: jest.fn(),
+}));
+
+jest.mock('formik', () => ({
+  ...jest.requireActual('formik'),
+  useFormik: jest.fn().mockReturnValue({
+    errors: {},
+    values: {},
+    // mocked useFormik to return formik status with submitError
+    status: { submitError: 'Unexpected error' },
+  }),
+}));
+
+describe('CreateRoleForm', () => {
+  it('renders create role form correctly', async () => {
+    render(<CreateRoleForm />);
+
+    expect(
+      screen.getByText(/enter name and description of role/i),
+    ).toBeInTheDocument();
+    expect(screen.getByTestId(/role-name/i)).toBeInTheDocument();
+    expect(screen.getByTestId(/role-description/i)).toBeInTheDocument();
+    expect(screen.getByText(/add users and groups/i)).toBeInTheDocument();
+  });
+
+  it('shows error if there is any error in formik status', async () => {
+    render(<CreateRoleForm />);
+
+    expect(
+      screen.getByText(/unable to create role. unexpected error/i),
+    ).toBeInTheDocument();
+  });
+});
--- a/plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
+++ b/plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+
+import { Content, Header, Page } from '@backstage/core-components';
+import { catalogEntityReadPermission } from '@backstage/plugin-catalog-common/alpha';
+import { RequirePermission } from '@backstage/plugin-permission-react';
+
+import { CreateRoleForm } from './CreateRoleForm';
+
+export const CreateRolePage = () => {
+  return (
+    <RequirePermission
+      permission={catalogEntityReadPermission}
+      resourceRef={catalogEntityReadPermission.resourceType}
+    >
+      <Page themeId="tool">
+        <Header title="Create role" />
+        <Content>
+          <CreateRoleForm />
+        </Content>
+      </Page>
+    </RequirePermission>
+  );
+};
--- a/plugins/rbac/src/components/CreateRole/RoleDetailsForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/RoleDetailsForm.tsx
@@ -1,0 +1,59 @@
+import React from 'react';
+
+import { TextField } from '@material-ui/core';
+
+type RoleDetailsFormProps = {
+  name: string;
+  description?: string;
+  nameError?: string;
+  handleBlur: React.FocusEventHandler<HTMLInputElement | HTMLTextAreaElement>;
+  handleChange: React.ChangeEventHandler<
+    HTMLTextAreaElement | HTMLInputElement
+  >;
+};
+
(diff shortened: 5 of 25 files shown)


debsmita1:
The create role review is missing

Screenshot 2023-11-29 at 11 46 35 PM



The delete icon is missing from the table

Screenshot 2023-11-29 at 11 46 17 PM



The empty state needs to be fixed

Screenshot 2023-11-29 at 11 46 08 PM


On hitting create and on successful creation user should be taken to the roles list page

debsmita1:
Would you be adding the permission check for creating roles as part of this issue https://github.com/janus-idp/backstage-plugins/issues/975 ?

debsmita1:
The Back and Reset buttons in the Figma are outlined,

divyanshiGupta:
> The Back and Reset buttons in the Figma are outlined,

@debsmita1 the figma is based on new theme but for now I have kept everything consistent with what we already have. @parvathyvr lmk wdyt.

parvathyvr:
> > The Back and Reset buttons in the Figma are outlined,
> 
> @debsmita1 the figma is based on new theme but for now I have kept everything consistent with what we already have. @parvathyvr lmk wdyt.

If we are using the existing theme, i am okay with having the button style in the current implementation.
1 change- Step 2 should be 'Add users ad groups'.

plugins/rbac/dev/index.tsx
```diff
@@ -76,6 +76,8 @@ class MockRBACApi implements RBACAPI {
   ): Promise<number> {
     return 204;
   }
+
+  async createRole(): Promise<any> {}
```
gashcrumb, line 80: Is there a better type than `any` for this?

plugins/rbac/src/components/CreateRole/CreateRoleForm.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    ).toBeInTheDocument();
+    expect(screen.getByTestId(/role-name/i)).toBeInTheDocument();
+    expect(screen.getByTestId(/role-description/i)).toBeInTheDocument();
+    expect(screen.getByText(/add members/i)).toBeInTheDocument();
+  });
+
+  it('handles form submission failure', async () => {
+    render(<CreateRoleForm />);
+
+    expect(screen.getByText(/unable to create role/i)).toBeInTheDocument();
```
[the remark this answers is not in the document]
divyanshiGupta: @gashcrumb On submission if there is any error, formik status gets updated with that error.
```
{formik.status?.submitError && (
      <Box>
          <Alert severity="error">{`Unable to create role. ${formik.status.submitError}`}</Alert>
       </Box>
  )}
```
So here I was just trying to test if any such error is present in the status, the error message is displayed or not. I have mocked formik and returning the status with an error to test the above condition.  I will update the test so that its not confusing.

divyanshiGupta:
@parvathyvr @debsmita1 for now I have added the following permission checks on accessibility of `Create` role button:

- policy.entity.create
- catalog.entity.read
- catalog.entity.create

If the user has all the above permissions then they can go ahead and create the role, otherwise the button is disabled. LMK if  anything else is required here.

divyanshiGupta:
> > @parvathyvr @debsmita1 for now I have added the following permission checks on accessibility of `Create` role button:
> > 
> > * policy.entity.create
> > * catalog.entity.read
> > * catalog.entity.create
> > 
> > If the user has all the above permissions then they can go ahead and create the role, otherwise the button is disabled. LMK if anything else is required here.
> 
> @divyanshiGupta `catalog.entity.create` this won't be necessary, we only want to fetch the users and groups

@debsmita1 I have added `catalog.entity.create` because during role creation we make an api call for creating the role. LMK if we dont need this check for that or `policy.entity.create` is sufficient?

divyanshiGupta:
> > > @parvathyvr @debsmita1 for now I have added the following permission checks on accessibility of `Create` role button:
> > > 
> > > * policy.entity.create
> > > * catalog.entity.read
> > > * catalog.entity.create
> > > 
> > > If the user has all the above permissions then they can go ahead and create the role and policies to it, otherwise the button is disabled. LMK if anything else is required here.
> > 
> > 
> > @divyanshiGupta `catalog.entity.create` this won't be necessary, we only want to fetch the users and groups
> 
> @debsmita1 I have added `catalog.entity.create` because during role creation we make an api call for creating the role as well. LMK if we dont need this check for that or `policy.entity.create` is sufficient?

@PatAKnight wdyt?

divyanshiGupta:
> > > The Back and Reset buttons in the Figma are outlined,
> > 
> > 
> > @debsmita1 the figma is based on new theme but for now I have kept everything consistent with what we already have. @parvathyvr lmk wdyt.
> 
> If we are using the existing theme, i am okay with having the button style in the current implementation. Few things to update -Step 2 should be 'Add users and groups'. -Also i notice we are using 'Create a Role' and 'Create new role'. I think let's go with 'Create role' in the header and in the stepper so that it will be consistent when we have 'Edit role'.

@parvathyvr Done.
Screenshot 2023-12-06 at 7 28 20 PM

plugins/rbac-common/src/permissions.ts
```diff
@@ -67,3 +69,34 @@ export const policyEntityPermissions = [
   policyEntityDeletePermission,
   policyEntityUpdatePermission,
 ];
+
+/**
+ * This permission is used to authorize actions that involve reading
+ * catalog entities.
+ */
+export const catalogEntityReadPermission = createPermission({
```
debsmita1, line 77: you do not need to create this

you can import from the catalog-common plugin

we are doing this in App.tsx
```
import { catalogEntityReadPermission } from '@backstage/plugin-catalog-common/alpha';
```

plugins/rbac-common/src/permissions.ts
```diff
@@ -67,3 +69,34 @@ export const policyEntityPermissions = [
...
+  attributes: {
+    action: 'read',
+  },
+  resourceType: RESOURCE_TYPE_CATALOG_ENTITY,
+});
+
+/**
+ * This permission is used to authorize the creation of new catalog entities.
+ */
+export const catalogEntityCreatePermission = createPermission({
```
debsmita1, line 88: Let's remove this as we do not need create

debsmita1 (changes requested):

debsmita1:
We should disable the route if a user is not authorized to create roles


https://github.com/janus-idp/backstage-plugins/assets/22490998/e263900e-faec-4e54-9159-53350dd0e5d6

plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { CreateRoleForm } from './CreateRoleForm';
+
+export const CreateRolePage = () => {
+  return (
+    <Page themeId="tool">
+      <Header title="Create role" />
+      <Content>
+        <CreateRoleForm />
+      </Content>
+    </Page>
```
debsmita1, line 14:
```suggestion
<RequirePermission
      permission={policyEntityCreatePermission}
      resourceRef={policyEntityCreatePermission.resourceType}
    >
    <Page themeId="tool">
      <Header title="Create role" />
      <Content>
        <CreateRoleForm />
      </Content>
    </Page>
    </RequirePermission>
```

plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  {part.text}
+                </span>
+              ))}
+              <br />
+              <span className={classes.optionDescription}>
+                {option.description}
+              </span>
+            </Box>
+          );
+        }}
```
debsmita1, line 115: Can you move this to a render component?

debsmita1:
I can add a user or group multiple times, and in this scenario, the backend does not generate an error
cc @PatAKnight @AndrienkoAleksandr

divyanshiGupta:
> I can add a user or group multiple times, and in this scenario, the backend does not generate an error cc @PatAKnight @AndrienkoAleksandr

If there is a role with some users/groups and then an api call is made to add same user/group to the role then the backend returns an error but if we add same user/group multiple times in a single api call then the role successfully gets created. To avoid this from UI I have disabled the members that are already selected and can be seen in the table. 

https://github.com/janus-idp/backstage-plugins/assets/20724543/00a8e50e-7673-4a91-824f-74ddb04f1689

cc: @parvathyvr @debsmita1

plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { catalogEntityReadPermission } from '@backstage/plugin-catalog-common/alpha';
+import { RequirePermission } from '@backstage/plugin-permission-react';
+
+import { CreateRoleForm } from './CreateRoleForm';
+
+export const CreateRolePage = () => {
+  return (
+    <RequirePermission
+      permission={catalogEntityReadPermission}
+      resourceRef={catalogEntityReadPermission.resourceType}
```
debsmita1, line 13: Should check for catalogEntityCreatePermission

divyanshiGupta: > > > > @parvathyvr @debsmita1 for now I have added the following permission checks on accessibility of `Create` role button:
> > > > 
> > > > * policy.entity.create
> > > > * catalog.entity.read
> > > > * catalog.entity.create
> > > > 
> > > > If the user has all the above permissions then they can go ahead and create the role and policies to it, otherwise the button is disabled. LMK if anything else is required here.
> > > 
> > > 
> > > @divyanshiGupta `catalog.entity.create` this won't be necessary, we only want to fetch the users and groups
> > 
> > 
> > @debsmita1 I have added `catalog.entity.create` because during role creation we make an api call for creating the role as well. LMK if we dont need this check for that or `policy.entity.create` is sufficient?
> 
> for creating the role we need policy.entity.create

@debsmita1 I am confused. Do we need catalog.entity.create for creating roles or not?

debsmita1: Sorry, my bad
I was looking for policyEntityCreate & catalogEntityRead .Found the 2 places where you have added the permission checks

plugins/rbac/src/components/RolesList.tsx
```diff
@@ -61,6 +66,17 @@ export const RolesList = () => {
           {toastMessage}
         </Alert>
       </Snackbar>
+      <span className={classes.createRoleButton}>
```
gashcrumb, line 69: Frustrating that the backstage components seem to avoid having a `className` property.  Anyway, it's really only a nit but I suggest maybe call this class `createRoleButtonWrapper` just so it's more clear earlier in the file where this classname is defined that it doesn't target the button but instead a wrapper component that's used to align the button.

plugins/rbac/src/components/CreateRole/RoleDetailsForm.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+
+import { TextField } from '@material-ui/core';
+import { FormikProps } from 'formik';
+
+import { CreateRoleFormValues } from './types';
+
+export const RoleDetailsForm = ({
+  formik,
```
gashcrumb, line 9: Here is another example where I would suggest not accepting the formik instance but instead accept the specific properties.   Let the parent component deal with the formik instance, this component then becomes much simpler and potentially usable with other form libraries even.

plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
+import { useAsync } from 'react-use';
+
+import { useApi } from '@backstage/core-plugin-api';
+
+import { LinearProgress, TextField } from '@material-ui/core';
```
gashcrumb, line 6: I mentioned this on @debsmita1's PR but probably good to mention it here too.  I think we should try and use Material 5 whenever possible, so `@mui/material/LinearProgress` , `@mui/material/TextField` and so on.  It could be too late to do this for this particular iteration and that's fine, we'll have to do it eventually when upstream drops Material 4.

invincibleJai:
@divyanshiGupta dev mode doesn't seem to work

image

invincibleJai:
@divyanshiGupta Thanks for the work, overall looks good. I have one observation

Create btn is disabled, and no info is shown for the reason though the user has create access to the policy entity 

image
 
I believe that is because the logged-in user didn't have create permission to `catalog-entity` so README should cover that or a tooltip

plugins/rbac/src/components/CreateRole/RoleDetailsForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+        value={name}
+        name="name"
+        onChange={handleChange}
+        onBlur={handleBlur}
+        error={!!nameError}
+      />
+      <TextField
+        label="Description"
+        variant="outlined"
+        helperText="Enter a brief description about the role(The purpose of the role)"
```
invincibleJai, line 47: Nit: this text is per UXD?

```suggestion
        helperText="Enter a brief description about the role (The purpose of the role)"
```

divyanshiGupta: @invincibleJai Yes it is as per https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=932-1542&mode=design&t=V40DvGtSyENzJFNB-4.

invincibleJai: so there is space :)

invincibleJai (approved):
LGTM

plugins/rbac/src/components/CreateRole/RoleDetailsForm.tsx
divyanshiGupta: yes, added the space.

divyanshiGupta:
> @divyanshiGupta Thanks for the work, overall looks good. I have one observation
> 
> Create btn is disabled, and no info is shown for the reason though the user has create access to the policy entity
> 
> image
> I believe that is because the logged-in user didn't have create permission to `catalog-entity` so README should cover that or a tooltip

@invincibleJai yes it need `policy.entity.create` and `catalog.entity.read` permissions. Good idea to add in README.

gashcrumb (approved):
:+1:

parvathyvr:
> @divyanshiGupta Thanks for the work, overall looks good. I have one observation
> 
> Create btn is disabled, and no info is shown for the reason though the user has create access to the policy entity
> image
> 
> I believe that is because the logged-in user didn't have create permission to `catalog-entity` so README should cover that or a tooltip

Hi @divyanshiGupta I have added an inline alert scenario if 'Create' button is disabled- https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=1684-572&mode=design&t=4ygFyKpEzJg5P0h6-4
(The alert message is for a different scenario- but i assume we can use this pattern to display to the user as to why 'Create' is disabled in the UI).

invincibleJai:
> > @divyanshiGupta Thanks for the work, overall looks good. I have one observation
> > Create btn is disabled, and no info is shown for the reason though the user has create access to the policy entity
> > image
> > I believe that is because the logged-in user didn't have create permission to `catalog-entity` so README should cover that or a tooltip
> 
> Hi @divyanshiGupta I have added an inline alert scenario if 'Create' button is disabled- https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=1684-572&mode=design&t=4ygFyKpEzJg5P0h6-4 (The alert message is for a different scenario- but i assume we can use this pattern to display to the user as to why 'Create' is disabled in the UI).

@parvathyvr thanks 👍 

@divyanshiGupta would you like to take this as follow up i.e as part of your final create flow PR?

divyanshiGupta:
> > > @divyanshiGupta Thanks for the work, overall looks good. I have one observation
> > > Create btn is disabled, and no info is shown for the reason though the user has create access to the policy entity
> > > image
> > > I believe that is because the logged-in user didn't have create permission to `catalog-entity` so README should cover that or a tooltip
> > 
> > 
> > Hi @divyanshiGupta I have added an inline alert scenario if 'Create' button is disabled- https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=1684-572&mode=design&t=4ygFyKpEzJg5P0h6-4 (The alert message is for a different scenario- but i assume we can use this pattern to display to the user as to why 'Create' is disabled in the UI).
> 
> @parvathyvr thanks 👍
> 
> @divyanshiGupta would you like to take this as follow up i.e as part of your final create flow PR?

@invincibleJai sure  I will address this in my next PR along with updating the README (based on what we decide if the change should be in rbac backend plugin or UI).

debsmita1:
LGTM

Issue #776, opened 2023-09-21
Tekton Plugin is importing a wrong Kubernetes version in case of Backstage upgrade


gsasanelli:
### Describe the bug

While upgrading backstage to 1.18.x version I got the error
"Error: Version bump failed for @backstage/plugin-kubernetes"
during the backstage-cli versions:bump command

Checking at the backstage-cli info command I found this:
@backstage/plugin-kubernetes 0.10.3, 0.9.3

And then checking your dependencies here:
https://github.com/janus-idp/backstage-plugins/blob/main/plugins/tekton/package.json

I found that you are using the 0.9.3 plugin version that now is upgraded at the version 0.10.3. This is blocking the upgrade process

### Expected Behavior
Upgrade should work without any problem

### What are the steps to reproduce this bug?
1. Have a Backstage installation (version less than 1.18.x) with Kubernetes and Tekton plugin installed
2. Try to upgrade using the command yarn backstage-cli versions:bump

### Versions of software used and environment
I'm upgrading Backstage from version 1.16 to 1.18.1

PR #994, opened 2023-12-06
[WIP] fix(rbac): CSV updates no longer require server restarts

### LINKED ISSUE #908: RBAC backend plugin - policy change requires server restart
### Describe the bug

When we specify a politcy (using the DB, or the CSV file)
the policy makes the job it was expected to
but when changing the policy, the new changes do not get applied unless the entire project is restarted

### Expected Behavior
expected to have the new plicy changes applied directly and not requiring restart of the project it self but dunamiclly loaded on the fly

1. configure the rbac plugin
2. configure a csv policy with deny for read and write
3. start the project
4. change the policy to allow
5. check the ui, it will be still on deny

### Versions of software used and environment

PatAKnight:
## Description

This PR introduces new functionality to facilitate the reloading of permission policies sourced from the CSV file. The issue at hand was the necessity of restarting the server every time there was an update to the permission policies in the CSV, which was quite inconvenient. The added feature enables the dynamic reloading of permission policies and roles directly from the CSV.

It leverages information provided by the permission framework (entity reference, permission, policy). This approach allows us to focus solely on roles associated with the entity reference in the CSV, subsequently retrieving permissions linked to that role, matching the given permission and policy. The aim is to enhance efficiency by selectively reloading relevant portions of the file, rather than the entire dataset.

## Fixes

- Fixes #908 

1. Create an RBAC permission policy CSV file
```csv
g, user:default/<your-user>, role:default/test

p, role:default/test, catalog-entity, read, allow
```
2. Update app-config to include your CSV file
```yaml
permission:
  enabled: true
  rbac:
    policies-csv-file: ./some/path/<permission-policy>.csv
```
3. Start the server and notice that you can read catalog entities
4. update the read permission in the CSV file from `allow` to `deny`
5. Notice that you can no longer read the catalog entities

***

The bulk of this work is finished. Nevertheless, a lingering issue with duplication still requires attention. If a permission policy initially comes from the REST API and is subsequently added in the CSV file, it triggers a problem. We anticipate that resolving [issue #776](https://github.com/janus-idp/backstage-showcase/issues/776) will give us a mechanism to determine the origin of a permission policy, whether it stems from the CSV file or the REST API.

Also, I would like to test how this works with bulk adding and editing using Casbin and the REST API. Currently, the REST API allows for the creation of roles, and will allow for the creation of permissions soon, in bulk. I want to test this after the mention issue is solved to ensure that there are no problems.

***

### DIFF
--- a/plugins/rbac-backend/src/helpers/csv.ts
+++ b/plugins/rbac-backend/src/helpers/csv.ts
@@ -1,0 +1,200 @@
+import { Enforcer, FileAdapter, newEnforcer, newModelFromString } from 'casbin';
+import { isEqual } from 'lodash';
+
+import { MODEL } from '../service/permission-model';
+import {
+  validateEntityReference,
+  validatePolicy,
+} from '../service/policies-validation';
+import {
+  checkForDuplicateGroupPolicies,
+  checkForDuplicatePolicies,
+  transformArraytoPolicy,
+  updatePolicies,
+} from './utils';
+
+export const loadFilteredPoliciesFromCSV = async (
+  policyFile: string,
+  enf: Enforcer,
+  entityRef: string,
+  resourceType: string,
+  action: string,
+) => {
+  const fileEnf = await newEnforcer(
+    newModelFromString(MODEL),
+    new FileAdapter(policyFile),
+  );
+
+  const roles = await enf.getFilteredGroupingPolicy(0, entityRef);
+
+  const policies: string[][] = [];
+  const enforcerPolicies: string[][] = [];
+
+  for (const role of roles) {
+    const policyByRole = await fileEnf.getFilteredPolicy(
+      0,
+      role[1],
+      resourceType,
+      action,
+    );
+    policies.push(...policyByRole);
+    const enforcerPolicy = await enf.getFilteredPolicy(
+      0,
+      role[1],
+      resourceType,
+      action,
+    );
+    enforcerPolicies.push(...enforcerPolicy);
+  }
+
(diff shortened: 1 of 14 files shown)


plugins/rbac-backend/src/helpers/csv.ts
```diff
@@ -0,0 +1,200 @@
...
+} from './utils';
+
+export const loadFilteredPoliciesFromCSV = async (
+  policyFile: string,
+  enf: Enforcer,
+  entityRef: string,
+  resourceType: string,
+  action: string,
+) => {
+  const fileEnf = await newEnforcer(
```
AndrienkoAleksandr, line 23: @PatAKnight, maybe we can reuse the same fileAdapter for both methods loadFilteredPoliciesFromCSV and loadFilteredGroupingPoliciesFromCSV to make some performance optimisation?

plugins/rbac-backend/src/helpers/csv.ts
```diff
@@ -0,0 +1,200 @@
...
+  if (policies.length === 0) {
+    policies.push(
+      ...(await fileEnf.getFilteredPolicy(1, resourceType, action)),
+    );
+  }
+
+  for (const policy of policies) {
+    const err = validatePolicy(transformArraytoPolicy(policy));
+    if (err) {
+      throw new Error(
```
AndrienkoAleksandr, line 63: I think we should throw warning at least for now, because broken policies file will make each permission validation request not working. Also we can implement warning events for frontend plugin in the feature.

PatAKnight:
Closing in favor of #1171

Issue #992, opened 2023-12-06
[RBAC Plugin] Add Edit Icon to the Users and groups Card on the Role Overview Page


debsmita1:
- Add Edit Icon to the roles on the Roles List Page
  - Disable the Edit icon if the user has no access to update roles
  - Clicking on the Edit icon should take the user to the Create/Edit Role Form with all the form data populated
- Add Edit Icon to the Users and groups Card on the Role Overview Page
  - Disable the Edit icon if the user has no access to update roles
  - Clicking on the Edit icon should take the user to the Create/Edit Role Form with the `Add members` section open
 - On hitting `Save` , users should be redirected to the roles list page and shown a toast saying the role updation is successful

PR #1001, opened 2023-12-06
feat(rbac): allow editing roles

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/992

**GIF:**

https://github.com/janus-idp/backstage-plugins/assets/22490998/3414c4f6-f564-4f24-b984-9003a3c19c3c

When a user is not authorized to update the roles
Screenshot 2023-12-11 at 7 52 29 PM
Screenshot 2023-12-11 at 7 45 43 PM

Updated:
Screenshot 2023-12-11 at 7 39 30 PM


plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+
+import { stringifyEntityRef } from '@backstage/catalog-model';
+
+import { Box, LinearProgress, makeStyles, TextField } from '@material-ui/core';
```
gashcrumb, line 5: Would it be worth favoring material 5 components at this stage?  Or we can wait until it's mandatory at some point in the future.

parvathyvr:
@debsmita1 some updates:
1. For the 'Edit role' can we update the step names with 'Edit' instead of 'Add'- I have updated that in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=932-4240&mode=design&t=Oig8tD3k17qFGsKJ-4).
2. About the scenario 'When a user is not authorized to update the roles' -for the disabled edit icon tooltip- i wonder if we could specify 'Unauthorized to edit' or something in those lines instead of 'Role cannot be updated'- so that user is aware of the role why it is disabled for them. WDYT? 
This goes in line with @ShiranHi  [comment here](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s?node-id=922:6877&mode=design#638922317) about improving the tooltip messages we show.

plugins/rbac/src/utils/create-role-utils.test.ts
```diff
@@ -1,27 +1,69 @@
...
+          label: 'group1',
+          etag: '2',
+          ref: 'group:default/group1',
+        },
+      ],
+    };
+
+    result = getRoleData(values);
+
+    expect(result).toEqual({
```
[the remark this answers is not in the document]
debsmita1: not much, testing how the role names will differ based on the data

plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
debsmita1: This is part of Divyanshi's Create Role PR. What do you suggest @divyanshiGupta & @parvathyvr  ?

debsmita1:
> @debsmita1 some updates:
> 
> 1. For the 'Edit role' can we update the step names with 'Edit' instead of 'Add'- I have updated that in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=932-4240&mode=design&t=Oig8tD3k17qFGsKJ-4).
> 2. About the scenario 'When a user is not authorized to update the roles' -for the disabled edit icon tooltip- i wonder if we could specify 'Unauthorized to edit' or something in those lines instead of 'Role cannot be updated'- so that user is aware of the role why it is disabled for them. WDYT?
>    This goes in line with @ShiranHi  [comment here](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s?node-id=922:6877&mode=design#638922317) about improving the tooltip messages we show.

Screenshot 2023-12-11 at 7 39 30 PM

parvathyvr:
> > @debsmita1 some updates:
> > 
> > 1. For the 'Edit role' can we update the step names with 'Edit' instead of 'Add'- I have updated that in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=932-4240&mode=design&t=Oig8tD3k17qFGsKJ-4).
> > 2. About the scenario 'When a user is not authorized to update the roles' -for the disabled edit icon tooltip- i wonder if we could specify 'Unauthorized to edit' or something in those lines instead of 'Role cannot be updated'- so that user is aware of the role why it is disabled for them. WDYT?
> >    This goes in line with @ShiranHi  [comment here](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s?node-id=922:6877&mode=design#638922317) about improving the tooltip messages we show.
> 
> Screenshot 2023-12-11 at 7 39 30 PM
> 
> Screenshot 2023-12-11 at 7 45 43 PM

LGTM! Thanks @debsmita1

plugins/rbac/dev/index.tsx
```diff
@@ -48,15 +48,37 @@ class MockRBACApi implements RBACAPI {
...
+  }
+
   async getUserAuthorization(): Promise<{ status: string }> {
     return {
       status: 'Authorized',
     };
   }
 
-  async getRole(role: string): Promise<Role[]> {
+  async getRole(role: string): Promise<any> {
```
invincibleJai, line 66: nit: type

plugins/rbac/src/components/CreateRole/CreateRoleForm.tsx
```diff
@@ -17,35 +17,70 @@ import { Alert } from '@material-ui/lab';
...
 import { RoleDetailsForm } from './RoleDetailsForm';
-import { CreateRoleFormValues } from './types';
+import { CreateRoleFormValues, SelectedMember } from './types';
 
-export const CreateRoleForm = () => {
-  const [activeStep, setActiveStep] = React.useState<number>(0);
+type CreateRoleFormProps = {
+  roleName: string;
+  step?: number;
+  members: { members: MemberEntity[]; loading: boolean; error: Error };
```
invincibleJai, line 32: nit: naming suggestion

```suggestion
  membersData: { members: MemberEntity[]; loading: boolean; error: Error };
```

plugins/rbac/src/components/CreateRole/CreateRoleForm.tsx
```diff
@@ -17,35 +17,70 @@ import { Alert } from '@material-ui/lab';
...
 import { AddedMembersTable } from './AddedMembersTable';
 import { AddMembersForm } from './AddMembersForm';
 import { RoleDetailsForm } from './RoleDetailsForm';
-import { CreateRoleFormValues } from './types';
+import { CreateRoleFormValues, SelectedMember } from './types';
 
-export const CreateRoleForm = () => {
-  const [activeStep, setActiveStep] = React.useState<number>(0);
+type CreateRoleFormProps = {
+  roleName: string;
```
invincibleJai, line 30: should this be optional as well?

plugins/rbac/src/hooks/useSelectedMembers.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  if (!roleName) {
+    return {
+      members: members as MemberEntity[],
+      error: error as Error,
+      loading: membersLoading,
+    };
+  }
+
+  data = Array.isArray(role)
```
invincibleJai, line 44: role loading/err state is missed

plugins/rbac/src/components/CreateRole/CreateRoleForm.tsx
debsmita1: passing empty string if the rolename is not present in the query params

plugins/rbac/dev/index.tsx
```diff
@@ -48,22 +48,44 @@ class MockRBACApi implements RBACAPI {
     return mockPolicies;
   }
 
+  async getAssociatedPolicies(
+    entityReference: string,
+  ): Promise<RoleBasedPolicy[]> {
+    return (
+      mockPolicies.filter(pol => pol.entityReference === entityReference) ||
```
gashcrumb, line 55: Doesn't the `filter` function return an empty array if there's no matches?  I think then the `||` here isn't needed.

plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
```diff
@@ -15,7 +39,25 @@ export const CreateRolePage = () => {
...
+          <RoleForm
+            initialValues={initialValues}
+            titles={{
+              formTitle: 'Create Role',
+              nameAndDescriptionTitle: 'Enter name and description of role ',
+              usersAndGroupsTitle: 'Add users and groups',
+              permissionPoliciesTitle: '',
+            }}
+            membersData={{
+              members: Array.isArray(members)
```
gashcrumb, line 51: So when `members` isn't an array, what can it be?  A `Response` object that might contain an error from the API?

plugins/rbac/src/components/CreateRole/RoleForm.tsx
```diff
@@ -17,35 +17,71 @@ import { Alert } from '@material-ui/lab';
...
+      values: RoleFormValues,
+      formikHelpers: FormikHelpers<RoleFormValues>,
     ) => {
       try {
-        const data = getRoleData(values);
-        const res = await rbacApi.createRole(data);
+        const newData = getRoleData(values);
+        const oldData = getRoleData(initialValues);
+
+        let res: Response;
```
gashcrumb, line 67: I wonder if this block of code dealing with the `Response` directly should be moved to the API layer.  Usually it's best to avoid having to deal with this at the UI component level.  

As an example I'm a big fan of the approach that was taken [here](https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/blob/main/src/apis/api.ts#L316-L366) when this was first developed, success/error callbacks are provided to the API function and those API call details are kept away from the UI component.  The caller instead provides the token and details to access the server and gets back a function that provides onSuccess and onError callbacks, one of these are called depending on the result of the remote call.  Not saying we adopt this approach now or at all, just an example of what I've seen work well.

gashcrumb (approved):
just a couple questions that aren't super-critical, overall :+1:

plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
debsmita1: yes a Response object, and the object will have the status code and the status message

plugins/rbac/src/components/CreateRole/RoleForm.tsx
debsmita1: Thank you for pointing this out, I handled this in the other API functions but overlooked it here. I couldn't create scenarios where the API call would throw an error. Even with an incorrect bearer token or query parameters, it would still enter the "then" block. I could only display the error to the user by reading the status code in the response. PTAL at the updated API functions.

gashcrumb: Ah yes, with the `fetch` API it doesn't throw a TypeError for some HTTP error return codes.  You instead have to handle it directly like:

```
fetch(someUrl).then(( response:Response ) => {
  if ( !response.ok ) {
    throw new Error(response.message); // or maybe a more informative string based on Response.status
  }
  // happy path here
}).catch(( err:Error ) => {
  console.log("Something bad happened: ", err);
});
```

That snippet I sent along used axios under the covers which does this already and hides the above.

plugins/rbac/src/api/RBACBackendClient.ts
```diff
@@ -162,5 +188,63 @@ export class RBACBackendClient implements RBACAPI {
...
+        method: 'PUT',
+        headers: {
+          'Content-Type': 'application/json',
+          Accept: 'application/json',
+          ...(idToken && { Authorization: `Bearer ${idToken}` }),
+        },
+        body: JSON.stringify(body),
+      },
+    );
+    if (jsonResponse.status !== 200 && jsonResponse.status !== 201) {
```
gashcrumb, line 217: Cool, this is a variation on the code snippet I mentioned in another comment @debsmita1 and should work fine.

plugins/rbac/src/components/CreateRole/RoleForm.tsx
gashcrumb: Another option with this section is maybe try and first just move the logic to a separate function that accepts `onSuccess` & `onError` callback functions.  Instead of returning a value or throwing an exception the function would instead either call `onSuccess` and supply the `roleName` as an argument or `onError` and supply the remote error as an argument.  This works nicely especially when there's multiple conditions to deal with like you have here.

plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
```diff
@@ -1,12 +1,36 @@
 import React from 'react';
+import { useAsync } from 'react-use';
 
 import { Content, Header, Page } from '@backstage/core-components';
+import { useApi } from '@backstage/core-plugin-api';
 import { catalogEntityReadPermission } from '@backstage/plugin-catalog-common/alpha';
 import { RequirePermission } from '@backstage/plugin-permission-react';
 
-import { CreateRoleForm } from './CreateRoleForm';
+import { get } from 'lodash';
```
invincibleJai, line 9: nit: we can avoid this

plugins/rbac/src/components/Router.tsx
```diff
@@ -3,31 +3,49 @@ import { Route, Routes } from 'react-router-dom';
...
-          resourceRef={policyEntityCreatePermission.resourceType}
-        >
-          <CreateRolePage />
-        </RequirePermission>
-      }
-    />
-  </Routes>
+  <ToastContextProvider>
+    <Routes>
+      <Route path="*" element={<RbacPage />} />
```
invincibleJai, line 25: why do we need `*`?

plugins/rbac/src/components/RolesList.tsx
```diff
@@ -30,6 +30,12 @@ export const RolesList = () => {
   const classes = useStyles();
   const { loading, data, retry, createRoleAllowed } = useRoles();
 
+  React.useEffect(() => {
+    return () => {
+      setToastMessage('');
+    };
+  }, [setToastMessage]);
```
invincibleJai, line 37: this doesn't seem right, toast ideally should be across the app if there is something at App level or at at plugin level.

This we can take as follow issue if needed

plugins/rbac/src/hooks/useSelectedMembers.test.ts
```diff
@@ -0,0 +1,232 @@
...
+          'user:default/amelia.park',
+          'user:default/calum.leavy',
+          'group:default/team-b',
+          'group:default/team-c',
+        ],
+        name: 'role:default/rbac_admin',
+      },
+    ]),
+    getMembers: jest.fn().mockReturnValue([
+      {
```
invincibleJai, line 22: nit: good to move the test data in `fixtures` so that can be used across if needed

plugins/rbac/src/hooks/useSelectedMembers.ts
```diff
@@ -0,0 +1,73 @@
+import { useAsync } from 'react-use';
+
+import { stringifyEntityRef } from '@backstage/catalog-model';
+import { useApi } from '@backstage/core-plugin-api';
+
+import { get } from 'lodash';
```
invincibleJai, line 6: nit: avoid lodash

plugins/rbac/src/components/RolesList.tsx
debsmita1: Will look into this as part of a separate ticket

invincibleJai:
@debsmita1 the commit needs to be cleaned

invincibleJai (approved):
LGTM

Issue #878, opened 2023-10-25
[Tekton Plugin] Download TaskRun logs


debsmita1:
### What do you want to improve?
There is a provision to download pod logs in the Topology plugin. We want to do the same for the Tekton Task logs

### What is the current behavior?
As of today, there is no way to download the Task logs.

Tekton Plugin Task Logs:
(https://github.com/janus-idp/backstage-plugins/pull/553)
<img width="1710" alt="Screenshot 2023-10-25 at 5 09 39 PM" src="https://github.com/janus-idp/backstage-plugins/assets/22490998/49d884b5-7c6b-4a3a-81dc-b97bf8a35799">

Topology Plugin Pods Logs:
(https://github.com/janus-idp/backstage-plugins/pull/616)
<img width="1710" alt="Screenshot 2023-10-25 at 5 08 22 PM" src="https://github.com/janus-idp/backstage-plugins/assets/22490998/a91884b7-0172-4155-9c39-ad273df0495b">

Openshift:
<img width="1240" alt="Screenshot 2023-10-25 at 6 03 07 PM" src="https://github.com/janus-idp/backstage-plugins/assets/22490998/e9b2299c-0452-4c88-9125-6cddb08b8b22">

### What will the new behavior be?
Add a provision to download the Task logs

PR #1014, opened 2023-12-08
feat(tekton): add support for downloading task and pipelinerun logs

karthikjeeyar:
**Fixes:**
https://github.com/janus-idp/backstage-plugins/issues/878

**Description:**

Add download actions to download task and pipelinerun logs.

**Screenshots:**

https://github.com/janus-idp/backstage-plugins/assets/9964343/d7f5c36c-d447-4de7-b213-049a5882905b

Screenshot 2023-12-08 at 5 54 03 PM

**Steps to test:**

 - Click on task node to open the logs modal.

**Unit tests:**

```
   getPodLogs
    ✓ should return empty logs if there are no pods (12 ms)
    ✓ should return logs if there are pods (2 ms)
    ✓ should display logs only for the pods that has logs (1 ms)
  
  PodLogsDownloadLink
    ✓ should download the pipelineRun logs (74 ms)
    ✓ should disable download button during download and enable it back the process is complete (30 ms)
    
  PipelineRunLogDownloader
     ✓ should not show download links (49 ms)
    ✓ should return download links (95 ms)
    
   PipelineRunLogDialog
    ✓ should not show pipeline run logs modal (130 ms)
    ✓ should show pipeline run logs modal (254 ms)
    ✓ should not show download links in the logs modal if there are no pods (67 ms)
    ✓ should show download links in the logs modal if pods are available (67 ms)
    
 ```

### CODE REVIEW

**Comment by debsmita1** [comment]
@karthikjeeyar could you resolve the code smells ?

**Comment by invincibleJai** [comment]
Thanks @karthikjeeyar for working on this. Overall looks good. Can you add tests for components as well?

**Comment by invincibleJai** [comment]
@karthikjeeyar While click on download i see the file name by default is `undefined` Can we use the task name ?

image

File: plugins/tekton/src/utils/pods-data.ts [hunk 1]
```diff
@@ -0,0 +1,116 @@
+import { V1Pod } from '@kubernetes/client-node';
```
gashcrumb, line 1: Thinking about this further after the first look, I wonder if this should go in a test file or more clearly indicate that this is test stuff that shouldn't get built into the actual plugin code.  I'm sure that's not happening now, but you never know, that can be changed down the line accidentally.

gashcrumb:
Hmm, I'm not seeing this last column when testing this locally but I do see these download links that are disabled, are there any additional required labels for this link to work?  Also, the disabled links still seem to do something, so that probably needs to be fixed, for example clicking on the "Download" link in dev mode reveals this error:

image

plugins/tekton/src/utils/pods-data.ts
karthikjeeyar: moved this test data file to `__fixtures__` folder to indicate this is test related file.

karthikjeeyar:
> While click on download i see the file name by default is undefined Can we use the task name?

Task name & pipelinerun name is used for `Download` and `Download all task logs` button respectively. 

In this case it seems like there isn't any valid pods available to fetch the logs, so I have added check to not show the download buttons if there are no pods and added more tests.

invincibleJai:
Thanks @karthikjeeyar works as expected. In the dev setup I observed option to download is not shown i blv that is because we would need to mock the proxy response for logs. It'll be nice if we can get this.

karthikjeeyar:
@invincibleJai I have mocked the`KubernetesProxyApi.getPodLogs` method in dev mode, now logs will be displayed and download buttons is functional in dev mode as well.


https://github.com/janus-idp/backstage-plugins/assets/9964343/c0154213-0787-468b-a6f2-02f40f3cb4ce

invincibleJai (approved):
LGTM

Issue #975, opened 2023-11-28
[RBAC plugin] Add support for adding permissions to the role


divyanshiGupta:
- Add support for adding permissions to the role via create role form implemented in https://github.com/janus-idp/backstage-plugins/issues/941

- It should also have review section as shown here https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=891-1186&mode=design&t=hKziidGVofZ3CnsT-0

- Show breadcrumbs on form page

- Show notification on successful creation of role on roles list page

PR #1021, opened 2023-12-14
feat(rbac): support for adding permission policies to roles

divyanshiGupta:
**Story:** https://github.com/janus-idp/backstage-plugins/issues/975

**Demo:**

https://github.com/janus-idp/backstage-plugins/assets/20724543/2bbcff8b-2eec-4e13-9773-6dc2e8fce6e6

ToDo:

- [x] Add tests
- [x] Add validations
- [x] Clean up


plugins/rbac/src/components/CreateRole/RoleForm.tsx
```diff
@@ -81,6 +83,18 @@ export const RoleForm = ({
...
+                if (policy.checked) {
+                  await rbacApi.createPolicy({
+                    entityReference: `role:${formik.values.namespace}/${formik.values.name}`,
+                    permission: `${pp.permission}`,
+                    policy: policy.label.toLowerCase(),
+                    effect: 'allow',
+                  });
+                }
+              }),
+            );
```
debsmita1, line 97: We should notify the user if there's any error while creating the policies

divyanshiGupta: @debsmita1 yes. I am adding validations.

divyanshiGupta:
cc: @parvathyvr

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -5,8 +5,11 @@ import {
   stringifyEntityRef,
 } from '@backstage/catalog-model';
 
+import _ from 'lodash';
```
gashcrumb, line 8: I think it's better to import just the one function you're using.  And I dunno, maybe this dependency isn't needed, here's [something to consider](https://youmightnotneed.com/lodash#uniqBy).

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -168,3 +171,45 @@ export const getSelectedMember = (
   }
   return {} as SelectedMember;
 };
+
+export const getPermissionPolicies = (policies: Policy[]) => {
```
debsmita1, line 175: Add a return type for this util

debsmita1:
- Clear the Permission dropdown when the plugin value is changed
- I can proceed without selecting a policy for permission. Do we want to allow that?

https://github.com/janus-idp/backstage-plugins/assets/22490998/6de03cd2-75af-4b01-ab16-30f69585237d

parvathyvr:
Hi @divyanshiGupta ,
Few points i noticed in the Review and create’ section that needs update. [Figma design to refer](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=912-6573&mode=design&t=0gSOnQGdH5Uxflcc-4)
1. Let’s use sentence case in the ‘Review and Create’ step as well- eg: ‘Name and description of the role”
2. Show the number of ‘Users and groups’ and the ‘Permission policies’.
3. Can we show the policies as comma separated?
4. Can we have a bit of space after the table heading?
5. The empty space when no rows are present- can we remove it?
Attaching a screenshot with the above points:
Review and create changes
Let me know if there are any questions-thank you!

plugins/rbac/src/components/CreateRole/SelectedPermissionPoliciesColumn.tsx
```diff
@@ -0,0 +1,22 @@
...
+    title: 'Policies',
+    field: 'policies',
+    render: (policies: RowPolicy[]) => {
+      return policies.reduce((acc: string, p) => {
+        if (p.checked) return acc.concat(`${p.label}  `);
+        return acc;
+      }, '');
+    },
+  },
+];
```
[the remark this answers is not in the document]
divyanshiGupta: @debsmita1 `Policies` field and render methods are different for both the components and the table data type is similar but not same. Also they for different tables, one is a backstage table and one is core html table. Backstage table types should not be used with a basic html table. Lets keep them separated for now.

parvathyvr:
Hi @divyanshiGupta some more points that came up when i reviewed along with @ShiranHi are:
1) Can we add a tooltip for the '-' icon in the 'Add permission policies' step? Can we show 'Remove' as a tooltip message when the user hovers over it?
2) About the blue banner we show to notify the user after the role creation-Can we make it center aligned? 
3) About the 'Add users and groups' step- ( i missed these points in the earlier PR :/)
   3.1 Can we make the count we show (1 Group, 2 Users) to small letters eg:(1 group, 2 users)? @debsmita1 this change is  applicable to the overview table as well.
   3.2 The users and groups in the table view are clickable as shown in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=912-5706&mode=design&t=bZ3ALoZVbTpDPjs3-4)- the idea was if user clicks on them we show the overview page of the respective user/ group in a new tab.
Thank you!

divyanshiGupta:
> Hi @divyanshiGupta some more points that came up when i reviewed along with @ShiranHi are:
> 
> 1. Can we add a tooltip for the '-' icon in the 'Add permission policies' step? Can we show 'Remove' as a tooltip message when the user hovers over it?
> 2. About the blue banner we show to notify the user after the role creation-Can we make it center aligned?
> 3. About the 'Add users and groups' step- ( i missed these points in the earlier PR :/)
>    3.1 Can we make the count we show (1 Group, 2 Users) to small letters eg:(1 group, 2 users)? @debsmita1 this change is  applicable to the overview table as well.
>    3.2 The users and groups in the table view are clickable as shown in the [design](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=912-5706&mode=design&t=bZ3ALoZVbTpDPjs3-4)- the idea was if user clicks on them we show the overview page of the respective user/ group in a new tab.
>    Thank you!

@parvathyvr The first point is already there
Screenshot 2023-12-19 at 4 18 54 PM

Also in 3.2 by overview page you mean below ?
Screenshot 2023-12-19 at 4 22 22 PM

parvathyvr:
Hi @divyanshiGupta About your earlier comments-

> The first point is already there
Awesome! I missed it.

> Also in 3.2 by overview page you mean below ?
That's correct.

debsmita1:
unable to scroll horizontally

Screenshot 2023-12-19 at 11 24 28 PM

divyanshiGupta:
> Hi @divyanshiGupta , Few points i noticed in the Review and create’ section that needs update. [Figma design to refer](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=912-6573&mode=design&t=0gSOnQGdH5Uxflcc-4)
> 
> 1. Let’s use sentence case in the ‘Review and Create’ step as well- eg: ‘Name and description of the role”
> 2. Show the number of ‘Users and groups’ and the ‘Permission policies’.
> 3. Can we show the policies as comma separated?
> 4. Can we have a bit of space after the table heading?
> 5. The empty space when no rows are present- can we remove it?
>    Attaching a screenshot with the above points:
> 
> Review and create changes Let me know if there are any questions-thank you!

@parvathyvr I am using backstage's `StructuredMetadataTable` which consume the table data as key/value pairs. Its implemented in a way that the key will be rendered in a specific way only and we cannot change the case as mentioned in your point 1 and for for 2 point we cannot use any character apart alpha-numeric as it looks like they are using a similar reg-ex for generating key name and removing and special characters so brackets are not getting added. Should we show just the  numbers?

debsmita1:
@divyanshiGupta Breadcrumbs are missing

divyanshiGupta:
@debsmita1 breadcrumbs are there

divyanshiGupta:
> @divyanshiGupta
> 
> * Breadcrumbs are missing
> 
> Screenshot 2023-12-20 at 7 49 02 PM - The Permission dropdown displays the validation error even before I touch the field

@debsmita1 yeah but that is because on plugin change I have to update the permission dropdown to show initial value so that the user can choose from the the new permissions and that is why the validation is run. Anyways the permission field is a required field so for not it should be ok. We can have another PR to fix this.

divyanshiGupta:
@parvathyvr I have resolved the comments. This is how it looks:

https://github.com/janus-idp/backstage-plugins/assets/20724543/9e957623-e369-4b39-a9ea-edd842e728ee

We can have another PR for re-aligning toast as it was not added in this.

Apart from that this is how I am handling any errors that occur while creating permissions:
Screenshot 2023-12-20 at 8 22 44 PM

If you have another idea on how this should be handles lmk and we can have a separate PR for that. Thanks!

parvathyvr:
> > Hi @divyanshiGupta , Few points i noticed in the Review and create’ section that needs update. [Figma design to refer](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=912-6573&mode=design&t=0gSOnQGdH5Uxflcc-4)
> > 
> > 1. Let’s use sentence case in the ‘Review and Create’ step as well- eg: ‘Name and description of the role”
> > 2. Show the number of ‘Users and groups’ and the ‘Permission policies’.
> > 3. Can we show the policies as comma separated?
> > 4. Can we have a bit of space after the table heading?
> > 5. The empty space when no rows are present- can we remove it?
> >    Attaching a screenshot with the above points:
> > 
> > Review and create changes Let me know if there are any questions-thank you!
> 
> @parvathyvr I am using backstage's `StructuredMetadataTable` which consume the table data as key/value pairs. Its implemented in a way that the key will be rendered in a specific way only and we cannot change the case as mentioned in your point 1 and for for 2 point we cannot use any character apart alpha-numeric as it looks like they are using a similar reg-ex for generating key name and removing and special characters so brackets are not getting added. Should we show just the numbers?

Hi @divyanshiGupta, 
The idea was user can get a summary of the groups/users and permission policies with those numbers- So i would like to include the numbers if we can-but if special characters are not allowed, i wonder how it would look? Like this 👇
without -

Are we allowed to use ‘-‘ in the page or that’s also considered as a special character? If we can use '-' that would be okay👇
with -

@ShiranHi do you have any suggestions if we cannot use special characters, how could we show the numbers ?

divyanshiGupta:
> > @parvathyvr I have resolved the comments. This is how it looks:
> > Screen.Recording.2023-12-20.at.8.19.21.PM.mov
> > We can have another PR for re-aligning toast as it was not added in this.
> > Apart from that this is how I am handling any errors that occur while creating permissions: Screenshot 2023-12-20 at 8 22 44 PM
> > If you have another idea on how this should be handles lmk and we can have a separate PR for that. Thanks!
> 
> Hi @divyanshiGupta this looks good! About this error scenario- I wonder if it’s worth showing it in the ‘Roles’ table page if the role got successfully created instead of staying in the stepper. Because in this scenario this looks like role creation didn’t happen and that user has to correct something. Also since the ‘Cancel’ is not implemented in the stepper- there is no way the user can exit the stepper? That is odd- I think we need to have the ‘Cancel’ implemented for the stepper. I hope we can have the “cancel’ button? CC @ShiranHi @yuvalgalanti

@parvathyvr sure we can think about a better way for handling this. The cancel button will be implemented in @debsmita1 's PR.

divyanshiGupta:
> > > Hi @divyanshiGupta , Few points i noticed in the Review and create’ section that needs update. [Figma design to refer](https://www.figma.com/file/E6C9X3OSVzlWkW7ZsbxN7s/Administration?type=design&node-id=912-6573&mode=design&t=0gSOnQGdH5Uxflcc-4)
> > > 
> > > 1. Let’s use sentence case in the ‘Review and Create’ step as well- eg: ‘Name and description of the role”
> > > 2. Show the number of ‘Users and groups’ and the ‘Permission policies’.
> > > 3. Can we show the policies as comma separated?
> > > 4. Can we have a bit of space after the table heading?
> > > 5. The empty space when no rows are present- can we remove it?
> > >    Attaching a screenshot with the above points:
> > > 
> > > Review and create changes Let me know if there are any questions-thank you!
> > 
> > 
> > @parvathyvr I am using backstage's `StructuredMetadataTable` which consume the table data as key/value pairs. Its implemented in a way that the key will be rendered in a specific way only and we cannot change the case as mentioned in your point 1 and for for 2 point we cannot use any character apart alpha-numeric as it looks like they are using a similar reg-ex for generating key name and removing and special characters so brackets are not getting added. Should we show just the numbers?
> 
> Hi @divyanshiGupta, The idea was user can get a summary of the groups/users and permission policies with those numbers- So i would like to include the numbers if we can-but if special characters are not allowed, i wonder how it would look? Like this 👇 without -
> 
> Are we allowed to use ‘-‘ in the page or that’s also considered as a special character? If we can use '-' that would be okay👇 with -
> 
> @ShiranHi do you have any suggestions if we cannot use special characters, how could we show the numbers ?

@parvathyvr We cannot use '-'. I have added the numbers.
Screenshot 2023-12-20 at 9 25 37 PM

parvathyvr (approved):
Thanks @divyanshiGupta!

invincibleJai (approved):
LGTM

PR #1025, opened 2023-12-15
rbac(feat): add vscode debug test configuration for rbac-backend

AndrienkoAleksandr:


invincibleJai:
@AndrienkoAleksandr i tried to test it on launching debug in  VScode , i see below in terminal. Any help with what am i missing?

image

jerolimov:
@invincibleJai, you need to select a rbac-backend test and add a break point. Then it's working fine for me. And I really like it :+1: 

From my side we can merge this if this helps people.

I have just two suggestions / ideas:

1. Can we create a lunch setting that is working for all unit tests?
2. Do we really need the `ts-jest` dependency?

I tried this launch setting and it spams the console with some more warning, but on the first look it's working as well?

```json
{
  "version": "1.0.0",
  "configurations": [
    {
      "type": "node",
      "request": "launch",
      "name": "Current selected test",
      "program": "${workspaceRoot}/node_modules/.bin/backstage-cli",
      "args": ["package", "test", "${relativeFile}"],
      "console": "integratedTerminal",
      "internalConsoleOptions": "neverOpen",
    }
  ]
}
```

@AndrienkoAleksandr can you take a look if this working for you?

AndrienkoAleksandr:
@jerolimov , hello. I tested your vscode configuration. It works fine for rbac-backend-plugin.

 > Can we create a launch setting that is working for all unit tests?
 
 Looks like, we can. You provided plenty universal configuration.
 
> Do we really need the ts-jest dependency?

It is required for my pull request, because my vscode configuration uses original jest binary and It needs typescript transformer and parser.
You provided vscode configuration based on backstage cli. And I see this binary works somehow without this dependency. I guess typescript transformer can be compiled to the cli binary like transitive dependency. Also I see your solution is working for different plugin packages.
 
 > Should it --watch for changes?

I guess you proposed configuration with --watch by default, because when I modify test file, then debugger executes test one more time. That's --watch behavior from my point of view.

@jerolimov I think you can create pull request with your idea, then I can close my pull request on favor yours.

jerolimov:
@AndrienkoAleksandr Feel free to take my json and update your PR with these changes, with or without ts-lint. I'm back next week and vscode lanucher isn't a prio for me currently. ;)

AndrienkoAleksandr:
@jerolimov ok, I close this pull request on favor https://github.com/janus-idp/backstage-plugins/pull/1047

PR #1031, opened 2023-12-19
fix(quay): fix infinite progress bar when scan is unsupported

rohitkrai03:
## Description
- Added logic to consider status of a security result and show `Unsupported` if scan results was unsupported.

## Screencast

https://github.com/janus-idp/backstage-plugins/assets/6041994/1bc926f8-bc63-4884-a531-1b35aac87afa

### DIFF
--- a/plugins/quay/src/components/QuayRepository/tableHeading.tsx
+++ b/plugins/quay/src/components/QuayRepository/tableHeading.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { Link, Progress, TableColumn } from '@backstage/core-components';
 
+import { Tooltip } from '@material-ui/core';
 import makeStyles from '@material-ui/core/styles/makeStyles';
 
 import type { Layer } from '../../types';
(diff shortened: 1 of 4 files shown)


plugins/quay/src/components/QuayRepository/tableHeading.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { Link, Progress, TableColumn } from '@backstage/core-components';
 
+import { Tooltip } from '@material-ui/core';
```
gashcrumb, line 5: If possible use the MUI 5 package, `@mui/core/Tooltip` probably.

rohitkrai03: Quay plugin still uses `"@material-ui/core": "^4.12.2"`. I don't see any other plugins using Material UI 5 package either. Has there been an update recently to move to MUI 5? Maybe we can create a follow up ticket for this upgrade?

gashcrumb: Upstream will move to it eventually.  We've only just started to consider this however and are trying to stick with it for new stuff.  Totally fine to punt this concern for this PR.

rohitkrai03: Yeah, doing the upgrade in this PR will increase the scope of this work. Will create an issue for upgrading to MUI 5.

karthikjeeyar (approved):
LGTM.

invincibleJai (approved):
LGTM

PR #1023, opened 2023-12-15
fix(web-terminal): fix #598: Cleanup some small code smells in the web-terminal plugin

jerolimov:
Fix #598: Cleanup 7 code smells reported by sonarcloud:

* https://sonarcloud.io/code?id=janus-idp_backstage-plugins&selected=janus-idp_backstage-plugins%3Aplugins%2Fweb-terminal%2Fsrc%2Fcomponents%2FTerminalComponent%2FTerminalComponent.tsx
  * Checked options in conditions instead of enforcing them with `!`
* https://sonarcloud.io/code?id=janus-idp_backstage-plugins&selected=janus-idp_backstage-plugins%3Aplugins%2Fweb-terminal%2Fsrc%2Fcomponents%2FTerminalComponent%2FTerminalComponent.test.tsx
  * switched (hihi) from a switch statement to an if statement because it has just one case
* https://sonarcloud.io/code?id=janus-idp_backstage-plugins&selected=janus-idp_backstage-plugins%3Aplugins%2Fweb-terminal%2Fsrc%2Fcomponents%2FTerminalComponent%2Fstatic%2Fxterm.css
  * removed commented code

PR #1036, opened 2023-12-22
feat(web-terminal): Improve README, add isWebTerminalAvailable util and fix some other small bugs

jerolimov:
While fixing some sonarcloud warnings in #1023, I've tested the web-terminal plugin and found some small issues I would like to fix/improve with this PR:

1. Extend the README:
   * With information that an entity needs the `kubernetes.io/api-server` so that the Web Terminal work fine. It just does nothing before for entities that doesn't have this annotation.
   * Added more verbose information how to install the plugin (yarn add and some tsx examples)
   * Used the new `isWebTerminalAvailable` util in the examples.
   * Updated the `app-config.yaml` to the right path `/webterminal/rest`.
2. Added a `isWebTerminalAvailable` util function that test if the current entity has the `kubernetes.io/api-server` annotation so that it easy to hide the web terminal route or card.
3. In `TerminalComponent`:
   * The UI checked in an endless loop if the DevWorkspace is ready (Running). I've added a sleep between each retry to reduce the number of network calls. The sleep time goes a little bit up, but not above 5 seconds at the moment and stops after 5 minutes.
   * Added a warning message if the `kubernetes.io/api-server` annotation isn't used and the new gate keeper `isWebTerminalAvailable` isn't used. As said before, it just does nothing in that case.
4. Fixed a crash in `TerminalComponent/utils/requests.ts` because tries to access status but at least once it wasn't there, because the resource was just created.

https://github.com/janus-idp/backstage-plugins/assets/139310/c5e152ae-7645-492a-8860-78ef30c25f17

(I tested also that the terminal still starts.)


mimotej:
LGTM thanks!

invincibleJai (approved):

Issue #993, opened 2023-12-06
[RBAC Plugin] Add Edit Icon to the Permission Policies Card on the Role Overview Page


debsmita1:
- Add Edit Icon to the Permission Policies Card on the Role Overview Page
  - Disable the Edit icon if the user has no access to update roles
  - Clicking on the Edit icon should take the user to the Create/Edit Role Form with the `Add permission policies` section open
 - Hitting on `Save` should associate the role with the updated permission policies and users should be redirected to the roles list page and shown a toast saying the role updation is successful

PR #1037, opened 2023-12-24
feat(rbac): allow editing permission policies

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/993

**GIF:**

https://github.com/janus-idp/backstage-plugins/assets/22490998/8c1d2a53-ded6-455c-b7aa-4fdaea3431e4

https://github.com/janus-idp/backstage-plugins/assets/22490998/6c3bfe18-7693-4d4a-8e41-2eeb2f7316a4

### DIFF
--- a/plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
@@ -47,17 +47,21 @@ export const AddMembersForm = ({
   };
 
   const membersOptions: SelectedMember[] = membersData.members
-    ? membersData.members.map((member: MemberEntity, index: number) => ({
-        label: member.spec?.profile?.displayName ?? member.metadata.name,
-        description: getDescription(member),
-        etag:
+    ? membersData.members.map((member: MemberEntity, index: number) => {
+        const tag =
           member.metadata.etag ??
-          `${member.metadata.name}-${member.kind}-${index}`,
-        type: member.kind,
-        namespace: member.metadata.namespace,
-        members: getMembersCount(member),
-        ref: stringifyEntityRef(member),
-      }))
+          `${member.metadata.name}-${member.kind}-${index}`;
+        return {
+          id: tag,
+          label: member.spec?.profile?.displayName ?? member.metadata.name,
+          description: getDescription(member),
+          etag: tag,
+          type: member.kind,
+          namespace: member.metadata.namespace,
+          members: getMembersCount(member),
+          ref: stringifyEntityRef(member),
+        };
+      })
     : ([] as SelectedMember[]);
 
   return (
--- a/plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
+++ b/plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
@@ -45,7 +45,7 @@ export const CreateRolePage = () => {
               formTitle: 'Create Role',
               nameAndDescriptionTitle: 'Enter name and description of role ',
               usersAndGroupsTitle: 'Add users and groups',
-              permissionPoliciesTitle: '',
+              permissionPoliciesTitle: 'Add permission policies',
             }}
             membersData={{
               members: Array.isArray(members)
--- a/plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useAsync } from 'react-use';
 
+import { Progress } from '@backstage/core-components';
 import { useApi } from '@backstage/core-plugin-api';
 
 import { makeStyles } from '@material-ui/core';
@@ -10,10 +11,11 @@ import FormHelperText from '@mui/material/FormHelperText';
 import { FormikErrors } from 'formik';
 
 import { rbacApiRef } from '../../api/RBACBackendClient';
+import { PermissionsData } from '../../types';
 import { getPluginsPermissionPoliciesData } from '../../utils/create-role-utils';
 import { initialPermissionPolicyRowValue } from './const';
 import { PermissionPoliciesFormRow } from './PermissionPoliciesFormRow';
-import { PermissionPolicyRow, RoleFormValues } from './types';
+import { RoleFormValues } from './types';
 
 const useStyles = makeStyles(theme => ({
   permissionPoliciesForm: {
@@ -27,8 +29,8 @@ const useStyles = makeStyles(theme => ({
 }));
 
 type PermissionPoliciesFormProps = {
-  permissionPoliciesRows: PermissionPolicyRow[];
-  permissionPoliciesRowsError: FormikErrors<PermissionPolicyRow>[];
+  permissionPoliciesRows: PermissionsData[];
+  permissionPoliciesRowsError: FormikErrors<PermissionsData>[];
   setFieldValue: (
     field: string,
     value: any,
@@ -48,13 +50,16 @@ export const PermissionPoliciesForm = ({
   const classes = useStyles();
   const rbacApi = useApi(rbacApiRef);
 
-  const { value: permissionPolicies, loading: permissionPoliciesLoading } =
-    useAsync(async () => {
-      return await rbacApi.listPermissions();
-    });
+  const {
+    value: permissionPolicies,
+    loading: permissionPoliciesLoading,
+    error: permissionPoliciesErr,
+  } = useAsync(async () => {
+    return await rbacApi.listPermissions();
+  });
 
   const permissionPoliciesData =
-    !permissionPoliciesLoading && permissionPolicies
+    !permissionPoliciesLoading && Array.isArray(permissionPolicies)
       ? getPluginsPermissionPoliciesData(permissionPolicies)
       : undefined;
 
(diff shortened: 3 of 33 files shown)


plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
```diff
@@ -45,7 +45,7 @@ export const CreateRolePage = () => {
               formTitle: 'Create Role',
               nameAndDescriptionTitle: 'Enter name and description of role ',
               usersAndGroupsTitle: 'Add users and groups',
-              permissionPoliciesTitle: '',
+              permissionPoliciesTitle: 'Add permission polcies',
```
divyanshiGupta, line 48:
```suggestion
              permissionPoliciesTitle: 'Add permission policies',
```

divyanshiGupta (changes requested):
@debsmita1 Please resolve lint and tsc errors.

parvathyvr (approved):
LGTM! Thanks @debsmita1 .

divyanshiGupta:
https://github.com/janus-idp/backstage-plugins/assets/20724543/3e1c8b1a-a1e8-4954-bfbd-72036c3f8af6

Screenshot 2024-01-05 at 3 39 46 PM

@debsmita1 as you can see in above rec/image that before the permissions are loaded the component is being rendered and hence it shows an error. IMO when the permissions are loading lets show a progress component and if there is any error once after the loading is completed then show the error. 

Also on clicking `cancel` IMO the user should be redirected to the previous page it came from. For ex in above case on clicking cancel it should navigate back to role overview page.

 cc: @parvathyvr

divyanshiGupta (changes requested):
https://github.com/janus-idp/backstage-plugins/assets/20724543/3e1c8b1a-a1e8-4954-bfbd-72036c3f8af6

Screenshot 2024-01-05 at 3 39 46 PM

@debsmita1 as you can see in above rec/image that before the permissions are loaded the error component is being rendered since the permissions are undefined and the above error is shown. IMO when the permissions are loading lets show a progress component and if there is any error/permissions are undefined once after the loading is completed then show the error. 

Also on clicking `cancel` IMO the user should be redirected to the previous page it came from. For ex in above case on clicking cancel it should navigate back to role overview page.

 cc: @parvathyvr

divyanshiGupta:
@debsmita1 I am also seeing an issue wherein sometimes the navigation is not working properly. For ex I saw an issue when clicked on save button on edit role form and it navigated to rbac page but the list was not rendered. This doesnt happen always.

Screenshot 2024-01-05 at 4 02 24 PM

debsmita1:
> Screen.Recording.2024-01-05.at.3.37.38.PM.mov 
> Screenshot 2024-01-05 at 3 39 46 PM
> @debsmita1 as you can see in above rec/image that before the permissions are loaded the error component is being rendered since the permissions are undefined and the above error is shown. IMO when the permissions are loading lets show a progress component and if there is any error/permissions are undefined once after the loading is completed then show the error.
> 
> Also on clicking `cancel` IMO the user should be redirected to the previous page it came from. For ex in above case on clicking cancel it should navigate back to role overview page.
> 
> cc: @parvathyvr


Done. PTAL
https://github.com/janus-idp/backstage-plugins/assets/22490998/fd74c844-1463-4a76-8935-6cc2837a34ec

debsmita1:
> @debsmita1 I am also seeing an issue wherein sometimes the navigation is not working properly. For ex I saw an issue when clicked on save button on edit role form and it navigated to rbac page but the list was not rendered and after some time is gets rendered. This doesnt happen always.
> 
> Screenshot 2024-01-05 at 4 02 24 PM

Couldn't reproduce this. If you see this again could you record the network logs ?

divyanshiGupta:
> > @debsmita1 I am also seeing an issue wherein sometimes the navigation is not working properly. For ex I saw an issue when clicked on save button on edit role form and it navigated to rbac page but the list was not rendered and after some time is gets rendered. This doesnt happen always.
> > Screenshot 2024-01-05 at 4 02 24 PM
> 
> Couldn't reproduce this. If you see this again could you record the network logs ?

@debsmita1 the issues seems to be with rendering the whole page. If the data is not available then the table should be shown with empty state but here the entire page is not rendering. I will check the network logs but in case of data failure also the whole page UI should not break. I will keep an eye on this issue.

divyanshiGupta:
@debsmita1 On going to the edit form with active step as `Edit permission policies` I dont see a error now but I dont see a progress bar either. When the policies are loading nothing is displayed and then suddenly the table is shown. 


https://github.com/janus-idp/backstage-plugins/assets/20724543/29738a11-6d4e-495c-91bf-feb3f5b0ecd2

divyanshiGupta:
@debsmita1 On adding a policy row and adding values to the fields a warning saying `MUI: The value provided to Autocomplete is invalid` is being shown. Please fix it.


https://github.com/janus-idp/backstage-plugins/assets/20724543/594979d0-6393-4cf3-bd61-4c7e7cd0634d

plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
```diff
@@ -121,33 +126,50 @@ export const PermissionPoliciesForm = ({
...
-            handleBlur={handleBlur}
-          />
-        ))}
-        <Button className={classes.addButton} size="small" onClick={onRowAdd}>
-          <AddIcon />
-          Add
-        </Button>
-      </div>
+      {permissionPoliciesLoading ? (
+        <Progress />
```
debsmita1, line 130: @divyanshiGupta  I am checking here already if the permission policies are still loading

divyanshiGupta: Yeah but it's not being shown 🤔 can you check why is that? @debsmita1

debsmita1:
> @debsmita1 On adding a policy row and adding values to the fields a warning saying `MUI: The value provided to Autocomplete is invalid` is being shown. Please fix it.
> 
>  Screen.Recording.2024-01-09.at.11.56.04.AM.mov

Fixed it

plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
debsmita1: Maybe because the value of `loading` updates from true to false way faster

divyanshiGupta: Yeah. So to handle this we will have to show the progress bar for a minimum of 1-2 secs so that the progress bar is shown correctly and if loading takes more than that we show the progress bar for the entire duration. LMK if you have some other idea.

invincibleJai: I think we should definitely have the loading/progress bar handled but at the same time, we should show the progress bar only if it's actually loading be it for a second or more, as loading gets false we should render the data.

@debsmita1 pls test with throttling in the browser.

debsmita1: I changed the network preset to 'Fast 3G' and the ProgressBar is visible
This means that the progress bar does show up when there's a delay in fetching the results

https://github.com/janus-idp/backstage-plugins/assets/22490998/fbd0702c-2d4b-4caf-a3c8-4a8ba09c8fbd

divyanshiGupta: yeah, the issue is because the loading is taking probably like some ms but I pointed this out because it shows nothing for those ms and then the table is rendered. Definitely if the loading takes more time the progress bar will be shown and that is where I see the inconsistency. But I suppose its not a big issue right now and we can always come back to it later if its pointed out.

divyanshiGupta:
@debsmita1 The PR looks good and works as expected but I see two minor issues:

https://github.com/janus-idp/backstage-plugins/assets/20724543/ce3604f5-615a-44cd-8271-7078eccb6d3e

1. Now I see a warning for members dropdown
2. On adding new policy row, permission and plugin dropdown's bg color is changed to blue

debsmita1:
> @debsmita1 The PR looks good and works as expected but I see two minor issues:
> 
>  Screen.Recording.2024-01-10.at.3.08.11.PM.mov 
> 1. Now I see a warning for members dropdown
> 2. On adding new policy row, permission and plugin dropdown's bg color is changed to blue

@divyanshiGupta  I couldn't find a warning for the members dropdown, but there appears to be an issue with the autocomplete components. The solution mentioned in the warning doesn't seem to be working. I found an open issue on github https://github.com/mui/material-ui/issues/29727. However, I did fix the background color of the plugin dropdown

invincibleJai (approved):
LGTM

Thanks @debsmita1, this looks good.

plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
```diff
@@ -69,7 +75,7 @@ export const AddMembersForm = ({
       <br />
       <Autocomplete
         options={membersOptions}
-        getOptionLabel={(option: SelectedMember) => option.label}
+        getOptionLabel={(option: SelectedMember) => option.label || ''}
```
divyanshiGupta, line 78: @debsmita1 is there a scenario where the label can be undefined? If yes then I am not sure showing empty string is the right option here. Maybe in stead of showing empty string we can show the kind or something that will be visible in the drop-down?

debsmita1: removed this change

divyanshiGupta (approved):
LGTM!

PR #1052, opened 2024-01-09
feat(ocm): add sorting to the ocm cluster table

### LINKED ISSUE #769: OCM cluster table isn't sortable
### Describe the bug

OCM clusters are presented in a table with columns: Name, Status, Infrastructure, Version, and Nodes.

The columns allow the user to click on them to order the rows by these columns, but it looks like it has no effect.

https://github.com/janus-idp/backstage-plugins/assets/139310/062e296a-6ce2-4c98-8e63-689f5e3a4f2b

### Expected Behavior

Table should be sortable.

1. Use a cluster with ocm and at least two connected clusters
2. Click on the columns to sort the table

Zaperex:
### Description
Adds sorting to the OCM cluster table

### What issue(s) does this PR fix
Fixes #769


invincibleJai:
@Zaperex test job seems to be failing

Zaperex:
Hmm it seems to be failing due to tests from the `@janus-idp/backstage-plugin-web-terminal` package at https://github.com/janus-idp/backstage-plugins/actions/runs/7494511967/job/20482135625#step:3:1257

plugins/ocm/src/components/ClusterStatusPage/ClusterStatusPage.tsx
```diff
@@ -173,10 +172,22 @@ const CatalogClusters = () => {
             title: 'Name',
             field: 'name',
             highlight: true,
+            customSort: (a, b) => {
+              return a.name.props.children.localeCompare(
+                b.name.props.children,
+                'en',
+              );
+            },
```
divyanshiGupta, line 180: Hi @Zaperex can you move the columns to a separate file and also create a type for data that each table row represents so that you dont have to type-cast params manually in the `customSort` function ? Something like:
```
         export const columns: TableColumn<RowType>[] = [{
            title: 'Name',
            field: 'name',
            highlight: true,
            customSort: (a: RowType, b: RowType) => {
              return a.name.props.children.localeCompare(
                b.name.props.children,
                'en',
              );
            },
          {
            title: 'Status',
            field: 'status',
            customSort: (a: RowType, b: RowType) => {
              const availabilityA = a.status.props.status.available;
              const availabilityB = b.status.props.status.available;
              if (availabilityA === availabilityB) return 0;
              return availabilityA ? -1 : 1;
            },
          },
           ...
         ]
```

divyanshiGupta (changes requested):

plugins/ocm/src/components/ClusterStatusPage/ClusterStatusPage.tsx
Zaperex: Just noting it down here that each row is actually the following type. I'll see if I can get the correct types.
```tsx
{
    name: React.JSX.Element;
    status: React.JSX.Element;
    infrastructure: string;
    version: React.JSX.Element;
    nodes: React.JSX.Element;
}
```

plugins/ocm/package.json
```diff
@@ -39,7 +39,8 @@
     "@material-ui/lab": "^4.0.0-alpha.45",
     "@patternfly/patternfly": "^5.1.0",
     "@patternfly/react-icons": "^5.1.1",
-    "react-use": "^17.4.0"
+    "react-use": "^17.4.0",
+    "semver": "^7.5.4"
```
divyanshiGupta, line 43: @Zaperex just wanted to confirm if the package addition here didn't have any yarn.lock updates?

divyanshiGupta (commented):
LGTM

plugins/ocm/package.json
Zaperex: The `semver` package should already be included in the `yarn.lock` so regeneration shouldn't be required

Zaperex:
@divyanshiGupta Not sure how to do that without destroying the commit history at the moment since I've merged from main a few times during the lifetime of this PR. Shouldn't there be a merge option in github that allows for a PR's commits to be squashed before it gets merged?

divyanshiGupta (approved):
/lgtm

Issue #1030, opened 2023-12-19
RBAC UI: RBAC UI List page doesn't watch policies


invincibleJai:
### Describe the bug

RBAC UI List page doesn't watch policies, it only watches roles.

<img width="1639" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/0c31361d-3305-4d61-b216-bc0caa804474">

### Expected Behavior

The list page should watch for policies

### What are the steps to reproduce this bug?

1. Install and configure RBAC backend and RBAC frontend plugin

### Versions of software used and environment

@janus-idp/backstage-plugin-rbac: 1.7.0
@janus-idp/backstage-plugin-rbac-backend: 2.0.0

PR #1057, opened 2024-01-11
fix(rbac): fix the roles table to also watch policies

debsmita1:
**Resolves:**
https://github.com/janus-idp/backstage-plugins/issues/1030

**GIF:**

https://github.com/janus-idp/backstage-plugins/assets/22490998/92d08248-00b4-4f54-aa4b-35292291e03a

Error scenarios

https://github.com/janus-idp/backstage-plugins/assets/22490998/7e3c9464-9b87-4950-a4f6-5a5e6ccbf7fa

Screenshot 2024-01-22 at 8 43 11 PM

### DIFF
--- a/plugins/rbac/src/components/RolesList/RolesList.tsx
+++ b/plugins/rbac/src/components/RolesList/RolesList.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { Table } from '@backstage/core-components';
+import { Table, WarningPanel } from '@backstage/core-components';
 
 import { makeStyles } from '@material-ui/core';
 
@@ -27,12 +27,13 @@ export const RolesList = () => {
 
   const [roles, setRoles] = React.useState<number | undefined>();
   const classes = useStyles();
-  const { loading, data, retry, createRoleAllowed, createRoleLoading } =
+  const { loading, data, retry, createRoleAllowed, createRoleLoading, error } =
     useRoles();
 
   const closeDialog = () => {
     setOpenDialog(false);
-    retry();
+    retry.roleRetry();
+    retry.policiesRetry();
   };
 
   const onAlertClose = () => {
@@ -49,6 +50,19 @@ export const RolesList = () => {
         createRoleAllowed={createRoleAllowed}
         createRoleLoading={createRoleLoading}
       />
+      {(error?.rolesError || error?.policiesError) && (
+        <div style={{ paddingBottom: '16px' }}>
+          <WarningPanel
+            message={error.rolesError || error.policiesError}
+            title={
+              error.rolesError
+                ? 'Something went wrong while fetching the roles'
+                : 'Something went wrong while fetching the permission policies'
+            }
+            severity="error"
+          />
+        </div>
+      )}
       <Table
         title={
           !loading && data?.length
--- a/plugins/rbac/src/hooks/useRoles.ts
+++ b/plugins/rbac/src/hooks/useRoles.ts
@@ -24,19 +24,24 @@ export const useRoles = (
   data: RolesData[];
   createRoleLoading: boolean;
   createRoleAllowed: boolean;
-  retry: () => void;
+  error: {
+    rolesError: string;
+    policiesError: string;
+  };
+  retry: { roleRetry: () => void; policiesRetry: () => void };
 } => {
   const rbacApi = useApi(rbacApiRef);
   const {
-    loading: rolesLoading,
     value: roles,
     retry: roleRetry,
+    error: rolesError,
   } = useAsyncRetry(async () => await rbacApi.getRoles());
 
-  const { loading: policiesLoading, value: policies } = useAsyncRetry(
-    async () => await rbacApi.getPolicies(),
-    [],
-  );
+  const {
+    value: policies,
+    retry: policiesRetry,
+    error: policiesError,
+  } = useAsyncRetry(async () => await rbacApi.getPolicies(), []);
 
   const deletePermissionResult = usePermission({
     permission: policyEntityDeletePermission,
@@ -67,7 +72,7 @@ export const useRoles = (
   });
   const data: RolesData[] = React.useMemo(
     () =>
-      roles && roles?.length > 0
+      Array.isArray(roles) && roles?.length > 0
         ? roles.reduce((acc: RolesData[], role: Role) => {
             const permissions = getPermissions(
               role.name,
@@ -107,10 +112,12 @@ export const useRoles = (
       catalogEntityReadPermissionResult,
     ],
   );
-  const loading = rolesLoading && policiesLoading;
+  const loading = !rolesError && !policiesError && !roles && !policies;
+
   useInterval(
     () => {
       roleRetry();
+      policiesRetry();
     },
     loading ? null : pollInterval || 10000,
   );
@@ -118,8 +125,18 @@ export const useRoles = (
   return {
     loading,
     data,
+    error: {
+      rolesError: (rolesError?.message ||
+        (typeof roles === 'object'
+          ? (roles as any as Response)?.statusText
+          : '')) as string,
+      policiesError: (policiesError?.message ||
+        (typeof policies === 'object'
+          ? (policies as any as Response)?.statusText
+          : '')) as string,
+    },
     createRoleLoading,
     createRoleAllowed,
-    retry: roleRetry,
+    retry: { roleRetry, policiesRetry },
   };
 };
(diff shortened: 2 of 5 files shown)


plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
```diff
@@ -73,7 +75,7 @@ export const AddMembersForm = ({
       <br />
       <Autocomplete
         options={membersOptions}
-        getOptionLabel={(option: SelectedMember) => option.label}
+        getOptionLabel={(option: SelectedMember) => option.label || ''}
```
divyanshiGupta, line 78: @debsmita1 We had this conversation regarding removing this here https://github.com/janus-idp/backstage-plugins/pull/1037#discussion_r1449893349. LMK if this was added back intentionally.

divyanshiGupta (changes requested):

plugins/rbac/src/components/RolesList/RolesList.tsx
```diff
@@ -42,6 +44,16 @@ export const RolesList = () => {
...
+  React.useEffect(() => {
+    let mounted = false;
+    if (!mounted && !loading) {
+      setIsMounted(true);
+    }
+    return () => {
+      mounted = true;
+    };
+  }, [loading]);
+
```
ciiay, line 56: The condition `if (!mounted && !loading) ` will always be true when `loading` is false as `mounted` is being initialized to `false`. Since `mounted` here is a local variable within each render, changes made by the cleanup function will not persist across renders. How about we use `useRef` instead?

```
import React, { useEffect, useRef, useState } from 'react';

export const RolesList = () => {
  // ...
  const [isMounted, setIsMounted] = useState(false);
  const mountedRef = useRef(false);

  useEffect(() => {
    mountedRef.current = true;
    if (!loading) {
      setIsMounted(true);
    }
    return () => {
      mountedRef.current = false;
    };
  }, [loading]);

  // ...
};
```

ciiay (changes requested):

ciiay (approved):
LGTM
Tested locally and it works perfectly. Thank you.

ciiay:
LGTM
Pulled new change and the yarn.lock file stays the same 👍

divyanshiGupta:
@debsmita1 why are there so many changes in yarn.lock when no package is being added/updated? Also can you rebase your PR with main and then re-push?

debsmita1:
> @debsmita1 why are there so many changes in yarn.lock when no package is being added/updated? Also can you rebase your PR with main and then re-push?

I had rebased my branch with the main, but then the CI complained about shared-react package version mismatch :/ To fix this, I removed the yarn.lock and node modules and re-installed it, generating an updated yarn.lock
I have now dropped the yarn.lock from my commit and the CI looks good now. PTAL

plugins/rbac/src/components/RolesList/RolesList.tsx
```diff
@@ -42,6 +45,16 @@ export const RolesList = () => {
     setRoles(searchResults.length);
   };
 
+  React.useEffect(() => {
+    mountedRef.current = true;
+    if (!loading) {
```
divyanshiGupta, line 50: @debsmita1 Value of `mountedRef` is being changed but not used anywhere. Am I missing something?

divyanshiGupta (commented):
@debsmita1 Can you squash your commits into 1 as the 1st  commit changes are no longer available in the PR?

divyanshiGupta:
> > @debsmita1 why are there so many changes in yarn.lock when no package is being added/updated? Also can you rebase your PR with main and then re-push?
> 
> I had rebased my branch with the main, but then the CI complained about shared-react package version mismatch :/ To fix this, I removed the yarn.lock and node modules and re-installed it, generating an updated yarn.lock I have now dropped the yarn.lock from my commit and the CI looks good now. PTAL

Yes there was an issue with shared-react version in main but that was fixed by Christoph's PR and that is why the CI works fine now. I figured that your PR was outdated  :)

plugins/rbac/src/components/RolesList/RolesList.tsx
ciiay: Right, I didn't realize that only `isMounted` is being used for the `isLoading` value assignment, in which case the below solution should be sufficient. I have confirmed that it works as the same behavior we looked together before @debsmita1 . Unless there's any concern why the local variable `mounted` was used in the first place.

```
  React.useEffect(() => {
    if (!loading) {
      setIsMounted(true);
    }
  }, [loading]);
```

When I tested this solution I found another issue of this component. The total number of roles keeps refreshing every time useRoles fetches data.

https://github.com/janus-idp/backstage-plugins/assets/26255262/1e5eecd8-6769-4121-8a61-aaf385734fa5

I came up a way to make it only refreshes when the data length changes.
```
export const RolesList = () => {
...
const isInitialLoad = React.useRef(true);
...
  React.useEffect(() => {
    if (!loading) {
      setIsMounted(true);  // set isMounted value on loading value changes

      if (isInitialLoad.current) {
        setRoles(data.length);  // update roles only on initial load or onSearchResultsChange
        isInitialLoad.current = false;
      }
    }
  }, [loading, data]);
 
 // use useMemo to avoid recalculation on every render
 const tableTitle = React.useMemo(() => {
    const roleCount = roles ?? data.length;
    return roleCount > 0 ? `All roles (${roleCount})` : 'All roles';
  }, [roles, data.length]);
...
      <Table
        title={tableTitle}  // still gives default value when roles is undefined or null
        ...
      />
...
```
How do you think?

divyanshiGupta: Are we using this `useEffect` because the value of `loading` is changing from the hook? In that case I feel it should be handled at the hook level. It should not flip the loading values from `true` -> `false` -> `true` during fetches. It should go from `true` -> `false` only once when the first set of data is loaded and not after that. After that just the data should be updated with new values. Then we can simply say `isLoading=loading` like previously. @debsmita1 WDYT?

```
React.useEffect(() => {
    if (!loading) {
      setIsMounted(true);
    }
  }, [loading]);
```

divyanshiGupta: About the count of the roles changing during subsequent fetches, I feel it should also be coming from the hook and we should not be calculating it in the component. The hook should maintain the state for total number of roles and the component should be able to consistently rely on that value.

debsmita1:
Updated the PR @divyanshiGupta @ciiay PTAL

ciiay (approved):
LGTM
Works great, thank you.

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -107,10 +105,11 @@ export const useRoles = (
       catalogEntityReadPermissionResult,
     ],
   );
-  const loading = rolesLoading && policiesLoading;
+  const loading = !roles && !policies;
```
divyanshiGupta, line 108: @debsmita1 this won't work if there is an error while fetching roles/policies as they will remain undefined. One thing you can do here is that check for errors as well. Like this `const loading =  !rolesError && !policiesError && !roles && !policies`. But since we already have a loading value you can also maintain a state for that and its value will only be set to false for the first time loading becomes false. I have mentioned the same here https://github.com/janus-idp/backstage-plugins/pull/1057#discussion_r1459004637. This is similar to what you were doing in the component file.

divyanshiGupta (changes requested):

plugins/rbac/src/hooks/useRoles.ts
debsmita1: @divyanshiGupta I  have tested this when there is an error while fetching the roles/policies. When there is an error with roles or policies, they return an object. Which will set loading as `false`

divyanshiGupta: Okay then that should be fine. Also we show errors in the list right? What will happen in case of a network error?

debsmita1: @divyanshiGupta  Earlier I was not handling network error scenarios nor was any error shown in the list page. I have updated the code and attached screenshots in the description. Thank you for pointing this out.

divyanshiGupta:
@debsmita1 CI is failing.

divyanshiGupta:
cc: @parvathyvr

parvathyvr (approved):
LGTM.

plugins/rbac/src/components/RolesList/RolesList.tsx
divyanshiGupta: > About the count of the roles changing during subsequent fetches, I feel it should also be coming from the hook and we should not be calculating it in the component. The hook should maintain the state for total number of roles and the component should be able to consistently rely on that value.

@debsmita1 there was another issue pointed out wherein the roles count keeps refreshing between fetches. I have suggested how we can handle it above. LMK if you want to fix it in this PR or have another ticket for that?

debsmita1: The role count now stays as is when the API is fetched. The only time it updates is when the search results update

divyanshiGupta (approved):
/approve

PR #1061, opened 2024-01-15
feat(notifications): make notifications NPM packages public

mareklibra:


plugins/notifications-backend/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "src/index.ts",
   "types": "src/index.ts",
   "license": "Apache-2.0",
-  "private": true,
+  "private": false,
```
invincibleJai, line 7: we can just remove it over specifying it as false

plugins/notifications/package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "@backstage/plugin-notifications-frontend",
+  "name": "@backstage/plugin-notifications",
   "description": "Frontend part of the notifications providing the Notifications API, Notifications page, left-side menu item and other REACT components.",
   "version": "0.1.0",
   "main": "src/index.ts",
   "types": "src/index.ts",
   "license": "Apache-2.0",
-  "private": true,
+  "private": false,
```
invincibleJai, line 8: same as above, we can avoid this over making it false

mareklibra: done

invincibleJai (approved):
/lgtm

PR #1064, opened 2024-01-15
feat(rbac): show warning alert when user is not authorised to create roles

divyanshiGupta:
Fixes: #1065 

Screenshot 2024-01-15 at 9 10 16 PM


schultzp2020 (approved):
/lgtm

divyanshiGupta:
@hmanwani-rh I have used this README link here for rbac prerequisites https://github.com/janus-idp/backstage-plugins/tree/main/plugins/rbac#prerequisites. LMK if some other link should be used here instead.

cc: @invincibleJai

PR #1066, opened 2024-01-16
fix(servicenow): make sure that the correct openapi generator is executed

kadel:
Tests for servicenow-actions were failing due to `Error: command --input not found`.
This was due to the wrong open API generator being executed. Instead of https://github.com/ferdikoomen/openapi-typescript-codegen the https://github.com/openapistack/openapicmd was started.


invincibleJai (approved):
/lgtm

PR #1071, opened 2024-01-17
test(web-terminal) Remove TerminalComponent.test.tsx race condition issue

jerolimov:
Investigate and fixed an issue that the `TerminalComponent.test.tsx` fails sometimes on the CI: [example](https://github.com/janus-idp/backstage-plugins/actions/runs/7542203125/job/20536126310?pr=783#step:3:2443)

The test contained a delay (800ms) to verify that it shows a `<Progress />` while loading the web terminal.

After removing that delay the "should start loading" tests failed until I noticed that the backstage `<Progress />` contain a delay until it was rendered:

https://github.com/backstage/backstage/blob/a070f29921cf7207a62e76abe5330f23364fdb07/packages/core-components/src/components/Progress/Progress.tsx#L24-L41

The tests run successfully after adding a ThemeProvider that disables that transition duration = delay.

After that, I removed the test anyway because the tests continue with the network call and logs an "update react component after unmount error". Instead of an extra test, I added the "show a <Progress /> component assertion to the "should display terminal window" as it contains the same steps.


invincibleJai (approved):
/lgtm

PR #1077, opened 2024-01-17
fix(notifications): a few fixes when adding the pkgs to the show-case

mareklibra:
These fixes have been created when adding the `Notifications` to the showcase instance.


plugins/notifications-backend/src/service/router.ts
```diff
@@ -19,6 +19,10 @@ import {
 } from './permissions';
 import { RouterOptions } from './types';
 
+// TODO: fix filename resolving
```
mareklibra, line 22: This is a hot-fix. We will search for a better way

plugins/notifications-backend/src/service/handlers.ts
```diff
@@ -353,7 +353,9 @@ function createQuery(
   // filter by text
   if (filter.containsText) {
     query.andWhere(function () {
-      this.whereILike('title', `%${filter.containsText}%`).orWhereILike(
+      // case-insensitive "whereILike" should be used instead but
```
ydayagi, line 356: it should be ilike. i would like to know what is the bug before considering a change in the code.

plugins/notifications-backend/package.json
```diff
@@ -55,6 +55,7 @@
     "supertest": "6.3.3"
   },
   "files": [
-    "dist"
+    "dist",
+    "src/openapi.yaml"
```
ydayagi, line 59: what is the use of adding the yaml here?

plugins/notifications-backend/src/service/router.ts
ydayagi: the file name was hard coded and so it should be. the backend is tied tightly to the spec.

ydayagi (changes requested):

plugins/notifications-backend/package.json
mareklibra: So it becomes part of the tgz published to the NPM registry. It was missing and so failing build.

plugins/notifications-backend/src/service/router.ts
mareklibra: We need to improve, so far it's still a hotfix, we need to have it working by tomorrow.
The former path worked well when used in the development wrapping backstage instance. But was not matching when the plugin is added as an NPM dependency to node_modules (which is are ultimate goal).

plugins/notifications-backend/src/service/handlers.ts
mareklibra:
```
Request failed with status 500 select * from `messages` left join (select * from `users` where `users`.`user` = 'default/guest') as `users` on `messages`.`id` = `users`.`message_id` where ((`is_system` = false and (`id` in (select `message_id` from `users` where `user` = 'default/guest')))) and (`title` ilike '%m%' or `message` ilike '%m%') and `created` > '2024-01-10T15:11:16.204Z' and (`read` = false or `read` is null) order by `created` desc limit 5 - near "ilike": syntax error type=errorHandler code=SQLITE_ERROR stack=SqliteError: select * from `messages` left join (select * from `users` where `users`.`user` = 'default/guest') as `users` on `messages`.`id` = `users`.`message_id` where ((`is_system` = false and (`id` in (select `message_id` from `users` where `user` = 'default/guest')))) and (`title` ilike '%m%' or `message` ilike '%m%') and `created` > '2024-01-10T15:11:16.204Z' and (`read` = false or `read` is null) order by `created` desc limit 5 - near "ilike": syntax error
```


Check the referenced knex issue, it's a bug and this is a workaround

plugins/notifications-backend/package.json
ydayagi: how do the rest of the files get into the tgz? if they get to the tgz then so should this one

plugins/notifications-backend/src/service/handlers.ts
ydayagi: we must not change it for all DBs just because there's a bug in knex with sqlite. we can try other solutions

plugins/notifications-backend/src/service/router.ts
ydayagi: that does not explain the technical problem. why should npm package care where my file is stored? i provide a folder with files and it should package it, no?

ydayagi (changes requested):

plugins/notifications-backend/package.json
ydayagi: only dist is javascript, correct? so that's what goes into the package? and u want to add the openapi spec. correct?

plugins/notifications-backend/src/service/handlers.ts
mareklibra: I agree. Just we should find it today. Or merge this and keep searching.
Obvious workaround can be an `if` for the SQLite but maybe there is something better.

plugins/notifications-backend/src/service/router.ts
mareklibra: Note: the Janus NPM packages have been published, so for sure we don't need my provate account mentioned bellow

plugins/notifications-backend/package.json
mareklibra: We might place it into `dist` as well if it makes sense. The point is to have the file reachable when the package is added as an NPM dependency. Our previous testing assumed the `plugins` folder.

pkliczewski:
@christophe-f we need this PR for tomorrow demo, please help use merge it?

PR #1086, opened 2024-01-18
chore(cd): fix semantic-release permission issues

kadel:
This should finally make possible for semantic-release to push tags even with branch protection enabled


invincibleJai (approved):
/lgtm

PR #1087, opened 2024-01-18
fix(notifications): hardcode openapi.yaml path for the demo

mareklibra:


mareklibra:
Demo is done and the changes are addressed in a clean way already, so closing here

PR #1089, opened 2024-01-18
feat(orchestrator): businesskey and context sharing

RichardW98:
businesskey and context sharing

### DIFF
--- a/plugins/orchestrator/src/components/ExecuteWorkflowPage/ExecuteWorkflowPage.tsx
+++ b/plugins/orchestrator/src/components/ExecuteWorkflowPage/ExecuteWorkflowPage.tsx
@@ -48,6 +48,20 @@ export const ExecuteWorkflowPage = () => {
       await orchestratorApi.getWorkflowDataInputSchema(workflowId),
     [orchestratorApi, workflowId],
   );
+  const { value: variables } = useAsync(async (): Promise<
+    Record<string, JsonValue>[] | undefined
+  > => {
+    if (businessKey !== undefined) {
+      const instance = await orchestratorApi.getInstance(businessKey);
+      const vs = instance.variables as Record<string, JsonValue>;
+      const response: Record<string, JsonValue>[] = [];
+      Object.entries(vs?.workflowdata ?? {}).forEach(([key, value]) => {
+        if (key !== 'workflowOptions') response.push({ [key]: value });
+      });
+      return response;
+    }
+    return undefined;
+  }, [businessKey]);
 
   const handleExecute = useCallback(
     async (getParameters: () => Record<string, JsonValue>) => {
@@ -107,6 +121,7 @@ export const ExecuteWorkflowPage = () => {
                 refSchemas={schemaResponse.schemas}
                 handleExecute={handleExecute}
                 isExecuting={isExecuting}
+                initialState={variables}
               />
             ) : (
               <JsonTextAreaForm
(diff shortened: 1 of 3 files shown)


caponetto:
This code is for the demo but we should probably polish the solution.
@batzionb could you please take a look too?

plugins/orchestrator/src/components/ExecuteWorkflowPage/ExecuteWorkflowPage.tsx
```diff
@@ -48,6 +48,20 @@ export const ExecuteWorkflowPage = () => {
...
+  const { value: variables } = useAsync(async (): Promise<
+    Record<string, JsonValue>[] | undefined
+  > => {
+    if (businessKey !== undefined) {
+      const instance = await orchestratorApi.getInstance(businessKey);
+      const vs = instance.variables as Record<string, JsonValue>;
+      const response: Record<string, JsonValue>[] = [];
+      Object.entries(vs?.workflowdata ?? {}).forEach(([key, value]) => {
+        if (key !== 'workflowOptions') response.push({ [key]: value });
+      });
```
caponetto, line 60: This logic is too specific and we would need a more generic one.
For this use case, one idea is to traverse the `instance.variables` object looking for property keys that match properties present in the schema(s). Whenever there is a match, you can include the property/value as initial state but it is very important to follow the structure of the schema(s).
Taking your use case as an example, you'd end up with a single match: `gitRepositoryUrl` because this property is present in both variables and the schema. On the other hand, `workflowOptions` would be discarded because it is not present in the schema.
cc @batzionb

caponetto (changes requested):

plugins/orchestrator/src/components/ExecuteWorkflowPage/ExecuteWorkflowPage.tsx
RichardW98: what we expect is to pop the value to the matching field and disable it. the value should be fixed and passed from assessment

RichardW98: yes this logic won't work if there're other properties not belong to schema

caponetto: Behind the scenes, we are using `react-jsonschema-form`. This library offers the possibility to disable fields through `uiSchema` (see [the code](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator/src/components/ExecuteWorkflowPage/StepperForm.tsx#L81)). Currently, we only use it to set the focus in the first field but this could be expanded to disable fields that have an initial state. See [the documentation](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/uiSchema#disabled).

Moreover, this new feature we have here will also be important for "workflow restart button without providing input parameters again".

batzionb:
What bug is this supposed to solve?
Can you include a Jira issue please?

plugins/orchestrator/src/components/ExecuteWorkflowPage/ExecuteWorkflowPage.tsx
batzionb: Maybe the backend should return the schema the UI should display

RichardW98: the ui schema is basically translated from input-schema.json

RichardW98:
yes, this is the jira issue: https://issues.redhat.com/browse/FLPATH-776

PR #1096, opened 2024-01-21
fix(notifications): create DB upon startup

ydayagi:
[FLPATH-875](https://issues.redhat.com//browse/FLPATH-875)
https://issues.redhat.com/browse/FLPATH-875


mareklibra:
I think it's good-enough for now.

Anyway, the upstream plugins instantiate the DB differently - they use services by the Backstage.
In addition, they use `knex migrations`.

If interested, please have a look at https://github.com/backstage/backstage/pull/22099, I have changed it there.
But considering context here, we should be ok with this change as well.

mareklibra (approved):

Issue #1093, opened 2024-01-19
[kiali] Kiali Plugin not reach the backend correctly when you add a path


aljesusg:
### Describe the bug

The kiali-backend not let to add a full path in the url.
When you set in the pp-config the kiali provider url with a subpath kiali-backend ignore it.

The error is in L150 in kiali-backend/src/client/fetch.ts

```ts
requestInit.url = new URL(endpoint, this.KialiDetails.url).href;
```

the URL class ignore if you set a path in the url

### Expected Behavior

Query to **https://kiali-istio-system.apps-crc.testing/kiali/api/auth/info**

### What are the steps to reproduce this bug?

1. Add in app-config the kiali provider the url with a path (Ex: https://kiali-istio-system.apps-crc.testing/kiali)
2. Check that kiali is triying to reach **https://kiali-istio-system.apps-crc.testing/api/auth/info** instead
**https://kiali-istio-system.apps-crc.testing/kiali api/auth/info**

Please add **plugin/kiali** label and assign this one to @josunect

josunect:
Can I get this assigned? Thanks

PR #1097, opened 2024-01-22
fix(kiali): update login url

josunect:
Fix for Kiali plugin to use a url with a path as the Kiali backend url. 

Tested with: 

- `http://localhost:3001`
- `http://localhost:20001/kiali`

Fixes https://github.com/janus-idp/backstage-plugins/issues/1093

### DIFF
--- a/plugins/kiali-backend/src/clients/fetch.ts
+++ b/plugins/kiali-backend/src/clients/fetch.ts
@@ -147,7 +147,7 @@ export class KialiFetcher {
       };
     }
 
-    const loginUrl = `${this.KialiDetails.url}/${endpoint}`;
+    const loginUrl = `${this.KialiDetails.url}/${endpoint.replace(/^\//g, '')}`;
     requestInit.url = new URL(loginUrl).href;
 
     if (this.KialiDetails.skipTLSVerify) {


plugins/kiali-backend/src/clients/fetch.ts
```diff
@@ -147,7 +147,8 @@ export class KialiFetcher {
       };
     }
 
-    requestInit.url = new URL(endpoint, this.KialiDetails.url).href;
+    const loginUrl = `${this.KialiDetails.url}/${endpoint}`;
```
aljesusg, line 150: Toa void results like : **https://kiali-istio-system.apps-crc.testing/kiali//api/status** with 2 "/" I would add

```ts
const loginUrl = `${this.KialiDetails.url}/${endpoint.replace(/^\//g, "");}`;
```

To be sure that it's going to work. What do you think?

aljesusg (changes requested):
What do you think about this?

plugins/kiali-backend/src/clients/fetch.ts
josunect: Thanks for the suggestion, @aljesusg ! 
I have included for testing: 
`url: http://localhost:20001/kiali/`
And works fine.

aljesusg: > Toa void results like : **https://kiali-istio-system.apps-crc.testing/kiali//api/status** with 2 "/" I would add
> 
> ```ts
> const loginUrl = `${this.KialiDetails.url}/${endpoint.replace(/^\//g, "");}`;
> ```
> 
> To be sure that it's going to work. What do you think?

MMM the same way we should be sure that user is going to set *https://kiali-istio-system.apps-crc.testing/kiali* and not  *https://kiali-istio-system.apps-crc.testing/kiali/*  (Without the last */*) this is why I wanted to use the URL class

```ts
new URL("/api/status", "https://kiali-istio-system.apps-crc.testing/").href
'https://kiali-istio-system.apps-crc.testing/api/status'
```

But I don't know how to deal with a pathname with this class ....maybe
```ts
const baseUrl = new URL(this.KialiDetails.url);
const pathName = baseUrl.pathname
const host = baseUrl.href.replace(pathName,"")
const target = new URL(`${pathName}${endpoint}`, host).href
```

aljesusg (approved):
LGFM

Issue #1101, opened 2024-01-23
[RBAC] Watch members and permission policies in the role overview page


debsmita1:
### Describe the bug

The role overview page doesn't watch the members and permission policies

### Expected Behavior

It should poll the members and permission policies associated with a role

### What are the steps to reproduce this bug?

Install and configure the RBAC UI plugin

### GIF

https://github.com/janus-idp/backstage-plugins/assets/22490998/e2d000c9-40a1-4a7f-9dcb-ce05f86498fe

PR #1102, opened 2024-01-23
fix(rbac): watch users and permission-policies

debsmita1:
Resolves:
https://github.com/janus-idp/backstage-plugins/issues/1101

GIF:

https://github.com/janus-idp/backstage-plugins/assets/22490998/388b7c2d-5e64-498c-9e61-06ce536a5c6e

### DIFF
--- a/plugins/rbac/src/hooks/usePermissionPolicies.ts
+++ b/plugins/rbac/src/hooks/usePermissionPolicies.ts
@@ -1,50 +1,80 @@
 import React from 'react';
-import { useAsync, useAsyncRetry } from 'react-use';
+import { useAsyncRetry, useInterval } from 'react-use';
 
 import { useApi } from '@backstage/core-plugin-api';
 
 import { rbacApiRef } from '../api/RBACBackendClient';
 import { getPermissionsData } from '../utils/rbac-utils';
 
-export const usePermissionPolicies = (entityReference: string) => {
+const getErrorText = (
+  policies: any,
+  permissionPolicies: any,
+): { name: number; message: string } | undefined => {
+  if (!Array.isArray(policies) && (policies as Response)?.statusText) {
+    return {
+      name: (policies as Response).status,
+      message: `Error fetching policies. ${(policies as Response).statusText}`,
+    };
+  } else if (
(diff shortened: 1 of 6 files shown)


ciiay:
Hi @debsmita1 This is probably not in this scope, I have tested the scenario where I edited permission policies on the Edit Role page, the role Overview page picks up the change but the RBAC role list page didn't get updated. Looks like the `RolesData.permissions` value is not being watched on the role list page. Let me know if I'm missing anything here.


https://github.com/janus-idp/backstage-plugins/assets/26255262/7cb10a7b-aa7d-463d-800b-9351f3c977b4

plugins/rbac/src/types.ts
```diff
@@ -17,7 +17,7 @@ export type RolesData = {
 
 export type MembersData = {
   name: string;
-  type: 'User' | 'Group';
+  type: string;
```
ciiay, line 20: Just out of curiosity why this change? Is there a value other than `'User' | 'Group'` ?

plugins/rbac/src/hooks/usePermissionPolicies.ts
```diff
@@ -1,50 +1,80 @@
...
-        name: (permissionPolicies as Response)?.status,
-        message: `Error fetching permission policies. ${(
-          permissionPolicies as Response
-        )?.statusText}`,
-      },
+    retry: { policiesRetry, permissionPoliciesRetry },
+    error:
+      (policiesError as Error) ||
+      (permissionPoliciesError as Error) ||
+      getErrorText(policies, permissionPolicies),
```
ciiay, line 78: Do we need type assertion here? `policiesError` and `permissionPoliciesError` are already in Error type.

ciiay (changes requested):
Tested working good. Just left a few small comments.

plugins/rbac/src/hooks/usePermissionPolicies.ts
debsmita1: good catch. Fixed it.

invincibleJai (approved):
/lgtm

PR #1104, opened 2024-01-23
feat(notifications): turn plugins into dynamic ones

mareklibra:
Fixes [FLPATH-877](https://issues.redhat.com//browse/FLPATH-877)

### DIFF
--- a/plugins/notifications-backend/README.md
+++ b/plugins/notifications-backend/README.md
@@ -4,13 +4,23 @@ This Backstage backend plugin provides REST API endpoint for the notifications.
 
 It's backed by a relational database, so far tested with PostgreSQL.
 
+## Deploying as a dynamic plugin
+
+The notifications backend plugin can be loaded either as a static or a dynamic plugin.
+
+To install it as a dynamic plugin, please follow instructions here: https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/dynamic-plugins.md#installing-a-dynamic-plugin-package-in-the-showcase
+
+To install it as a static plugin, several steps are required as described below.
+
+In any case, do not miss the info about configuration and especially about creating entities in Catalog as described below.
+
 ## Getting started
 
 The plugin uses a relational database to persist messages, it has been tested with the SQLite and PostgreSQL.
 
 Upon backend's plugin start, the `backstage_plugin_notifications` database and its tables are created automatically.
 
-### Optional: PostgreSQL
+### Optional: PostgreSQL setup
 
 **To use the Backstage's default SQLite, no specific configuration is needed.**
 
@@ -58,6 +68,8 @@ If PostgreSQL is used, additional configuration in the in the `app-config.yaml` 
     store: memory
 ```
 
+## Deploy as a static plugin
+
 ### Add NPM dependency
 
 ```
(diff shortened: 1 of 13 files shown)


plugins/notifications-backend/src/service/router.ts
```diff
@@ -23,7 +25,14 @@ import { RouterOptions } from './types';
 export async function createRouter(
   options: RouterOptions,
 ): Promise<express.Router> {
-  const { logger, dbConfig, catalogClient } = options;
+  const { logger, database, discovery, config } = options;
+
+  // workaround for creating the database when client is not sqlite
```
gashcrumb, line 30: Is this a normal thing?  Is there an issue for this somewhere?

mareklibra: Nope.  I fixed that in the upstream PR:
https://github.com/backstage/backstage/pull/22099/files#diff-72c8723c8f4d71d407e124d229ed18258167dc4547c9babc7ac3faac3226b413R36

and plan to cherry-pick the change to this PR as well.
But in an independent PR.

plugins/notifications-backend/src/service/router.ts
```diff
@@ -23,7 +25,14 @@ import { RouterOptions } from './types';
...
+  const dbClientWorkaround = await database.getClient();
```
ydayagi, line 31: please don't call it workaround. it only makes people notice it. it is what it is. a db client

plugins/notifications-backend/config.d.ts
ydayagi, line 1: what is the reason for this interface? and why not simply put it in types.ts?

plugins/notifications-backend/README.md
```diff
@@ -80,25 +92,14 @@ import { PluginEnvironment } from '../types';
...
     identity: env.identity,
     logger: env.logger,
     permissions: env.permissions,
     tokenManager: env.tokenManager,
-    dbConfig,
-    catalogClient,
-    externalCallerSecret,
+    database: env.database,
+    discovery: env.discovery,
+    config: env.config,
```
ydayagi, line 102: what is the reason for adding the whole config?

plugins/notifications-backend/README.md
```diff
@@ -80,25 +92,14 @@ import { PluginEnvironment } from '../types';
 export default async function createPlugin(
   env: PluginEnvironment,
 ): Promise<Router> {
-  // workaround for creating the database when client is not sqlite
-  const dbClient = await env.database.getClient()
-  dbClient.destroy()
-
-  const catalogClient = new CatalogClient({ discoveryApi: env.discovery });
```
ydayagi, line 87: why replace this line?

plugins/notifications-backend/README.md
```diff
@@ -80,25 +92,14 @@ import { PluginEnvironment } from '../types';
...
-
   return await createRouter({
     identity: env.identity,
     logger: env.logger,
     permissions: env.permissions,
     tokenManager: env.tokenManager,
-    dbConfig,
-    catalogClient,
-    externalCallerSecret,
+    database: env.database,
```
ydayagi, line 100: what is the reason for passing the database?

plugins/notifications-backend/src/service/router.ts
ydayagi, line 31: this should not be here. the db creation whether required or not (sqlite does not reuqire it) is not part of the plugin. is it a limitation due to dynamic plugin?

plugins/notifications-backend/README.md
mareklibra: The dynamic plugins call the `createRouter` and provide it with the references. We can not customize it as we did in our implementatino of the `createPlugin`

plugins/notifications-backend/src/service/router.ts
mareklibra: yes, to unify the flows

mareklibra: We have actual `dbClient` below. So renaming

Issue #1068, opened 2024-01-16
[RBAC Plugin]: Show error if duplicate permission policies are added


divyanshiGupta:
In Add permissions and policies stepper step, show duplicate error in permission field and disable stepper next button, if user has added duplicate permission policies. Confirm with UX.

PR #1117, opened 2024-01-24
feat(rbac): disable selected permissions

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/1068

New implementation:
Based on UX input this PR now disables selected permissions.

https://github.com/janus-idp/backstage-plugins/assets/20724543/0b958264-9d70-437d-8e37-035cd97244a0

Old implementation:
This PR adds validations to show field level error for permission fields that are not unique for the same plugin.

https://github.com/janus-idp/backstage-plugins/assets/20724543/a00538a8-a3a6-468b-9ad5-a9f8d7d9d9c3

### DIFF
--- a/plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
@@ -149,6 +149,14 @@ export const PermissionPoliciesForm = ({
               }
               onRemove={() => onRowRemove(index)}
               handleBlur={handleBlur}
+              getPermissionDisabled={(permission: string) => {
+                const pluginPermissionPolicies = permissionPoliciesRows.filter(
+                  ppr => ppr.plugin === pp.plugin,
+                );
+                return !!pluginPermissionPolicies.find(
+                  ppp => ppp.permission === permission,
+                );
+              }}
             />
           ))}
           <Button className={classes.addButton} size="small" onClick={onRowAdd}>
--- a/plugins/rbac/src/components/CreateRole/PermissionPoliciesFormRow.tsx
+++ b/plugins/rbac/src/components/CreateRole/PermissionPoliciesFormRow.tsx
@@ -28,6 +28,7 @@ type PermissionPoliciesFormRowProps = {
   onChangePermission: (permission: string, policies?: string[]) => void;
   onChangePolicy: (isChecked: boolean, policyIndex: number) => void;
   handleBlur: React.FocusEventHandler<HTMLInputElement | HTMLTextAreaElement>;
+  getPermissionDisabled: (permission: string) => boolean;
 };
 
 export const PermissionPoliciesFormRow = ({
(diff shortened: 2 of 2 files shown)


debsmita1:
Is this expected ?
Screenshot 2024-01-25 at 11 14 15 AM

parvathyvr:
Hi @divyanshiGupta I wonder if it makes sense to disable a permission if it's already selected from the dropdown and it's present with the policies selected in the list?I think we are doing this behaviour  in the users and groups selection, right?
 If user removes the permission policy row by using the '-' sign, we can enable it in the dropdown. That way, we can avoid the user having duplicates and the situation @divyanshiGupta has showed. WDYT?

divyanshiGupta:
> Hi @divyanshiGupta I wonder if it makes sense to disable a permission if it's already selected from the dropdown and it's present with the policies selected as a row?I think we are doing this behaviour of disabling selected users/groups in the users and groups selection, right? If user removes the permission policy row by using the '-' sign, we can enable it in the dropdown. That way, we can avoid the user having duplicates and the situation @debsmita1 has shown. WDYT?

@parvathyvr Yes I did that for add members dropdown and it makes sense to disable selected permissions for permissions dropdown as well. I have made the changes PTAL.

parvathyvr:
> > Hi @divyanshiGupta I wonder if it makes sense to disable a permission if it's already selected from the dropdown and it's present with the policies selected as a row?I think we are doing this behaviour of disabling selected users/groups in the users and groups selection, right? If user removes the permission policy row by using the '-' sign, we can enable it in the dropdown. That way, we can avoid the user having duplicates and the situation @debsmita1 has shown. WDYT?
> 
> @parvathyvr Yes I did that for add members dropdown and it makes sense to disable selected permissions for permissions dropdown as well. I have made the changes PTAL.

Thanks for making the changes @divyanshiGupta . LGTM! 
Just to make sure-now when the user removes the row of a permission policy using '-', that particular  permission becomes enabled and can be selected from the dropdown again, right?

divyanshiGupta:
> > > Hi @divyanshiGupta I wonder if it makes sense to disable a permission if it's already selected from the dropdown and it's present with the policies selected as a row?I think we are doing this behaviour of disabling selected users/groups in the users and groups selection, right? If user removes the permission policy row by using the '-' sign, we can enable it in the dropdown. That way, we can avoid the user having duplicates and the situation @debsmita1 has shown. WDYT?
> > 
> > 
> > @parvathyvr Yes I did that for add members dropdown and it makes sense to disable selected permissions for permissions dropdown as well. I have made the changes PTAL.
> 
> Thanks for making the changes @divyanshiGupta . LGTM! Just to make sure-now when the user removes the row of a permission policy using '-', that particular permission becomes enabled and can be selected from the dropdown again, right?

Yes @parvathyvr. 

https://github.com/janus-idp/backstage-plugins/assets/20724543/25a8e57a-71a2-4abb-a8bc-8f2ada118ade

invincibleJai (approved):
/lgtm

Issue #1017, opened 2023-12-12
[Topology/Tekton] : pod logs do polling to stream logs even for the ones which are completed


invincibleJai:
### Describe the bug

pod logs do polling to stream logs even for the ones which are completed

<img width="1724" alt="image" src="https://github.com/janus-idp/backstage-plugins/assets/5129024/b44eee76-9dec-400e-8db3-ec05b3d06744">

### Expected Behavior 

If pod logs are completed and then polling shouldn't happen

### What are the steps to reproduce this bug?

1. Install and configure Topology or Tekton plugin
2. check for pod logs

### Versions of software used and environment

@janus-idp/backstage-plugin-tekton: 3.1.14
@janus-idp/backstage-plugin-topology: 1.16.14

PR #1122, opened 2024-01-25
fix(topology/tekton): stop polling after log streaming completes

debsmita1:
Resolves:
https://github.com/janus-idp/backstage-plugins/issues/1017

### DIFF
--- a/plugins/tekton/README.md
+++ b/plugins/tekton/README.md
@@ -119,7 +119,7 @@ The Tekton plugin enables you to visualize the `PipelineRun` resources available
    yarn workspace app add @janus-idp/backstage-plugin-tekton
    ```
 
-1. Enable the PipelineRun list in the **CI/CD** tab on the entity view page.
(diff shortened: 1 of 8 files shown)


plugins/shared-react/src/utils/pod.ts
```diff
@@ -0,0 +1,6 @@
+import { V1Pod } from '@kubernetes/client-node';
+
+export const shouldStopPolling = (pod: V1Pod) =>
+  pod?.status?.phase === 'Succeeded' ||
+  pod?.status?.phase === 'Failed' ||
+  pod?.status?.phase === 'Running';
```
karthikjeeyar, line 6: I think we should not stop polling the data when the pipelinerun is in running state so we can see the logs streaming.

```suggestion
  pod?.status?.phase === 'Failed';
```

karthikjeeyar (approved):
This is a nice improvement. :tada:

/lgtm

PR #1124, opened 2024-01-25
fix(orchestrator): improve workflow instances page table size

batzionb:
[FLPATH-887](https://issues.redhat.com/browse/FLPATH-887)
Currently the default page size is 5, when often the workflow runs length are more than 5 and there's a lot of space left on the page that is wasted.
To improve this, it will show a table with 20 rows and pagination if there are more than 20 rows, otherwise it'll show a table without pagination.
This behavior matches the [backstage catalog tables](https://github.com/backstage/backstage/blob/v1.14.0/plugins/catalog/src/components/CatalogTable/CatalogTable.tsx#L228).
In order to reuse the CatalogPage and avoid duplicating this code, this would be possible when there will be better integration of backstage with the orchestrator and the workflow instances will be backstage entities.

Short list:
image

Long list:
[Screencast from 25-01-24 13:53:58.webm](https://github.com/janus-idp/backstage-plugins/assets/22211154/a27ecfd3-d4bc-4417-9aa5-b64a321ccf0e)

### DIFF
--- a/plugins/orchestrator/src/__fixtures__/fakeLongProcessInstanceList.ts
+++ b/plugins/orchestrator/src/__fixtures__/fakeLongProcessInstanceList.ts
@@ -1,0 +1,65 @@
+import {
+  ProcessInstance,
+  ProcessInstanceState,
+  WorkflowCategory,
+} from '@janus-idp/backstage-plugin-orchestrator-common';
+
+import { fakeWorkflowOverviewList } from './fakeWorkflowOverviewList';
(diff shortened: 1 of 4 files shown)


caponetto:
Hey @batzionb, please revert the changes outside the `orchestrator` package.

jkilzi:
This PR can be closed in favour of https://github.com/janus-idp/backstage-plugins/pull/1127

PR #1126, opened 2024-01-25
chore(orchestrator): update app-config.janus-idp.yaml to use dataIndex

mareklibra:


plugins/orchestrator-backend/app-config.janus-idp.yaml
```diff
@@ -1,4 +1,5 @@
 orchestrator:
-  sonataFlowService:
-    baseUrl: http://localhost
-    port: 8080
+  dataIndexService:
+    url: http://sonataflow-platform-data-index-service.sonataflow-infra
```
caponetto, line 3: Not sure about using this very specific URL but it should be overridden by the user's config anyway.

caponetto (approved):

PR #1133, opened 2024-01-26
[WIP]feat(rbac): turn rbac plugin into a dynamic plugin

ciiay:
For [RHIDP-1050](https://issues.redhat.com/browse/RHIDP-1050): Add/enable RBAC UI plugin to showcase/RHDH and test the RBAC UI flow


.turbo-cookie
divyanshiGupta, line 1: This file is not needed

packages/backend/src/plugins/kubernetes.ts
divyanshiGupta, line 1: Looks like this file got pushed accidentally.

yarn.lock
divyanshiGupta, line 1: Looks like while rebasing some conflict headers/separators were not removed. I will suggest that you discard these changes and pull yarn.lock again from main and then do `yarn install` to correctly generate any changes and avoid any issues that might occur because of a bad yarn.lock file.

plugins/rbac-backend/dist-dynamic/package.json
divyanshiGupta, line 1: Do we need this file?

plugins/rbac-backend/dist-dynamic/yarn.lock
divyanshiGupta, line 1: Same as above. Do we need this?

plugins/rbac-backend/dist-dynamic/package.json
ciiay: No, the entire `plugins/rbac-backend/dist-dynamic/` is not needed. Added accidentally. It's been removed.

ciiay:
@divyanshiGupta Thanks for the review. I addressed the comments you left and removed changes in `yarn.lock` as I didn't change anything in `package.json`. Please confirm this updated PR, thanks.

Issue #1108, opened 2024-01-24
[Kiali] Overview not load the metrics first time.


aljesusg:
### Describe the bug

When you go to Kiali tab you can see the information related with namespaces but the metrics are not loaded 
![image](https://github.com/janus-idp/backstage-plugins/assets/3019213/9ef86876-aee2-4141-95c5-49957cd1f99a)

They are loaded after refresh or select a dropdown option so it seems that there is a problem with react.useEffect here

![image](https://github.com/janus-idp/backstage-plugins/assets/3019213/a7115958-bb4c-4d2e-b881-f348431716a9)

### Expected Behavior

See the metrics and health first time

### What are the steps to reproduce this bug?

1. Go to Kiali tab
2. Check that health and metrics are not loaded

### Versions of software used and environment

kilai plugin 1.8.1

PR #1139, opened 2024-01-30
fix(kiali): show username when auth is anonymous

aljesusg:
Fix https://github.com/janus-idp/backstage-plugins/issues/1108

- [X] Aligned toolbar
- [X] Fix metrics first load
- [X] Set anonymous by default due if there is not login kiali will fail 
- [X] Fix kiali Header

# Toolbar

image


josunect (approved):
LGTM 

image

PR #1143, opened 2024-01-30
chore: sync yarn.lock

invincibleJai:
Running yarn install on main results in the following diff

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -249,15 +249,6 @@
     "@aws-sdk/types" "^3.222.0"
     tslib "^1.11.1"
 
-"@aws-crypto/sha256-js@^5.0.0":
-  version "5.2.0"
-  resolved "https://registry.yarnpkg.com/@aws-crypto/sha256-js/-/sha256-js-5.2.0.tgz#c4fdb773fdbed9a664fc1a95724e206cf3860042"
-  integrity sha512-FFQQyu7edu4ufvIZ+OadFpHHOt+eSTBaYaki44c+akjg7qZg9oOQeLlk77F6tSYqjDAFClrHJk9tMf0HdVyOvA==
-  dependencies:
-    "@aws-crypto/util" "^5.2.0"
-    "@aws-sdk/types" "^3.222.0"
-    tslib "^2.6.2"
-
 "@aws-crypto/supports-web-crypto@^3.0.0":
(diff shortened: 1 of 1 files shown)


caponetto:
Thanks @invincibleJai!
I wonder if we could have a check after [this step](https://github.com/janus-idp/backstage-plugins/blob/main/.github/actions/build/action.yaml#L19)
Something like this
```yaml
    - name: "Check yarn.lock"
      shell: bash
      run: |
        echo "Check yarn.lock"
        git diff yarn.lock;
        [ "0" == "$(git diff yarn.lock | wc -l | tr -d ' ')" ]
```
The build would break if there is a diff in the lock.

karthikjeeyar (approved):
/lgtm

invincibleJai:
> Thanks @invincibleJai! I wonder if we could have a check after [this step](https://github.com/janus-idp/backstage-plugins/blob/main/.github/actions/build/action.yaml#L19) Something like this
> 
> ```yaml
>     - name: "Check yarn.lock"
>       shell: bash
>       run: |
>         echo "Check yarn.lock"
>         git diff yarn.lock;
>         [ "0" == "$(git diff yarn.lock | wc -l | tr -d ' ')" ]
> ```
> 
> The build would break if there is a diff in the lock.

yeah we should be adding this

@kadel ^^ and if you can approve the PR

gashcrumb:
> > Thanks @invincibleJai! I wonder if we could have a check after [this step](https://github.com/janus-idp/backstage-plugins/blob/main/.github/actions/build/action.yaml?rgh-link-date=2024-01-30T13%3A58%3A15Z#L19) Something like this
> > ```yaml
> >     - name: "Check yarn.lock"
> >       shell: bash
> >       run: |
> >         echo "Check yarn.lock"
> >         git diff yarn.lock;
> >         [ "0" == "$(git diff yarn.lock | wc -l | tr -d ' ')" ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The build would break if there is a diff in the lock.
> 
> yeah we should be adding this
> 
> @kadel ^^ and if you can approve the PR

BTW I created https://github.com/janus-idp/backstage-plugins/issues/1169 and will shortly also create a similar issue for backstage-showcase.

PR #1163, opened 2024-02-02
fix(rbac): set up higher jest timeout for rbac db tests

AndrienkoAleksandr:


AndrienkoAleksandr:
I re-executed tests five times and each attempt is green.

invincibleJai (approved):
/lgtm

PR #1168, opened 2024-02-05
fix: add missing alpha  dynamic plugin entry points

davidfestal:
When embedding alpha modules, the embedded module dependencies were not hoisted to the embedding module.

This PR fixes this issue.


schultzp2020 (approved):
/lgtm
/approve

PR #1171, opened 2024-02-05
fix(rbac): csv updates no longer require server restarts

### LINKED ISSUE #908: RBAC backend plugin - policy change requires server restart

PatAKnight:
## Description

This PR introduces new functionality to facilitate the reloading of permission policies sourced from the CSV file. The issue at hand was the necessity of restarting the server every time there was an update to the permission policies in the CSV, which was quite inconvenient. The added feature enables the dynamic reloading of permission policies and roles directly from the CSV.

It leverages information provided by the permission framework (entity reference, permission, policy). This approach allows us to focus solely on roles associated with the entity reference in the CSV, subsequently retrieving permissions linked to that role, matching the given permission and policy. The aim is to enhance efficiency by selectively reloading relevant portions of the file, rather than the entire dataset.

There are duplication checks within this proposed PR that checks if there are duplicates within just the CSV file as well as duplication coming from the REST API and app configuration file. This will throw warnings to the console for the Admin to deal with.

## Fixes

- Fixes #908 

1. Create an RBAC permission policy CSV file
```csv
g, user:default/<your-user>, role:default/test

p, role:default/test, catalog-entity, read, allow
```
2. Update app-config to include your CSV file
```yaml
permission:
  enabled: true
  rbac:
    policies-csv-file: ./some/path/<permission-policy>.csv
```
3. Start the server and notice that you can read catalog entities
4. update the read permission in the CSV file from `allow` to `deny`
5. Notice that you can no longer read the catalog entities

***

One caveat, as of right now there is no full synchronization of the CSV file. I originally elected to avoid this approach to ensure that we are as efficient as possible. The problem with this is that the RBAC Frontend plugin will occasionally be out of sync with the CSV file. This is apparent whenever you create a new role or permissions that you are not directly attached to. The mechanism for reloading the permissions and roles will only look at the user that is being checked and the permission that will will be looking to authorize. Which will result in the frontend only being updated whenever a particular user makes an action.

To solve this problem will result in either a change in how our GET endpoints are called or adding an additional endpoint. 

We could update the GET endpoints to refresh the CSV file to ensure that we are as up to date as possible (costly as this will reload the entire CSV file whenever an admin navigates to the Administrators page) 

Or we can add an additional endpoint that could be used to refresh the page on demand (less costly as this will only reload the CSV file when called, however will be out of sync until a refresh occurs)

***

### DIFF
--- a/plugins/rbac-backend/src/__fixtures__/data/valid-csv/rbac-policy.csv
+++ b/plugins/rbac-backend/src/__fixtures__/data/valid-csv/rbac-policy.csv
@@ -7,4 +7,6 @@ p, role:default/catalog-writer, catalog-entity, read, allow
 p, role:default/catalog-writer, catalog.entity.create, use, allow
 p, role:default/catalog-deleter, catalog-entity, delete, deny
 
-p, user:default/known_user, test.resource.deny, use, allow
+p, role:default/known_role, test.resource.deny, use, allow
+
+g, user:default/known_user, role:default/known_role
--- a/plugins/rbac-backend/src/__fixtures__/data/valid-csv/rbac-policy.csv
+++ b/plugins/rbac-backend/src/__fixtures__/data/valid-csv/rbac-policy.csv
@@ -1,0 +1,10 @@
+g, user:default/guest, role:default/catalog-writer
+g, user:default/guest, role:default/catalog-reader
+g, user:default/guest, role:default/catalog-deleter
+
+p, role:default/catalog-writer, catalog-entity, update, allow
+p, role:default/catalog-writer, catalog-entity, read, allow
+p, role:default/catalog-writer, catalog.entity.create, use, allow
+p, role:default/catalog-deleter, catalog-entity, delete, deny
+
+p, user:default/known_user, test.resource.deny, use, allow
--- a/plugins/rbac-backend/src/file-permissions/csv.ts
+++ b/plugins/rbac-backend/src/file-permissions/csv.ts
@@ -15,12 +15,60 @@ import {
   validatePolicy,
 } from '../service/policies-validation';
 
+const addPolicy = async (
+  policy: string[],
+  enf: EnforcerDelegate,
+  policyMetadataStorage: PolicyMetadataStorage,
+  logger: Logger,
+): Promise<void> => {
+  const source = await policyMetadataStorage?.findPolicyMetadata(policy);
+
+  if (!(await enf.hasPolicy(...policy))) {
+    await enf.addPolicy(policy, 'csv-file');
+  } else if (source?.source !== 'csv-file') {
+    logger.warn(
+      `Duplicate policy: ${policy[0]}, ${policy[1]}, ${policy[2]} found with the source ${source?.source}`,
+    );
+  }
+};
+
+const removeEnforcerPolicies = async (
+  enforcerPolicies: string[][],
+  tempEnf: Enforcer,
+  enf: EnforcerDelegate,
+  policyMetadataStorage: PolicyMetadataStorage,
+): Promise<void> => {
+  for (const policy of enforcerPolicies) {
+    const enfPolicySource =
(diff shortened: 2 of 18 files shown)


plugins/rbac-backend/src/__fixtures__/data/valid-csv/rbac-policy.csv
```diff
@@ -0,0 +1,10 @@
+g, user:default/guest, role:default/catalog-writer
+g, user:default/guest, role:default/catalog-reader
+g, user:default/guest, role:default/catalog-deleter
+
+p, role:default/catalog-writer, catalog-entity, update, allow
+p, role:default/catalog-writer, catalog-entity, read, allow
+p, role:default/catalog-writer, catalog.entity.create, use, allow
+p, role:default/catalog-deleter, catalog-entity, delete, deny
+
+p, user:default/known_user, test.resource.deny, use, allow
```
AndrienkoAleksandr, line 10: Maybe we should not promote direct assign permission to the role in the tests.

PatAKnight: Yes, I agree

AndrienkoAleksandr:
I think we need to make this feature configurable.

plugins/rbac-backend/src/helper.ts
```diff
@@ -35,3 +38,10 @@ export async function removeTheDifference(
     await enf.removeGroupingPolicy(role, source, true);
   }
 }
+
+export const transformArraytoPolicy = (
```
AndrienkoAleksandr, line 42: Minor: transformArraytoPolicy => transformArrayToPolicy

plugins/rbac-backend/src/service/permission-policy.ts
```diff
@@ -343,4 +292,23 @@ export class RBACPermissionPolicy implements PermissionPolicy {
...
   }
+
+  private isAuthorized = async (
+    userIdentity: string,
+    permission: string,
+    action: string,
+  ): Promise<boolean> => {
+    const filter: string[] = [userIdentity, permission, action];
+    if (this.policiesFile) {
+      await loadFilteredCSV(
```
AndrienkoAleksandr, line 303: Could you apply try catch with log error, just in case unexpected error?

plugins/rbac-backend/src/file-permissions/csv.ts
```diff
@@ -0,0 +1,365 @@
...
+
+      enforcerPolicies = enforcerPolicies.filter(
+        policyCheck => !isEqual(policyCheck, effectFlipPolicy),
+      );
+    }
+
+    await addPolicy(policy, enf, policyMetadataStorage, logger);
+  }
+
+  await removeEnforcerPolicies(
```
AndrienkoAleksandr, line 168: Maybe we should call this method on the beginning function?

plugins/rbac-backend/src/service/permission-policy.ts
PatAKnight: Question, wouldn't the above [try catch](https://github.com/janus-idp/backstage-plugins/blob/eb6042c38c9993e0ff43086e81c039aa91d75a71/plugins/rbac-backend/src/service/permission-policy.ts#L219:L277) handle any potential errors that might be thrown from `loadFilteredCSV` since `authorize` is called within that try catch?

AndrienkoAleksandr (approved):

AndrienkoAleksandr:
LGTM

PR #1178, opened 2024-02-07
feat(rbac): backend part - store role description to the database 

AndrienkoAleksandr:

https://github.com/janus-idp/backstage-showcase/issues/898

### DIFF
--- a/plugins/rbac-backend/src/helper.test.ts
+++ b/plugins/rbac-backend/src/helper.test.ts
@@ -1,0 +1,250 @@
+import { RoleMetadataDao } from './database/role-metadata';
+import {
+  deepSortedEqual,
+  metadataStringToPolicy,
+  policiesToString,
+  policyToString,
+  removeTheDifference,
+  transformArrayToPolicy,
+} from './helper';
+// Import the function to test
+import { EnforcerDelegate } from './service/enforcer-delegate';
+
+describe('helper.ts', () => {
+  describe('policyToString', () => {
+    it('should convert permission policy to string', () => {
+      const policy = [
+        'user:default/some-user',
+        'catalog-entity',
+        'read',
+        'allow',
+      ];
+      const expectedString =
+        '[user:default/some-user, catalog-entity, read, allow]';
+      expect(policyToString(policy)).toEqual(expectedString);
+    });
+  });
+  describe('policiesToString', () => {
+    it('should convert one permission policy to string', () => {
+      const policies = [
+        ['user:default/some-user', 'catalog-entity', 'read', 'allow'],
+      ];
+      const expectedString =
+        '[[user:default/some-user, catalog-entity, read, allow]]';
+      expect(policiesToString(policies)).toEqual(expectedString);
+    });
+
+    it('should convert empty permission policy array to string', () => {
+      const policies = [[]];
+      const expectedString = '[[]]';
+      expect(policiesToString(policies)).toEqual(expectedString);
+    });
+  });
+
+  describe('metadataStringToPolicy', () => {
+    it('parses a permission policy string', () => {
+      const policy = '[user:default/some-user, catalog-entity, read, allow]';
+      const expectedPolicy = [
+        'user:default/some-user',
+        'catalog-entity',
+        'read',
+        'allow',
+      ];
+      expect(metadataStringToPolicy(policy)).toEqual(expectedPolicy);
+    });
+
+    it('parses a grouping policy', () => {
+      const policy = '[user:default/some-user, role:default/dev]';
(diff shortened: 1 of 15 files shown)


AndrienkoAleksandr:
> Hmm... I found a bug: update doesn't update description, when I changed only role description and provided role without changes. I will fix it.

Done.

plugins/rbac-backend/src/helper.test.ts
```diff
@@ -0,0 +1,249 @@
+import { RoleMetadataDao } from './database/role-metadata';
+import {
+  deepSortedEqual,
+  metadataStringToPolicy,
+  policiesToString,
+  policyToString,
+  removeTheDifference,
+  transformArrayToPolicy,
+} from './helper'; // Import the function to test
```
PatAKnight, line 9: Looks like you are failing the prettier check, you can fix it by running `yarn prettier:fix`

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -364,9 +366,12 @@ export class PolicesServer {
...
 
-      for (const role of roles) {
-        await this.enforcer.addOrUpdateGroupingPolicy(role, 'rest', false);
-      }
+      const metadata: RoleMetadataDao = {
+        roleEntityRef: roleRaw.name,
+        source: 'rest',
+        description: roleRaw.metadata?.description ?? '',
+      };
+      await this.enforcer.addGroupingPolicies(roles, metadata);
```
PatAKnight, line 374: I am fine with leaving this change.

Using just the below in a PUT request will change the source from `legacy` to `rest`.
```
{
    "oldRole":
        {
            "memberReferences": ["group:default/group-ab"],
            "name": "role:default/group-ab"
        },
    "newRole":
        {
            "memberReferences": ["group:default/group-ab"],
            "name": "role:default/group-ab"
        }
}
```

I think it would be best to capture all of the information around legacy in the documentation. Just haven't figured out where the best place would be. Or even if it would be worth doing at this moment.

plugins/rbac-backend/src/service/policies-rest-api.ts
```diff
@@ -208,9 +212,7 @@ export class PolicesServer {
 
       const processedPolicies = await this.processPolicies(policyRaw);
 
-      for (const policy of processedPolicies) {
-        await this.enforcer.addOrUpdatePolicy(policy, 'rest', false);
-      }
+      await this.enforcer.addPolicies(processedPolicies, 'rest');
```
PatAKnight, line 215: Same as the comment on `await this.enforcer.addOrUpdateGroupingPolicy(role, 'rest', false);`

PatAKnight (commented):
It is up to you on whether or not you want to update the [API documentation](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/apis.md) for role descriptions in this PR. Honestly, we need to update this documentation anyways and could do it under a separate PR if that would work best for you. It got out of sync whenever I made the changes to allow for multiple permission policies to be added at once.

PatAKnight:
Also, for some reason, I was unable to test this with your frontend part. The backend part worked just fine for me, but the descriptions were not showing up in the frontend part. Not certain what was going on there.

plugins/rbac-backend/src/plugin.ts
```diff
@@ -53,7 +53,12 @@ export const rbacPlugin = createBackendPlugin({
         tokenManager,
       }) {
         const winstonLogger = loggerToWinstonLogger(logger);
-
+        const isRbacPluginEnabled =
+          config.getOptionalBoolean(`permission.enabled`);
+        if (!isRbacPluginEnabled) {
+          winstonLogger.info(`rbac backend plugin is disabled`);
+          return;
+        }
```
PatAKnight, line 61: I like the idea of this change. However, I don't believe it is working like we think it should as the RBAC plugin is still building.

If I set `permission.enabled: false` and check the database, I can see that the permissions and roles were added to the database.

PatAKnight (commented):
Everything with the database works fine for me. Tested by building the database using main and then checking out and rerunning your PR

plugins/rbac-backend/src/plugin.ts
PatAKnight: I am not really finding anywhere that would be a good place for this change to go that will prevent the RBAC plugin from building. We might have to look upstream to see how the permission framework is enabled / disabled.

AndrienkoAleksandr: Yes, you're right, I should revert this change.

AndrienkoAleksandr: We need separated pr for this purpose.

Issue #1063, opened 2024-01-15
Add support for the new backend system to  the RBAC backend plugin.


davidfestal:
Add support for the new backend system to  the RBAC backend plugin.
This implies:
- Creating and providing the `BackendFeatureFactory` default export
- Thinking about necessary extension points
- Having a way to gather the `pluginID` of `BackendFeature`s created with `createBackendPlugin()`, in order to pass the list of pluginIDs to the RBAC plugin (as an extension ?).

davidfestal:
related upstream backstage RFC: https://github.com/backstage/backstage/issues/17538

PR #1179, opened 2024-02-07
feat: add support for the new backend system to the `rbac-backend` plugin

davidfestal:
Add support for the new backend system to the rbac backend plugin.

This fixes issue #1063


AndrienkoAleksandr (approved):
Looks good. And I tested with backstage-plugins - it works.

PR #1180, opened 2024-02-07
feat(kiali): add KialiPage Component

aljesusg:
This PR allow to add a KialiPage full integration in the menu and let use it like Dynamic Plugin

image


josunect (approved):
LGTM

image

PR #1182, opened 2024-02-07
feat(orchestrator): add OpenAPI v2 implementations

gciavarrini:
Add v2 endpoints' implementation and unit tests.


plugins/orchestrator-common/scripts/openapi.sh
ydayagi, line 1: I think that this file already exists in another PR

plugins/orchestrator-common/src/auto-generated/api/definition.ts
ydayagi, line 1: I think that this file already exists in another PR

plugins/orchestrator-common/src/auto-generated/api/models/schema.ts
ydayagi, line 1: I think that this file already exists in another PR

ydayagi (approved):

plugins/orchestrator-common/scripts/openapi.sh
gciavarrini: This is a draft PR. It depends on PR #1123, as per description. This means that it is built on top of it.
Once PR #1123 is merged, i can rebase this PR on top of main, so you won't see this change.

plugins/orchestrator-common/src/auto-generated/api/definition.ts
gciavarrini: This is a draft PR. It depends on PR #1123, as per description. This means that it is built on top of it.
Once PR #1123 is merged, i can rebase this PR on top of main, so you won't see this change.

plugins/orchestrator-backend/dist-dynamic/yarn.lock
ydayagi, line 1: i think we should not push this file

ydayagi (changes requested):

plugins/orchestrator-backend/dist-dynamic/yarn.lock
gciavarrini: It must be pushed IMO.
Every time you change the package.json from the backend plugin, you need to regenerate the package.json and yarn.lock files inside the dist-dynamic folder

batzionb: @ydayagi why do you think this change should not be pushed?
What's the difference between this case and any other nodejs application?
The yarn.lock needs to be updated to insure that when the application is installed it gets the correct versions of all the node_modules tree

caponetto:
@batzionb is taking care of the review.

plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts
```diff
@@ -0,0 +1,215 @@
...
+      return 'Error';
+    case ProcessInstanceState.Completed.valueOf():
+      return 'Completed';
+    case ProcessInstanceState.Aborted.valueOf():
+      return 'Aborted';
+    case ProcessInstanceState.Suspended.valueOf():
+      return 'Suspended';
+    default:
+      throw new Error(
+        'state is not one of the vlalues of type ProcessInstanceStatusDTO',
```
batzionb, line 132:
```suggestion
        'state is not one of the values of type ProcessInstanceStatusDTO',
```

plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts
```diff
@@ -0,0 +1,215 @@
...
+  WorkflowSpecFile,
+  WorkflowSpecFileDTO,
+} from '@janus-idp/backstage-plugin-orchestrator-common';
+
+// Mapping functions
+export function mapToWorkflowOverviewDTO(
+  overview: WorkflowOverview,
+): WorkflowOverviewDTO {
+  return {
+    workflowId: overview.workflowId,
```
batzionb, line 29: you can use object spread to make this shorter
return {
  ...overview,
  category: mapWorkflowCategoryDTOFromString(overview.category),
}
same in other mapping functions

plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts
```diff
@@ -0,0 +1,215 @@
...
+    category: mapWorkflowCategoryDTOFromString(overview.category),
+    avgDurationMs: overview.avgDurationMs,
+    description: overview.description,
+  };
+}
+
+export function mapWorkflowCategoryDTOFromString(
+  category?: string,
+): WorkflowCategoryDTO {
+  const lowerCase = category?.toLocaleLowerCase();
```
batzionb, line 43: how about 
return category?.toLocalLowerCase === "assessment" ? "assessment" : "infrastructure";

gciavarrini: done

plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts
gciavarrini: thanks! fixed

plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts
gciavarrini: I agree that it looks nicer, but if it's ok for you, I'd prefer to do these changes in a following PR to be sure I won't break anything (I'll add more unit tests).

batzionb (approved):

PR #1185, opened 2024-02-07
feat(rbac): use relative links

### LINKED ISSUE #1165: RBAC - Discover and implement adjustments needed for the plugin to work in the showcase provided "Administration" tab
### What do you want to improve?

Adjust the exported components from the RBAC fronttend plugin

### What is the current behavior?

Currently the RBAC frontend plugin exposes one component that contains the entire page.  To improve the usage of these views for a dynamic front end it would be good to add some additional exposed components, as currently when trying to mount the available page component:

image

### What will the new behavior be?

The existing page component will still be available, and an additional component will be available that contains only the RBAC plugin components excluding the surrounding page and header components.

gashcrumb:
This change updates the links between components from absolute links to relative links to better support being run from a different route root. This commit also adds a property to help better integrate this plugin into the backstage-showcase Administration page while still retaining it's existing functionality.  Some examples:

image

image

image

This fixes #1165 

Also note that this needs https://github.com/janus-idp/backstage-showcase/pull/955 to fully test this locally.

### DIFF
--- a/plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
+++ b/plugins/rbac/src/components/CreateRole/CreateRolePage.tsx
@@ -37,7 +37,7 @@ export const CreateRolePage = () => {
       resourceRef={catalogEntityReadPermission.resourceType}
     >
       <Page themeId="tool">
-        <Header title="Create role" type="RBAC" typeLink="/rbac" />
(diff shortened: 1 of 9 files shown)


gashcrumb:
@divyanshiGupta @debsmita1 @invincibleJai FYI

Issue #1109, opened 2024-01-24
[RBAC Plugin] Create/edit role form add users/groups table showing 1 member for a group that has 0


divyanshiGupta:
### Describe the bug
Create/edit role form add users/groups table showing 1 member for a group that has 0. 
If there any members already in the group then it shows the correct number.

<img width="1792" alt="Screenshot 2024-01-24 at 3 33 35 PM" src="https://github.com/janus-idp/backstage-plugins/assets/20724543/2e56ef4e-089b-4622-b283-4f0e1e1b2099">

<img width="1792" alt="Screenshot 2024-01-24 at 3 33 55 PM" src="https://github.com/janus-idp/backstage-plugins/assets/20724543/f64cac6a-8981-44b9-8236-f1d69f7386eb">

### Expected Behavior
The table should show 0 members if there are no members in the group.

PR #1187, opened 2024-02-08
fix(rbac): show 0 if no members in a group

divyanshiGupta:
Fixes: https://github.com/janus-idp/backstage-plugins/issues/1109

Screenshot 2024-02-08 at 2 26 35 PM


debsmita1:
verified this locally. Would be great if you could add a test. Thanks!

gashcrumb (approved):

PR #1189, opened 2024-02-08
fix(rbac): add test for 0 members in group

divyanshiGupta:
Add test for https://github.com/janus-idp/backstage-plugins/pull/1187


debsmita1:
Thanks @divyanshiGupta  !

PR #1196, opened 2024-02-11
fix(orchestrator): the instance details card content is cropped

jkilzi:
Addresses:
- https://issues.redhat.com/browse/FLPATH-959

[Screencast from 2024-02-11 15-21-42.webm](https://github.com/janus-idp/backstage-plugins/assets/77008341/2eaec5c0-265d-4be7-bb91-198eb37f7d0b)

### DIFF
--- a/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
+++ b/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 
 import { Content, InfoCard, Link } from '@backstage/core-components';
 import { PathParams, RouteFunc, useRouteRef } from '@backstage/core-plugin-api';
-import { AboutField } from '@backstage/plugin-catalog';
 
 import { Grid, makeStyles } from '@material-ui/core';
 import moment from 'moment';
@@ -11,18 +10,16 @@ import {
   AssessedProcessInstance,
   parseWorkflowVariables,
   ProcessInstance,
-  ProcessInstanceStateValues,
   QUERY_PARAM_ASSESSMENT_INSTANCE_ID,
 } from '@janus-idp/backstage-plugin-orchestrator-common';
 
 import { VALUE_UNAVAILABLE } from '../constants';
-import { executeWorkflowRouteRef, workflowInstanceRouteRef } from '../routes';
-import { capitalize } from '../utils/StringUtils';
+import { executeWorkflowRouteRef } from '../routes';
 import { buildUrl } from '../utils/UrlUtils';
 import { EditorViewKind, WorkflowEditor } from './WorkflowEditor';
-import { WorkflowInstanceStatusIndicator } from './WorkflowInstanceStatusIndicator';
 import { WorkflowProgress } from './WorkflowProgress';
 import { WorkflowRunDetail, WorkflowSuggestion } from './WorkflowRunDetail';
+import { WorkflowRunDetails } from './WorkflowRunDetails';
 import { WorkflowVariablesViewer } from './WorkflowVariablesViewer';
 
 export const mapProcessInstanceToDetails = (
@@ -87,10 +84,10 @@ const getNextWorkflows = (
 
 const useStyles = makeStyles(_ => ({
   topRowCard: {
-    height: '258px',
+    height: '20rem',
   },
   middleRowCard: {
-    height: 'calc(2 * 258px)',
+    height: 'calc(2 * 20rem)',
   },
   bottomRowCard: {
     height: '100%',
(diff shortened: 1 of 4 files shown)


jkilzi:
> I suggest to add to storybook a story with a long description so we can easily see how it behaves with a long description

That's a good idea, I rather add the storybook in a separate PR in order not to delay the fix

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -87,10 +84,10 @@ const getNextWorkflows = (
 
 const useStyles = makeStyles(_ => ({
   topRowCard: {
-    height: '258px',
```
batzionb, line 90: change this to rem

batzionb:
Please include a screenshot with a multi line description to make sure this fix resolves all the size issues of the card

jkilzi:
@batzionb I added the story as requested
You can check it by running SB locally through this URL: http://localhost:6006/?path=/story/orchestrator-workflowdetails--story-1

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
jkilzi: It turns out that the `rem`s are set to be a scale of 10 pixels, so changing this to `2.58rem` will yield the same result.

batzionb:
In the workflow definition page, the behavior for handling a long description is to have a scrollbar
In the instance page, it's handled with ellipsis
I believe it should be the same experience
[Screencast from 13-02-24 14:29:41.webm](https://github.com/janus-idp/backstage-plugins/assets/22211154/a52894cd-7608-44b6-b17e-0a07fcb6513c)

jkilzi:
> In the workflow definition page, the behavior for handling a long description is to have a scrollbar In the instance page, it's handled with ellipsis I believe it should be the same experience [Screencast from 13-02-24 14:29:41.webm](https://github.com/janus-idp/backstage-plugins/assets/22211154/a52894cd-7608-44b6-b17e-0a07fcb6513c)

Agreed, but such change is out of scope of this PR since it is addressing a bug. Please open a follow up task for it.

batzionb:
I cloned the repository and tested it 
I opened the WorkflowDetails very long description story and I don't see the assessed by part
image

Only after I add "overflow: auto" I can see it with a scroll bar

batzionb:
On a small screen I get this
image

I would expect the grid to be responsive, when the screen gets smaller there should be less columns

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
batzionb: actually, rems are relative to the default font size and pixels are absolute
the default font size is 16px so 1 rem is by default 16px
However, if you change the default font size, using pixels, you won't get the scaling relative to the font size. 
So this is an accessitiliby issue.
For example, I changed my font size in chrome settings to be very large.
This is the details card I see:
image

jkilzi:
> On a small screen I get this image
> 
> I would expect the grid to be responsive, when the screen gets smaller there should be less columns

TTBOMK users won't access the UI from tablets or small screens. IMO this comment applies to all the screens we have implemented but requires further discussion. I would rather tackle the issues related to responsive design in a dedicated ticket because there's need to review all the screens and define along with the UXD team the desired layout for each case.

batzionb:
> > On a small screen I get this image
> > I would expect the grid to be responsive, when the screen gets smaller there should be less columns
> 
> TTBOMK users won't access the UI from tablets or small screens. IMO this comment applies to all the screens we have implemented but requires further discussion. I would rather tackle the issues related to responsive design in a dedicated ticket because there's need to review all the screens and define along with the UXD team the desired layout for each case.

I guess the words small screen were misleading, I was talking about my laptop when the window is slightly minimized.
responsiveness is something you get by default when using material ui, typically you won't need to be worried about it. You can see in other places in the application we use grids, you get the responsiveness for free. for some reason, we don't get it here, I'm not sure why.

jkilzi:
@batzionb I'm attaching a new suggestion on how to address this issue differently (spoiler alert, I've changed the layout a bit).
- [Part 1](https://github.com/janus-idp/backstage-plugins/assets/77008341/df9f883a-2a8d-46db-8de6-57c5e81d7c82)
- [Part 2](https://github.com/janus-idp/backstage-plugins/assets/77008341/edd4bfea-48b3-48fe-b6f0-cf07553683db)

Please let me know if this aligns better with the design you have in mind :pray:

batzionb (approved):

PR #1198, opened 2024-02-12
feat(kiali): workloads details page overview tab

josunect:
Kiali page for workloads details.


aljesusg:
It would be great make the links blue like kiali standalone. Users need to know that there is a link to another view
image


The overview of a resource looks great but I don't know if you added the breadcumb ind ev env or not I am going to launch it

image

aljesusg:
Yes...we need toa dd the breadcumb and add some screenshot in the README but looks great

image

josunect:
> Yes...we need toa dd the breadcumb and add some screenshot in the README but looks great
> 
> image

I've added the breadcrumb:

image

Still need to fix a couple of things, but, what do you think about the styles?

aljesusg (approved):
LGFM

PR #1201, opened 2024-02-13
test(quay): add first playwright tests

jrichter1:
Adds a playwright based test suite for quay plugin. This is to help with automation for features contributed by RHTAP.

Also adds a github action that runs said automation for any plugin that would have it as PR check.

### DIFF
--- a/.github/workflows/pr-playwright.yaml
+++ b/.github/workflows/pr-playwright.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   changes:
-    name: Scan for chages
+    name: Scan for changes
     runs-on: ubuntu-latest
     outputs:
       plugins: ${{ steps.scan.outputs.plugins }}
--- a/.github/workflows/pr-playwright.yaml
+++ b/.github/workflows/pr-playwright.yaml
@@ -1,0 +1,134 @@
+name: Playwright tests
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  changes:
+    name: Scan for chages
+    runs-on: ubuntu-latest
+    outputs:
+      plugins: ${{ steps.scan.outputs.plugins }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - name: Determine changes
+        id: scan
+        env:
+          HEAD: ${{ github.event.pull_request.head.sha }}
+          BASE: ${{ github.event.pull_request.base.sha }}
+        run: |
+          root=$(pwd)
+          cd plugins
+
+          changed=()
+          for f in */; do
+            if git diff --name-only $BASE $HEAD | grep $f -q; then
+              if [[ ! -L "$f" && -f "$f/package.json" ]]; then
+                cd $f
+
+                if npm run | grep ui-test -q; then
+                  changed+=($f)
+                fi
+                cd $root/plugins
+              fi
+            fi
+          done
+
+          JSON="[$(echo ${changed[@]} | sed 's/ /,/g')]"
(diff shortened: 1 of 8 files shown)


invincibleJai:
Thanks @jrichter1, This looks good. would be good to update `CONTRIBUTING.MD` with steps needed to run and provide test  results in the PR description 


Locally i needed to install playwright as well

`yarn playwright install --with-deps chromium`

.github/workflows/pr-playwright.yaml
```diff
@@ -0,0 +1,98 @@
+name: Playwright tests
+
+on:
+  pull_request:
```
invincibleJai, line 4: as this workflow runs on pull_request so does that run for quay only if changes are in quay plugin or even if it's not changed?

jrichter1: right now it will run for all plugins that have the `ui-test` npm script, regardless of changes

is that an issue?

invincibleJai: That would be nice if we can run It based on change

jrichter1: updated

jrichter1:
> Thanks @jrichter1, This looks good. would be good to update `CONTRIBUTING.MD` with steps needed to run and provide test results in the PR description
> 
> Locally i needed to install playwright as well
> 
> `yarn playwright install --with-deps chromium`

updated

invincibleJai:
@jrichter1 is there a way to run it in headed mode, by default it runs in headless mode by default, so I tried to specify the 
`–headed` to run on headed mode but didn't work as this could be useful at times during writing tests

plugins/quay/tests/quay.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    ];
+    const thead = page.locator('thead');
+
+    for (const col of columns) {
+      await expect(thead.getByText(col)).toBeVisible();
+    }
+  });
+
+  test('Vulnerabilities are listed', async () => {
+    const severity = ['High:', 'Medium:', 'Low:'];
```
invincibleJai, line 36: how is the data being fetched, will it change with time are always have these severities?

plugins/quay/playwright.config.ts
```diff
@@ -0,0 +1,33 @@
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
```
invincibleJai, line 6: will this config be at each plugin level or can it be at the app level as this would be similar for most plugins?

plugins/quay/tests/quay.spec.ts
jrichter1: it is being fetched from a quay repo that I've made for this reason. The way I see it, if the image doesn't get updated, the number of vulnerabilities should not decrease. As far as the test cares, the number just needs to stay above zero.

plugins/quay/playwright.config.ts
jrichter1: It is going to be similar for most plugins. I've put it at the plugin level since it was indicated to me that plugins might change code bases in the future to make it self contained and easy to move.

plugins/quay/tests/quay.spec.ts
invincibleJai: yeah, it doesn't change with time then it should be good.

invincibleJai:
@subhashkhileri @josephca FYI

.github/workflows/pr-playwright.yaml
```diff
@@ -0,0 +1,134 @@
...
+    branches: [main]
+
+jobs:
+  changes:
+    name: Scan for chages
```
karthikjeeyar, line 9: typo:
```suggestion
    name: Scan for changes
```

karthikjeeyar (approved):

PR #1203, opened 2024-02-13
fix(rbac): reduce the catalog calls when build graph

PatAKnight:
## Description

This PR aims to reduce the number of times we make a call to the catalog while building the group hierarchy graph for a user. In our original approach, we would find that for each call to the authorize endpoint, we would retrieve the number of roles that the RBAC plugin is aware of, multiplied by the hierarchy level of the user. So, in the case of 52 roles and a user nested under four groups, we would end up with 52 * 4 = 208 catalog calls.

In the update, I chose to make a single call to retrieve all of the groups at once and then build the graph based on the retrieved information. This now means that for every authorize endpoint, we only make a call to the catalog the same number of times as the number of roles that the RBAC plugin is aware of.

Another aim of this PR is to separate some of the code from the BackstageRoleManager. Originally, we had the Role, which stored all of the roles that the RBAC plugin is aware of, and the AncestorSearchMemo, which was used to build the graph. I opted to move these into their own files to hopefully make maintenance easier.

### DIFF
--- a/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
+++ b/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
@@ -93,7 +93,7 @@ export class AncestorSearchMemo {
     const groupName = `group:${group.metadata.namespace?.toLocaleLowerCase(
       'en-US',
     )}/${group.metadata.name.toLocaleLowerCase('en-US')}`;
-    if (!memo.hasEntityRef(group.metadata.name)) {
+    if (!memo.hasEntityRef(groupName)) {
       memo.setNode(groupName);
     }
 
--- a/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
+++ b/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
@@ -109,10 +109,7 @@ export class AncestorSearchMemo {
   async buildUserGraph(memo: AncestorSearchMemo) {
     const userGroups = this.allGroups.filter(group => {
       const members = group.spec?.members as string[];
-      if (members && members.includes(this.userName)) {
-        return true;
-      }
-      return false;
+      return members?.includes(this.userName);
     });
     userGroups.forEach(group => this.traverseGroups(memo, group));
   }
--- a/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
+++ b/plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
@@ -15,7 +15,7 @@ export class AncestorSearchMemo {
   private tokenManager: TokenManager;
   private catalogApi: CatalogApi;
 
-  private userEntityRef: string;
+  private userName: string;
 
   private allGroups: Entity[];
 
@@ -25,7 +25,7 @@ export class AncestorSearchMemo {
     catalogApi: CatalogApi,
   ) {
     this.graph = new Graph({ directed: true });
-    this.userEntityRef = userEntityRef.split('/')[1];
+    this.userName = userEntityRef.split('/')[1];
     this.tokenManager = tokenManager;
     this.catalogApi = catalogApi;
     this.allGroups = [];
@@ -109,7 +109,7 @@ export class AncestorSearchMemo {
   async buildUserGraph(memo: AncestorSearchMemo) {
     const userGroups = this.allGroups.filter(group => {
       const members = group.spec?.members as string[];
-      if (members && members.includes(this.userEntityRef)) {
+      if (members && members.includes(this.userName)) {
         return true;
       }
       return false;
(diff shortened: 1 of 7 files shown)


plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  private allGroups: Entity[];
+
+  constructor(
+    userEntityRef: string,
+    tokenManager: TokenManager,
+    catalogApi: CatalogApi,
+  ) {
+    this.graph = new Graph({ directed: true });
+    this.userEntityRef = userEntityRef.split('/')[1];
```
AndrienkoAleksandr, line 28: This field actually is not userEntityRef any more (after split with `/`). I think it should have a name userName or something like that.

plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+  }
+
+  getNodes(): string[] {
+    return this.graph.nodes();
+  }
+
+  async getAllGroups(): Promise<void> {
+    const { token } = await this.tokenManager.getToken();
+    const { items } = await this.catalogApi.getEntities(
```
AndrienkoAleksandr, line 69: We need to check if catalog api really return all list groups. Theoretically it could support pagination and return list group limited by page size.

PatAKnight: Didn't think about the potential for pagination, will test to see if that is an issue and update accordingly.

AndrienkoAleksandr:
Hello. Sorry, I tested your pr with group inheritance permissions:

```
# group inheritance
p, role:default/catalog_user, catalog-entity, read, allow
p, role:default/catalog_user, scaffolder-action, use, allow
p, role:default/catalog_user, scaffolder-template, read, allow
p, role:default/catalog_user, catalog.entity.create, create, allow
p, role:default/catalog_user, catalog.location.create, create, allow

g, group:default/team-a, role:default/catalog_user
```

My user is member of group: group:default/team-a. But "register new component" feature was not activated, so this permissions doesn't work. I see only user level permissions works:

```
# group inheritance
p, role:default/catalog_user, catalog-entity, read, allow
p, role:default/catalog_user, scaffolder-action, use, allow
p, role:default/catalog_user, scaffolder-template, read, allow
p, role:default/catalog_user, catalog.entity.create, create, allow
p, role:default/catalog_user, catalog.location.create, create, allow

g, user:default/my-user-member-of-team-a, role:default/catalog_user
```

PatAKnight:
How is your user and group set up in the catalog? I am able to make group inheritance work. 

Here is the group that I am apart of: 
```
apiVersion: backstage.io/v1alpha1
kind: Group
metadata:
  namespace: default
  name: group-1m
spec:
  type: team
  profile:
    displayName: group-1m
  children: []
  parent: group-1g
  members:
    - 1k
    - 1l
    - pataknight
```

Here is my user:
```
apiVersion: backstage.io/v1alpha1
kind: User
metadata:
  namespace: default
  name: pataknight
spec:
  memberOf:
    - group-1m
```

and here are the permissions and roles:
```
g, group:default/group-1m, role:default/test

p, role:default/test, catalog.entity.create, create, allow
p, role:default/test, catalog.location.create, create, allow
```

plugins/rbac-backend/src/role-manager/ancestor-search-memo.ts
PatAKnight: With about 520 groups, I don't seem to be running into a pagination issue. It looks like there is a separate property for the `GetEntitiesRequest` that would be used to [limit](https://backstage.io/docs/reference/catalog-client.getentitiesrequest/) the number of entities that are returned.

If you want me to test with more, I can. Just let me know if 520 groups is sufficient.

plugins/rbac-backend/src/role-manager/role-manager.test.ts
```diff
@@ -791,6 +847,147 @@ describe('BackstageRoleManager', () => {
...
+      );
+
+      const test = await roleManager.hasLink(
+        'user:default/mike',
+        'role:default/team-g',
+      );
+      expect(test).toBeFalsy();
+    });
+
+    // user:default/mike should inherits role from group:default/team-f, and we have a complex graph
```
AndrienkoAleksandr, line 925: Minor group:default/team-f => group:default/team-e , because You are going to test this group in the follow test

plugins/rbac-backend/src/role-manager/role-manager.test.ts
```diff
@@ -791,6 +847,147 @@ describe('BackstageRoleManager', () => {
...
+        'team-b',
+      ]);
+      const groupEMock = createGroupEntity('team-e', undefined, [
+        'team-f',
+        'team-g',
+      ]);
+      const groupGMock = createGroupEntity(
+        'team-g',
+        'team-e',
+        ['team-d'],
```
AndrienkoAleksandr, line 882: Minor, it doesn't affect test: ['team-d'] => [] or redraw Hierarchy, please.

AndrienkoAleksandr: And the same in the next test

AndrienkoAleksandr:
Your fix with group inheritance works fine.

AndrienkoAleksandr (approved):

PR #1204, opened 2024-02-13
fix(rbac): use sqlite with in memory config for db disabled option

AndrienkoAleksandr:
# What does this pull request do

Use SQLite with in-memory config for the disabled database option. Earlier, we used the file adapter for this mode, but our code didn't save content to the file. So, in fact, the RBAC backend was working with in-memory mode without persisting any data. Also, the file location was absolutely wrong and bad - the node_modules located with the plugin binary. So, the easiest way to achieve the same functionality with zero support effort is to use SQLite3 with in-memory config.

# What issue is referenced in this pull request?

https://issues.redhat.com/browse/RHIDP-1244

### DIFF
--- a/plugins/rbac-backend/src/service/policy-builder.ts
+++ b/plugins/rbac-backend/src/service/policy-builder.ts
@@ -1,22 +1,15 @@
 import {
   DatabaseManager,
   PluginEndpointDiscovery,
-  resolvePackagePath,
   TokenManager,
 } from '@backstage/backend-common';
-import { DatabaseService } from '@backstage/backend-plugin-api';
 import { CatalogClient } from '@backstage/catalog-client';
 import { Config, ConfigReader } from '@backstage/config';
 import { IdentityApi } from '@backstage/plugin-auth-node';
 import { RouterOptions } from '@backstage/plugin-permission-backend';
 import { PermissionEvaluator } from '@backstage/plugin-permission-common';
 
-import {
-  FileAdapter,
-  newEnforcer,
-  newModelFromString,
-  StringAdapter,
-} from 'casbin';
+import { newEnforcer, newModelFromString } from 'casbin';
 import { Router } from 'express';
 import { Logger } from 'winston';
 
--- a/plugins/rbac-backend/src/service/policy-builder.ts
+++ b/plugins/rbac-backend/src/service/policy-builder.ts
@@ -4,13 +4,19 @@ import {
   resolvePackagePath,
   TokenManager,
 } from '@backstage/backend-common';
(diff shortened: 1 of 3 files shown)


plugins/rbac-backend/src/service/policy-builder.ts
```diff
@@ -39,30 +38,35 @@ export class PolicyBuilder {
...
+            connection: ':memory:',
+          },
+        },
+      });
+
+      env.logger.warn(`rbac backend plugin data won't be persistent`);
     }
 
+    const databaseManager =
+      await DatabaseManager.fromConfig(dbConfig).forPlugin('permission');
```
PatAKnight, line 63: We don't need the `await` here
```suggestion
    const databaseManager =
      DatabaseManager.fromConfig(dbConfig).forPlugin('permission');
```

plugins/rbac-backend/src/service/policy-builder.ts
```diff
@@ -39,30 +38,35 @@ export class PolicyBuilder {
...
+        backend: {
+          database: {
+            client: 'better-sqlite3',
+            connection: ':memory:',
+          },
+        },
+      });
+
+      env.logger.warn(`rbac backend plugin data won't be persistent`);
     }
```
PatAKnight, line 60: We could possibly go with an even simpler approach of just setting `dbConfig = env.config`. This is because for both options we are just looking at what is set in the config value `backend.database`. And if I am not mistaken, the backend will not start without the database being defined within the app-config. We can then remove the `const databaseEnabled = env.config.getOptionalBoolean('permission.rbac.database.enabled')` above and also our config value in `config.d.ts`. 

I think this could help with any potential confusions where someone set the database to true under `permission.rbac.database.enabled` but is using better-sqlite3. Or whenever they set the database to false and are using postgres.

What do you think?

PatAKnight (commented):
It looks like you have a failing test. It is this test [here](https://github.com/janus-idp/backstage-plugins/blob/55c3927fb5211e1ec78719fd38740eb29e481962/plugins/rbac-backend/src/service/policy-builder.test.ts#L123:L155). 

Otherwise it worked for me when trying out your PR locally. Attempted with postgres and without and had no issues.

plugins/rbac-backend/src/service/policy-builder.ts
AndrienkoAleksandr: Ok let's make it simplier. Project manager provided the same feedback, so let's drop permission.rbac.database.enabled option.

AndrienkoAleksandr:
Close on favour: https://github.com/janus-idp/backstage-plugins/pull/1214

PR #1205, opened 2024-02-13
fix(plugins): updated administration support button title

ciiay:
For [RHIDP-1141](https://issues.redhat.com/browse/RHIDP-1141): [janus-idp/backstage-plugins] Update or remove support button on dynamic-plugins-info

example-support-button-1


plugins/dynamic-plugins-info/src/components/DynamicPluginsInfoContent/DynamicPluginsInfoContent.tsx
```diff
@@ -7,7 +7,11 @@ import { DynamicPluginsTable } from '../DynamicPluginsTable/DynamicPluginsTable'
 export const DynamicPluginsInfoContent = () => (
   <>
     <ContentHeader title="">
-      <SupportButton title="Support">Some placeholder text</SupportButton>
+      <SupportButton>
+        The administration area in Developer Hub enables administrators to
```
gashcrumb, line 11: This button only shows up within the context of the list of installed plugins, but the description is probably more appropriate for the entire administration section.  I think probably we need a more focused text string here that's just for the table view that this plugin provides, @ShiranHi @invincibleJai WDYT?

gashcrumb: @ciiay For now, let's use a string like "All of the installed plugins", this will be consistent with similar text in the app like in the catalog page or API page.

ShiranHi:
@gashcrumb do we also have this button on the RBAC page? Will the text be the same across all tabs in the administration area? If not, we can customize the text for each tab

gashcrumb:
> @gashcrumb do we also have this button on the RBAC page? Will the text be the same across all tabs in the administration area? If not, we can customize the text for each tab

The support button is part of the individual plugin UI.  So the RBAC page will need to have one added, but it will be able to have it's own short text prompt.

gashcrumb (approved):

PR #1206, opened 2024-02-14
feat(rbac): save and display role description in the frontend

AndrienkoAleksandr:

https://github.com/janus-idp/backstage-showcase/issues/898

P.S.: It would be nice to merge it for GA 1.1

### DIFF
--- a/plugins/rbac/src/hooks/useRole.ts
+++ b/plugins/rbac/src/hooks/useRole.ts
@@ -1,0 +1,31 @@
+import { useAsync } from 'react-use';
+
+import { useApi } from '@backstage/core-plugin-api';
+
+import { Role } from '@janus-idp/backstage-plugin-rbac-common';
+
+import { rbacApiRef } from '../api/RBACBackendClient';
+
+export const useRole = (
+  roleEntityRef: string,
+): {
+  loading: boolean;
+  role: Role | undefined;
+  roleError: Error;
+} => {
+  const rbacApi = useApi(rbacApiRef);
+  const {
+    value: roles,
+    loading,
+    error: roleError,
+  } = useAsync(async () => await rbacApi.getRole(roleEntityRef));
+
+  return {
+    loading,
+    role: Array.isArray(roles) ? roles[0] : undefined,
+    roleError: (roleError as Error) || {
+      name: (roles as Response)?.status,
+      message: `Error fetching the role. ${(roles as Response)?.statusText}`,
+    },
+  };
+};
--- a/plugins/rbac/src/hooks/useSelectedMembers.ts
+++ b/plugins/rbac/src/hooks/useSelectedMembers.ts
@@ -9,24 +9,20 @@ import { rbacApiRef } from '../api/RBACBackendClient';
 import { SelectedMember } from '../components/CreateRole/types';
 import { MemberEntity } from '../types';
 import { getSelectedMember } from '../utils/rbac-utils';
+import { useRole } from './useRole';
 
 export const useSelectedMembers = (
   roleName: string,
 ): {
   members: MemberEntity[];
   selectedMembers: SelectedMember[];
+  role: Role | undefined;
   membersError: Error;
-  roleError?: Error;
+  roleError: Error;
   loading: boolean;
 } => {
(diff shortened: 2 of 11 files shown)


plugins/rbac/src/hooks/useSelectedMembers.ts
```diff
@@ -60,6 +64,7 @@ export const useSelectedMembers = (
   return {
     selectedMembers: data,
     members: membersArr,
+    roleDescription,
```
invincibleJai, line 67: will it make sense to return role metadata here and avoid the new hook `useRoleMetadata` or refactor role fetching to `useRoleMetadata` and consume it here

invincibleJai:
Thanks @AndrienkoAleksandr for PR ❤️ 

/cc @debsmita1 
/cc @divyanshiGupta

plugins/rbac/src/hooks/useSelectedMembers.ts
AndrienkoAleksandr: Done, I used second option.

AndrienkoAleksandr:
@invincibleJai, I think I handled your feedback.

plugins/rbac/src/hooks/useSelectedMembers.ts
```diff
@@ -10,23 +10,42 @@ import { SelectedMember } from '../components/CreateRole/types';
...
+
+  return {
+    loading,
+    role: Array.isArray(roles) ? roles[0] : undefined,
+    roleError: (roleError as Error) || {
+      name: (roles as Response)?.status,
+      message: `Error fetching the role. ${(roles as Response)?.statusText}`,
+    },
+  };
+};
```
divyanshiGupta, line 35: @AndrienkoAleksandr can you move this hook to a separate file?

divyanshiGupta:
@AndrienkoAleksandr can you also squash your commits into one and align PR title with commit title.

plugins/rbac/src/hooks/useSelectedMembers.ts
AndrienkoAleksandr: Done.

AndrienkoAleksandr:
> Since now we have support for adding/editing description, should we have edit icon on About card similar to other cards which will directly take the user to 1st step of edit form?
 
 +1, but I'm not sure that I have time to do that in this pr.
  
 >  We have some other fields also in the about card so we can skip it also for the time being. But just wanted to put it out there.
 
I think I have backend issue https://issues.redhat.com/browse/RHIDP-1382?filter=-1 around this topic. And this pr's included update role?.metadata?.modifiedBy and role?.metadata?.lastModified. So when we will resolve backend issue, then backend will provided read only fields "modifiedBy" and "lastModified" in the metadata  and frontend will display them without other farther modifications.

AndrienkoAleksandr:
> @AndrienkoAleksandr can you also squash your commits into one and align PR title with commit title?

Done.

divyanshiGupta:
@AndrienkoAleksandr on not adding any description while creating role, user should see No description in About card but its not showing anything.


https://github.com/janus-idp/backstage-plugins/assets/20724543/4a7db736-95f2-4410-b1e3-08ddf2f455b2

plugins/rbac/src/hooks/useRole.ts
```diff
@@ -0,0 +1,31 @@
...
+    error: roleError,
+  } = useAsync(async () => await rbacApi.getRole(roleEntityRef));
+
+  return {
+    loading,
+    role: Array.isArray(roles) ? roles[0] : undefined,
+    roleError: (roleError as Error) || {
+      name: (roles as Response)?.status,
+      message: `Error fetching the role. ${(roles as Response)?.statusText}`,
+    },
```
invincibleJai, line 29: with this data and roleError will always be there i.e as below, ideally it should be there only if there is error.

when data is there, roleError looks like below

```
message: "Error fetching the role. undefined"
name:  undefined
```

Above change is not related to this PR so we can take this as an issue and get this improved

cc @divyanshiGupta

invincibleJai (approved):
/lgtm

PR #1207, opened 2024-02-14
fix(kiali): update styles, remove item details links

josunect:
- Fix scroll and avoid table resize based on context
- Remove links for items details


aljesusg (approved):
LGFM

Issue #999, opened 2023-12-06
[RBAC] Admin user configured doesn't have access to all permission policies by default


invincibleJai:
### Describe the bug

When I simply add admins in the permission configuration as below and not policies in `rbac-policy.csv` as readme specify https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/README.md#configure-policy-admins

```
permission:
 enabled: true
 rbac:
  admin:
   users:
    - name: user:default/<user>
```
    

then the admin only has access to `policy-entity` and not other plugin entities like `catalog-entity`

I explicitly need to specify each policy in `rbac-policy.csv` to get access to other plugin entities even for admin

`p, user:default/<user>, catalog.entity.read, use, allow`

### Expected Behavior

Admin should have access to all the permission policies by default

### What are the steps to reproduce this bug?

1. Setup backend plugin following https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/README.md
2. add admins in the permission configuration as below in `app-config.yaml` and not policies in `rbac-policy.csv` as readme specify https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/README.md#configure-policy-admins

```
permission:
 enabled: true
 rbac:
  admin:
   users:
    - name: user:default/<user>
```
3. start the instance
4. Admin users do not have access to Read/create a catalog and this may happen with other permission policy as well

### Versions of software used and environment

"@janus-idp/backstage-plugin-rbac-backend":  1.6.6

debsmita1:
cc @PatAKnight @AndrienkoAleksandr

zjpersc:
I would agree that this is an important item. Ideally, we don't want to prime the pump or depend on the CSV file. With the large number of users and permission configs we believe we may end up with, leveraging the database solely would be preferred. The key is having the admin users automatically having the ability to create/manage permissions policies.

PR #1208, opened 2024-02-14
fix(rbac): allow for super users to have allow all access

PatAKnight:
## Description

Adds a 'super user' configuration value that grants users allow all access to their backstage instance. In this approach, I settled for creating an additional config value to capture the individuals who would be labeled super users within a backstage instance. This way we still preserve the integrity of the rbac admin role and have the ability to have users that will allow all priveliges.

Also, updated the documentation with the additional config value. Added an additional test for reloading the CSV file that was missed previously. This test checks if a user that originally was granted an allow for a particular permission, looses said allow result when removed from the csv file. 

## Fixes

- Fixes: #999

### DIFF
--- a/plugins/rbac-backend/README.md
+++ b/plugins/rbac-backend/README.md
@@ -95,7 +95,7 @@ NOTE: If you are using Red Hat Developer Hub backend plugin is pre-installed and
 
 ### Configure policy admins
 
-The RBAC plugin empowers you to manage permission policies for users and groups with a designated group of individuals known as policy administrators. These administrators are granted access to the RBAC plugin's REST API and user interface.
+The RBAC plugin empowers you to manage permission policies for users and groups with a designated group of individuals known as policy administrators. These administrators are granted access to the RBAC plugin's REST API and user interface as well as the ability to read from the catalog.
 
 You can specify the policy administrators in your application configuration as follows:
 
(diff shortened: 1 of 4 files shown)


plugins/rbac-backend/README.md
```diff
@@ -109,7 +109,21 @@ permission:
         - name: group:default/admins
 ```
 
-For more information on the available API endpoints, refer to the [API documentation](./docs/apis.md).
+The RBAC plugin also enables you to grant users the title of 'super user,' which provides them with unrestricted access throughout the Backstage instance.
```
AndrienkoAleksandr, line 112:
```suggestion
The RBAC plugin also enables you to grant users the title of 'super user', which provides them with unrestricted access throughout the Backstage instance.
```

plugins/rbac-backend/src/service/permission-policy.ts
```diff
@@ -166,6 +172,13 @@ export class RBACPermissionPolicy implements PermissionPolicy {
       'permission.rbac.policyFileReload',
     );
 
+    if (superUsers && superUsers.length > 0) {
+      for (const user of superUsers) {
+        const userName = user.getString('name');
```
AndrienkoAleksandr, line 177: It would be nice to add validation `validateEntityReference` before putting userName to the list

AndrienkoAleksandr: We have such validation for admins.

AndrienkoAleksandr: Oh, never mind we have validation bottom

AndrienkoAleksandr (approved):

PR #1213, opened 2024-02-15
fix(rbac): fix rbac tab route

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/RHIDP-1323

- Fixes the tab route
- using the latest rbac-backend package

**GIF:**

https://github.com/janus-idp/backstage-plugins/assets/22490998/44e0a92f-cd79-4b61-ac3e-f3919412e8a4


gashcrumb (approved):
Good to ensure the static use case is covered too, :+1:

PR #1224, opened 2024-02-16
fix(rbac): add data-testid, names and aria-label to RBAC UI components

debsmita1:
**Resolves:**
https://issues.redhat.com/browse/RHIDP-1341


plugins/rbac/src/components/EditRole.tsx
```diff
@@ -29,7 +29,7 @@ const EditRole = ({
         <IconButton
           color="inherit"
           component={Link}
-          aria-label="Update"
+          aria-label={dataTestId}
```
invincibleJai, line 32: we shouldn't update aria-label like this, it should be alternative text used for accessibility

PR #1226, opened 2024-02-19
feat(kiali): add Graph

aljesusg:
## Kiali Graph

image

## Check there is not endpoint
image


leandroberetta:
Hi @aljesusg, I'm noticing this graph is based on Cytoscape, but I thought we want the new PF graph for this effort, is there any reason for having the Cy instead of the PF graph?.

aljesusg:
> Hi @aljesusg, I'm noticing this graph is based on Cytoscape, but I thought we want the new PF graph for this effort, is there any reason for having the Cy instead of the PF graph?.

PF graph is based in patternfly components and we need to be more aligned with material-ui, ...be able to use BS toolbar style etc

leandroberetta:
> > Hi @aljesusg, I'm noticing this graph is based on Cytoscape, but I thought we want the new PF graph for this effort, is there any reason for having the Cy instead of the PF graph?.
> 
> PF graph is based in patternfly components and we need to be more aligned with material-ui, ...be able to use BS toolbar style etc

I agree to be aligned with material UI, but as we said in the past, whatever is missing in material UI, we need to add them by using other libraries, and I think there is a push to use the PF topology over others technologies.

PR #1228, opened 2024-02-20
test(tekton): add playwright tests for the plugin

jrichter1:
playwright tests for the tekton plugin and its new and shiny RHTAP features

should also stop the CI from finding changes in some PRs simply because they were based on an old commit


plugins/tekton/tests/tekton.spec.ts
```diff
@@ -0,0 +1,319 @@
+import { expect, Locator, Page, test } from '@playwright/test';
+
+test.describe('Tekton plugin', () => {
+  let page: Page;
+
+  test.beforeAll(async ({ browser }) => {
+    const context = await browser.newContext();
+    page = await context.newPage();
+    await page.goto('http://localhost:3000/tekton');
```
gashcrumb, line 9: It would be awesome if this could be overridden via an environment variable.

jrichter1: added PLUGIN_BASE_URL to the config

invincibleJai:
/approve

Thanks @jrichter1 

image

PR #1233, opened 2024-02-20
fix(orchestrator): implementation of getWorkflowById (v2)

gciavarrini:
By mistake the PR that added the new openapi v2 endpoints broke the `/workflows/:workflowId` v1 route.
In addition, the v2 endpoint wasn't properly added due to a wrong rebase.

### DIFF
--- a/plugins/orchestrator-common/src/openapi/openapi.yaml
+++ b/plugins/orchestrator-common/src/openapi/openapi.yaml
@@ -105,6 +105,30 @@ paths:
             text/plain:
               schema:
                 type: string
+  /v2/workflows/{workflowId}:
+    get:
+      operationId: getWorkflowById
+      description: Get a workflow by ID
+      parameters:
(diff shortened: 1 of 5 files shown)


plugins/orchestrator-common/src/openapi/openapi.yaml
```diff
@@ -105,6 +105,30 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/WorkflowDTO'
+        '500':
+          description: Error workflow by id
+          content:
+            text/plain:
+              schema:
+                type: string
```
jkilzi, line 131: Wouldn't it be better to return an object instead of just plain text?

gciavarrini: I can do that, but i'd like to change all the 500 definitions. 
Can i do that in another PR?

PR #1236, opened 2024-02-20
feat(orchestrator): add openapi generated typescript client

ydayagi:
FLPATH-828
https://issues.redhat.com/browse/FLPATH-828


plugins/orchestrator-common/src/auto-generated/openapi-client/.openapi-generator-ignore
```diff
@@ -0,0 +1,23 @@
+# OpenAPI Generator Ignore
```
gashcrumb, line 1: If these are generated files, why commit them to the repo?  Would it be better to generated this as a post-install thing perhaps?

ydayagi: we generate only when needed. upon demand.

plugins/orchestrator-common/package.json
```diff
@@ -35,7 +35,8 @@
     "clean": "backstage-cli package clean",
     "prepack": "backstage-cli package prepack",
     "postpack": "backstage-cli package postpack",
-    "openapi": "./scripts/openapi.sh"
+    "openapi": "./scripts/openapi.sh",
```
batzionb, line 38: I cloned the PR to try to generate the client and see how it works
My instinct was to go to orchestrator-backend and generate the client from there, I needed to look at the PR diff to realise it's in orchestrator-common.
Can it be moved to orchestrator-backend?

ydayagi: I was told to put it in common. please raise the issue In the slack channel

batzionb: Yes it should be in common
Seems like this work is already done in this [PR](https://github.com/janus-idp/backstage-plugins/pull/1649)

PR #1252, opened 2024-02-22
fix(orchestrator): improvements to backend services

caponetto:
- Some places in the services assume the response array is non-empty or the response object is not undefined, which could not be the case. So I've changed these parts to a more defensive code.
- When executing a workflow, I've added a call to fetch the instance details just to make sure the data is available. 

Please note that none of these improvements is changing the current behavior.

### DIFF
--- a/plugins/orchestrator-backend/src/service/api/v1.ts
+++ b/plugins/orchestrator-backend/src/service/api/v1.ts
@@ -16,8 +16,12 @@ import {
 } from '@janus-idp/backstage-plugin-orchestrator-common';
 
 import { DataIndexService } from '../DataIndexService';
+import { retryAsyncFunction } from '../Helper';
 import { SonataFlowService } from '../SonataFlowService';
 import { WorkflowService } from '../WorkflowService';
+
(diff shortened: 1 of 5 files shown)


plugins/orchestrator-backend/src/service/api/v1.ts
```diff
@@ -146,21 +150,31 @@ export namespace V1 {
...
+      endpoint: definition.serviceUrl,
       businessKey,
     });
 
     if (!executionResponse) {
       throw new Error(`Couldn't execute workflow ${workflowId}`);
     }
 
+    // Making sure the instance data is available before returning
+    await retryAsyncFunction({
```
gciavarrini, line 171: should we use ` retryAsyncFunction` for all the endpoints that interacts with dataindex? e.g. abort workflow, get instance by id, get workflows list, etc.)

caponetto: I'd use it only by demand. But let's see first if this code resolves the issue we are seeing.

jkilzi (approved):

PR #1277, opened 2024-02-27
fix(orchestrator): handle nullable start/state properties of process instance

caponetto:
https://issues.redhat.com/browse/FLPATH-1035

- Add debug logs for SonataFlow responses
- Remove unused `SonataFlow#fetchWorkflows` method
- Handle null values for `start` and `state` properties (ProcessInstance)
- Handle `PENDING` state (ProcessInstance)

### DIFF
--- a/plugins/orchestrator-common/src/auto-generated/api/models/schema.ts
+++ b/plugins/orchestrator-common/src/auto-generated/api/models/schema.ts
@@ -142,7 +142,8 @@ export interface components {
       | 'Error'
       | 'Completed'
       | 'Aborted'
-      | 'Suspended';
+      | 'Suspended'
+      | 'Pending';
     WorkflowRunStatusDTO: {
       key?: string;
       value?: string;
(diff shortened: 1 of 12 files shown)


jkilzi (approved):

plugins/orchestrator-common/src/auto-generated/api/models/schema.ts
```diff
@@ -142,7 +142,8 @@ export interface components {
       | 'Error'
       | 'Completed'
       | 'Aborted'
-      | 'Suspended';
+      | 'Suspended'
+      | 'Pending';
```
gciavarrini, line 146: @caponetto 
this schema is auto generated from ` plugins/orchestrator-common/src/openapi/openapi.yaml` spec.
Please change the spec, and then run `yarn --cwd plugins/orchestrator-common openapi`

The files that will be automatically updated are:
`plugins/orchestrator-common/src/auto-generated/api/definition.ts`,
`plugins/orchestrator-common/src/auto-generated/api/models/schema.ts`

caponetto: Oops, on it.

PR #1285, opened 2024-02-28
fix(rbac): improve error handling in retrieving permission metadata.

AndrienkoAleksandr:

Log errors when requesting rbac-backend permission metadata instead of propagating them. Ignore errors caused by non-JSON responses during permission metadata retrieval.

### What are the issues referenced in the pull request?

Fixes: https://issues.redhat.com/browse/RHIDP-1453
Workaround for: https://issues.redhat.com/browse/RHIDP-1456

### DIFF
--- a/plugins/rbac-backend/src/service/plugin-endpoints.ts
+++ b/plugins/rbac-backend/src/service/plugin-endpoints.ts
@@ -43,8 +43,8 @@ export class PluginPermissionMetadataCollector {
   constructor(
     private readonly discovery: PluginEndpointDiscovery,
     private readonly pluginIdProvider: PluginIdProvider,
+    private readonly logger: Logger,
     config: Config,
-    logger: Logger,
   ) {
     this.pluginIds = this.pluginIdProvider.getPluginIds();
     this.urlReader = UrlReaders.default({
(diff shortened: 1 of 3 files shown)


plugins/rbac-backend/src/service/plugin-endpoints.ts
```diff
@@ -108,9 +114,12 @@ export class PluginPermissionMetadataCollector {
...
         }
       } catch (err) {
-        if (!isError(err) || err.name !== 'NotFoundError') {
-          throw err;
+        if (isError(err) && err.name === 'NotFoundError') {
+          continue;
         }
+        this.logger.error(
+          `Failed to retrieve permission metadata for ${pluginId}. ${err}`,
+        );
```
PatAKnight, line 122: My only concern with logging the permission to the console is that it could create a lot of spam. If there are ten plugins that are loading into an RHDH instance that do not have permissions, then there will be ten log statements each time the `/api/permission/plugins/polices` endpoint is accessed.

AndrienkoAleksandr: I don't see a lot of logs. This is because, for backstage-plugins, if a plugin doesn't have a well-known endpoint, the API returns a 404 error, and we skip logging. In the case of backstage-showcase, if a plugin doesn't have a well-known endpoint, the API returns HTML, which our code fails to parse, and we silently skip this error (as a workaround).

AndrienkoAleksandr: Ideally, it would be nice to retrieve and cache the list of permissions, but we cannot be sure when all dynamic plugins have completed startup.

PatAKnight (approved):
/lgtm
/approve

PR #1298, opened 2024-02-29
feat(rbac): query the catalog database when building graph

PatAKnight:
## Description

As part of the work to increase the performance of the RBAC backend plugin, this PR further reduces the number of catalog api calls to zero by instead attempting to query the catalog database table 'relations'. This relations table contains all of the relations between two entities within the catalog. We can query this table to retrieve relations of 'memberOf' for a specified user and 'childOf' for groups that have parents. We can then begin to build the graph using this minimum amount of information.

Went in and fixed the broken tests to now work. Also, fixed an issue with the logger mock within the `role-manager.test.ts` not being reset after each test. Now, the correct log statements will show up ensuring that we have accurate tests. Also, renamed `enforcer.delegate.test.ts` to `enforcer-delegate.test.ts` to match the format of other test file names.

Added additional tests to cover the new methods added to the `AncestorSearchMemo` as well as added some missing tests for `RoleList`. In `ancestor-search-memo.test.ts`, I opted for a simple graph when testing `traverseRelations` and `buildUserGraph`. My reasoning for this is because the logic for `traverseRelations` is very similar to `traverseGroups` where we already have some fairly in depth graphs for testing. The only difference as of right now is that the `traverseRelations` graph will start with the user whereas the `traverseGroups` graph will start with the groups that the user is associated with.

EX traverse relations -
```
         -> group-a -> group-c
user -> |
         -> group-b -> goup-d
```

EX traverse groups - 
```
group-a -> group-c

group-b -> goup-d
```

### DIFF
--- a/plugins/rbac-backend/src/role-manager/role-list.test.ts
+++ b/plugins/rbac-backend/src/role-manager/role-list.test.ts
@@ -1,10 +1,16 @@
 import { RoleList } from './role-list';
 
 describe('RoleList', () => {
-  const roleList = new RoleList('user:default/adam');
-  const newRole = new RoleList('role:default/test');
-  const extraRole = new RoleList('role:default/extra');
-  roleList.addRole(extraRole);
+  let roleList: RoleList;
+  let newRole: RoleList;
+  let extraRole: RoleList;
+
+  beforeEach(() => {
+    roleList = new RoleList('user:default/adam');
+    newRole = new RoleList('role:default/test');
+    extraRole = new RoleList('role:default/extra');
+    roleList.addRole(extraRole);
+  });
 
   describe('addRole', () => {
     it('should add a role', () => {
@@ -19,11 +25,18 @@ describe('RoleList', () => {
 
       expect(roleList.hasRole('role:default/test', 1)).toEqual(true);
       expect(roleList.hasRole('role:default/does-not-exist', 1)).toEqual(false);
+      expect(roleList.getRoles().length).toEqual(2);
+
+      roleList.addRole(newRole);
+
+      expect(roleList.getRoles().length).toEqual(2);
     });
   });
 
   describe('getRoles', () => {
     it('should return all roles', () => {
+      roleList.addRole(newRole);
+
       const allRoles = roleList.getRoles();
       expect(allRoles).toEqual([extraRole, newRole]);
     });
@@ -31,6 +44,10 @@ describe('RoleList', () => {
 
   describe('deleteRole', () => {
     it('should remove a role', () => {
+      roleList.addRole(newRole);
+
+      expect(roleList.hasRole('role:default/test', 1)).toEqual(true);
+
       roleList.deleteRole(newRole);
 
       expect(roleList.hasRole('role:default/test', 1)).toEqual(false);
--- a/plugins/rbac-backend/src/role-manager/role-list.test.ts
+++ b/plugins/rbac-backend/src/role-manager/role-list.test.ts
@@ -1,0 +1,40 @@
+import { RoleList } from './role-list';
+
+describe('RoleList', () => {
+  const roleList = new RoleList('user:default/adam');
+  const newRole = new RoleList('role:default/test');
+  const extraRole = new RoleList('role:default/extra');
+  roleList.addRole(extraRole);
+
+  describe('addRole', () => {
+    it('should add a role', () => {
+      roleList.addRole(newRole);
+
+      expect(roleList.hasRole('role:default/test', 1)).toEqual(true);
+      expect(roleList.hasRole('role:default/does-not-exist', 1)).toEqual(false);
+    });
+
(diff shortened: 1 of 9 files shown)


plugins/rbac-backend/src/role-manager/role-list.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  describe('addRole', () => {
+    it('should add a role', () => {
+      roleList.addRole(newRole);
+
+      expect(roleList.hasRole('role:default/test', 1)).toEqual(true);
+      expect(roleList.hasRole('role:default/does-not-exist', 1)).toEqual(false);
+    });
+
+    it('should not add a duplicate of an existing role', () => {
+      roleList.addRole(newRole);
```
AndrienkoAleksandr, line 18: This test is the same with 'addRole'

plugins/rbac-backend/src/role-manager/role-manager.test.ts
```diff
@@ -658,7 +667,7 @@ describe('BackstageRoleManager', () => {
       );
       expect(result).toBeFalsy();
       expect(loggerMock.warn).toHaveBeenCalledWith(
-        'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-b"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for group: group:default/team-a',
+        'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-b"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for group: role:default/team-a',
```
AndrienkoAleksandr, line 670: Error message looks strange, because message talks about group, but prints role.

AndrienkoAleksandr: I think that's because you used: 

roleManager.hasLink(
   'user:default/mike',
   'role:default/team-b',
 ); 
      
 Second argument is role, but as far I remember casbin engine asks only about user or group entity reference like the second argument.

plugins/rbac-backend/src/role-manager/role-manager.test.ts
```diff
@@ -907,7 +916,7 @@ describe('BackstageRoleManager', () => {
       );
       expect(result).toBeFalsy();
       expect(loggerMock.warn).toHaveBeenCalledWith(
-        'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-c"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for group: group:default/team-a',
+        'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-c"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for group: role:default/team-e',
```
AndrienkoAleksandr, line 919: Looks like 'user:default/mike' doesn't have cycle dependency for role: group:default/team-e.

plugins/rbac-backend/src/role-manager/role-list.test.ts
PatAKnight: Yes, it is the same as the first test but it is only because I am trying to test this [scenario](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/src/role-manager/role-list.ts#L12:L14) to get 100% test coverage. Since I am not recreating the Role Manager in this test, I am essentially calling `addRole` a second time with the `newRole` to invoke that it there isn't another one added. To further expand on this test, I could potentially check the length of `roleList` to ensure that there are still only two roles. What do you think?

plugins/rbac-backend/src/role-manager/role-manager.test.ts
PatAKnight: It is kind of strange but hasLink will compare the user that we are checking permissions of against the roles that it is aware of. So the second argument will be a role so long as grouping policies are set up with roles. 

But I can update the error message to instead say: 'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-b"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for role: role:default/team-a'

or: 'Detected cycle dependencies in the Group graph: [["group:default/team-a","group:default/team-b"]]. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for groups: group:default/team-a, group:default/team-b'

plugins/rbac-backend/src/role-manager/role-manager.test.ts
PatAKnight: Yes, user:default/mike doesn't have a cycle dependency for role:default/team-e with group:default/team-e. However, there is a cycle dependency in the graph so no matter what role is chosen for mike, there will always be the warning of a cyclic dependency in the graph.

AndrienkoAleksandr (approved):

PR #1299, opened 2024-02-29
fix(orchestrator): walk around the state field is empty issue when fetch instance

yzhao583:
Jira: [https://issues.redhat.com/browse/FLPATH-1060](https://issues.redhat.com/browse/FLPATH-1060)

When run workflow the `state` field of the `instance` object could be `null`, but actually it should not be. We walk around it in this fix.

### DIFF
--- a/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
+++ b/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
@@ -105,7 +105,8 @@ export const WorkflowInstancePage = ({
     (curValue: AssessedProcessInstance | undefined) =>
       !!curValue &&
       (curValue.instance.state === 'ACTIVE' ||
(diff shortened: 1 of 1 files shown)


caponetto:
@yzhao583 don't we also need to hide the buttons (Rerun and Abort) when the state is null?

yzhao583:
@caponetto IMO, predictable actions provide a better user experience than hiding/showing some actions. In addition, this is a walk around, we may not want over-engineering it.

caponetto:
> @caponetto IMO, predictable actions provide a better user experience than hiding/showing some actions. In addition, this is a walk around, we may not want over-engineering it.

NP! I don't think changing a number is over-engineering it and 5s might be too long to refresh the data, but that's ok.

jkilzi:
LGTM

caponetto (approved):

PR #1303, opened 2024-03-02
fix: text filtering in Notifications

mareklibra:
The filtering used to be done twice - by both the BE and the FE. Since we stopped using a regular table, the FE filtering did not find match and so filtered-out all the messages.

With this fix, just the BE performs the filter.

Fixes: [FLPATH-1067](https://issues.redhat.com//browse/FLPATH-1067)


mareklibra:
[Uploading screencast.flpath-1067.fixTextFilter.webm…]()

PR #1307, opened 2024-03-04
chore(deps): upgrade to 1.23.4

schultzp2020:
Upgrade to backstage 1.23.4


package.json
```diff
@@ -25,7 +25,7 @@
     "new": "janus-cli new --do-not-edit-packages",
     "release": "multi-semantic-release",
     "prepare": "husky install",
-    "versions:bump": "backstage-cli versions:bump && find . -name 'package.json' ! -path '*/node_modules/*' -exec sed -i '' -e '/devDependencies/,/\\}/{ s/\"\\^/\"/; }' {} \\;"
+    "versions:bump": "backstage-cli versions:bump && find . -name 'package.json' ! -path '*/node_modules/*' -exec sed -i -e '/devDependencies/,/\\\\}/{ s/\\\"\\\\^/\\\"/; }' {} ; && yarn install"
```
nickboldt, line 28: ah, I see you're removing the OSX sed implementation in favour of the one true gnu sed implementation. +1. 

you might be able to remove some of those extraneous `\\` slashes by using `sed -r` to enable extended regex

nickboldt (approved):
sure, anything that annoys osx users gets a +1 from me. OSX users: use gnu sed.

PR #1309, opened 2024-03-06
feat(orchestrator): verify availability and cache workflow definition IDs

caponetto:
https://issues.redhat.com/browse/FLPATH-962
https://issues.redhat.com/browse/FLPATH-1099

Working around the fact that the Data Index is not cleaned up when a workflow CR is deleted. Similarly to what the template catalog does, now we have a cache of workflow definition IDs that periodically fetches the service URLs to confirm they are up (currently at 5-second intervals). If the service is not up, its definition ID is removed from the cache. Thus, the cache is verified on every call to Data Index and we can control if we want to skip or throw an error when an expected workflow definition ID is not in the cache. We can revisit this workaround once the SonataFlow Gateway API is implemented.

Also in this PR:

- Change abort instance endpoint to `/instances/<instanceId>/abort` instead of `/workflows/...`.
- Change the response to `void` when the abort instance operation is successful. The returned object was not being used.
- Add to the Workflow table the same polling mechanism we have for the instance table.
- Group DevMode-related code from `SonataFlowService` into a new service named `DevModeService`.
- Create `OrchestratorService` to encapsulate the communication between the backend plugin and SonataFlow infra.
- Move V1 and V2 APIs to classes. This way, services don't need to be passed as arguments to every function.
- Remove duplicated  `api.register('executeWorkflow'...`.
- Reorganize the backend router slightly to align with the other changes.
- Remove legacy `WorkflowService` as it is not necessary anymore.
- Fixed V2 tests to align with the other changes.
- Update OpenAPI auto-generated code to align with the other changes.


caponetto:
@gciavarrini I'm still validating this code but would appreciate it if you could help me with the review

caponetto:
@batzionb @JudeNiroshan please be aware of this code refactoring

caponetto:
This PR is ready for review. (I see 4 issues raised by SonarClould but they belong to another plugin)

plugins/orchestrator-backend/src/service/DevModeService.ts
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    this.logger.info(`Reloading workflows from ${this.connection.repoUrl}`);
+    const localPath = this.connection.resourcesPath;
+    if (await fs.pathExists(localPath)) {
+      this.logger.info(`Path ${localPath} already exists. Skipping clone.`);
+      return;
+    }
+
+    await fs.remove(localPath);
```
gciavarrini, line 232: why do we need this `remove`?
in line 227 you check if `pathExists`, so in line 232 there is no need to remove.

caponetto: Good catch. This code was only moved from `SonataFlowService` but I think we can remove this line.

gciavarrini (approved):
lgtm

batzionb:
@caponetto 
it's hard to follow up on which change is related to which fix
Can you please split it to separate PRs?

- Refactoring 
- Fix for https://issues.redhat.com/browse/FLPATH-962
- Fix for https://issues.redhat.com/browse/FLPATH-1099

Especially since the caching solution is a workaround, it should be easy to remove it, and this would be easier to accomplish if it's clear to see which changes introduce the caching

batzionb:
What happens if the orchestrator is loaded before the cache is initialized?

caponetto:
> @caponetto it's hard to follow up on which change is related to which fix Can you please split it to separate PRs?
> 
> * Refactoring
> * Fix for https://issues.redhat.com/browse/FLPATH-962
> * Fix for https://issues.redhat.com/browse/FLPATH-1099
> 
> Especially since the caching solution is a workaround, it should be easy to remove it, and this would be easier to accomplish if it's clear to see which changes introduce the caching

Sorry @batzionb, the work is already done here. The PR is not big, so I don't see a reason to split it. Also, everything is somehow connected, that's why I've bundled all changes in one single PR.
It would be easy to remove the cache if we want to, since everything is done in the `OrchestratorService`.
The cache initialization is part of the router initialization, so there is no way the plugin is initialized before the cache.

batzionb:
@caponetto 
I guess we have different definitions for what a large PR is...
I understand it's a lot of effort to split it so I'll review it as is

caponetto:
When running locally as static plugin, the `scheduler` needs to be passed in your `packages/backend/src/plugins/orchestrator.ts`:
```ts
import { Router } from 'express';

import { createRouter } from '@janus-idp/backstage-plugin-orchestrator-backend';

import { PluginEnvironment } from '../types';

  env: PluginEnvironment,
): Promise<Router> {
  return await createRouter({
    config: env.config,
    logger: env.logger,
    discovery: env.discovery,
    catalogApi: env.catalogApi,
    urlReader: env.reader,
    scheduler: env.scheduler, // <----------------
  });
}
```

batzionb (approved):

PR #1313, opened 2024-03-06
feat(orchestrator): support pagination for /instances and /overview

JudeNiroshan:
JIRA: [FLPATH-749](https://issues.redhat.com/browse/FLPATH-749)

## Description

Use query parameters in the GET request to identify the following options:

- pageSize
- page
- orderBy
- orderDirection

E.g.: http://localhost:7007/api/orchestrator/v2/workflows/instances?pageSize=10&page=2&orderBy=start&orderDirection=DESC

Response structure for the /v2/workflows/instances & /v2/workflows/overview should be as below:

 
```
{
    items: [],
    paginationInfo: {
        pageSize: 10,
        page: 2,
        totalCount: 50
    }
}
```

### DIFF
--- a/plugins/orchestrator-backend/src/service/DataIndexService.ts
+++ b/plugins/orchestrator-backend/src/service/DataIndexService.ts
@@ -93,7 +93,7 @@ export class DataIndexService {
   }
 
   public async getWorkflowInfos(
-    pagination: Pagination,
+    pagination?: Pagination,
   ): Promise<WorkflowInfo[]> {
     this.logger.info(`getWorkflowInfos() called: ${this.dataIndexUrl}`);
 
@@ -120,9 +120,9 @@ export class DataIndexService {
   }
 
   public async fetchProcessInstances(
-    pagination: Pagination,
+    pagination?: Pagination,
   ): Promise<ProcessInstance[] | undefined> {
-    pagination.sortField ??= FETCH_PROCESS_INSTANCES_SORT_FIELD;
+    if (pagination) pagination.sortField ??= FETCH_PROCESS_INSTANCES_SORT_FIELD;
 
     const graphQlQuery = buildGraphQlQuery({
       type: 'ProcessInstances',
--- a/plugins/orchestrator-backend/src/service/DataIndexService.ts
+++ b/plugins/orchestrator-backend/src/service/DataIndexService.ts
@@ -12,6 +12,9 @@ import {
 } from '@janus-idp/backstage-plugin-orchestrator-common';
 
 import { ErrorBuilder } from '../helpers/errorBuilder';
+import { buildGraphQlQuery } from '../helpers/queryBuilder';
+import { Pagination } from '../types/pagination';
+import { FETCH_PROCESS_INSTANCES_SORT_FIELD } from './constants';
 
 export class DataIndexService {
   private client: Client;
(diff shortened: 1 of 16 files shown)


plugins/orchestrator-backend/src/service/DataIndexService.ts
```diff
@@ -147,6 +154,24 @@ export class DataIndexService {
...
+    });
+    this.logger.debug(`GraphQL query: ${graphQlQuery}`);
+    const result = await this.client.query(graphQlQuery, {});
+
+    if (result.error) {
+      this.logger.error(`Error when fetching instances: ${result.error}`);
+      throw result.error;
+    }
+
+    const idArr = result.data.ProcessInstances as ProcessInstance[];
```
batzionb, line 170: If you're already getting all of the instances why not just do the pagination logic here? 
I'm concerned this is making things slower... the whole idea of pagination is to make things faster

batzionb: @caponetto 
WDYT

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -344,8 +351,8 @@ function setupInternalRoutes(
     },
   );
 
-  router.get('/instances', async (_, res) => {
-    await V1.getInstances(services.dataIndexService)
+  router.get('/instances', async (req, res) => {
+    await V1.getInstances(services.dataIndexService, buildPagination(req))
```
batzionb, line 355: please remove the changes from V1

JudeNiroshan: [743cfca](https://github.com/janus-idp/backstage-plugins/pull/1313/commits/743cfcaa764696587fc59150f8291cc1d1387123)

plugins/orchestrator-backend/src/service/DataIndexService.ts
batzionb: can you please check what the difference in payload is?

JudeNiroshan: It depends on the number of **all instances** and the **process definition structure**. When the number of nodes is high, the instance will have a larger JSON output. When the result payload is larger, it can lead to [graphQL query complexity](https://hygraph.com/docs/api-reference/basics/query-complexity). It will result in a runtime error. 

However, if we have the `totalCount` as a result output[1] in the graphQL, we can remove this method. 

[1] https://github.com/apache/incubator-kie-kogito-apps/issues/2012

Here is a simple comparison which I did with the default orchestrator container we use in our dev setup. I have mostly executed the Hello world workflow. I have 12 workflow instances.

- GraphQL query without pagination params(returns all instance data) - 2.76KB
- GraphQL query only to fetch all IDs - 450B

plugins/orchestrator-backend/src/service/DataIndexService.ts
```diff
@@ -121,10 +119,19 @@ export class DataIndexService {
...
-      '{ ProcessInstances ( orderBy: { start: ASC }, where: {processId: {isNull: false} } ) { id, processName, processId, businessKey, state, start, end, nodes { id }, variables, parentProcessInstance {id, processName, businessKey} } }';
-
+  public async fetchProcessInstances(
+    pagination?: Pagination,
+  ): Promise<ProcessInstance[] | undefined> {
+    if (pagination) pagination.sortField ??= FETCH_PROCESS_INSTANCES_SORT_FIELD;
+
+    const graphQlQuery = buildGraphQlQuery({
+      type: 'ProcessInstances',
+      queryBody:
```
batzionb, line 129: IIUC the behavior in v1 in which by default it's sorted asc by start won't be here anymore

batzionb (approved):

PR #1315, opened 2024-03-07
fix(notifications): return a detailed message for validation errors

ydayagi:
[FLPATH-1091](https://issues.redhat.com//browse/FLPATH-1091)
https://issues.redhat.com/browse/FLPATH-1091

### DIFF
--- a/plugins/notifications-backend/src/service/router.ts
+++ b/plugins/notifications-backend/src/service/router.ts
@@ -142,7 +142,8 @@ export async function createRouter(
     }
     const validation = api.validateRequest(req as Request);
     if (!validation.valid) {
-      throw validation.errors;
+      res.status(500).json({ status: 500, err: validation.errors });
+      return;
     }
 
     api.handleRequest(req as Request, req, res).catch(next);


plugins/notifications-backend/src/service/router.ts
```diff
@@ -142,7 +142,8 @@ export async function createRouter(
     }
     const validation = api.validateRequest(req as Request);
     if (!validation.valid) {
-      throw validation.errors;
+      res.status(500).json({ status: 500, err: validation.errors });
```
mareklibra, line 145: HTTP 500 means that the service failed internally, a bug or similar.
Incorrect input is HTTP 422.

ydayagi: the behavior that express/node/backstage had from the beginning was 500. I am not about to change what they do. my change is for restoring something that was lost (regression) by one of the PRs or version update of a library (express? node? backstage?).
regardless of what backstage/express does, anything but 200 is wrong. HTTP is the transport layer for REST. we must not use its status codes. 200 means that the HTTP server received the request and was able to parse it and find a handler for it. any error by that handler is application specific and out of the scope of HTTP. there is a bad (common) behavior by REST implementors to use 500. That's not so bad. anything else is very bad.
some implementors (the good ones) use only 200 and an object in the response to indicate error or return value.
this is why HTTP inspection tools (WAF, DoS, IDS, IPS etc) have configurations for cusotm pages (custom page not found, custom not authorized etc).
what a good app should do is:
1. define a root error/excpetion object (e.g. OrcException)
2. put an error/exception handler in the upper most level to catch this type
3. upon such exception return 200 with the exception in the HTTP body/content
We don't have that. so for now we should limit ourselves to 500.
Definitely not 4xx. 4xx means that the request is bad. But HTTP wise it is fine.

mareklibra: I think its a common and useful practice to use HTTP codes for reporting errors/issues/redirects/etc by REST APIs.
There is a ton of articles/works confirming this approach ([example](https://blog.treblle.com/the-importance-of-http-status-codes-in-rest-apis/)). The REST leverages HTTP by definition, search for the first work by Roy Fielding.

Introducing an application-specific contract in responses is reinventing the wheel and making mess which  is complicating the use of the API.

I do not agree with the provided clarification. Anyway, it's still better to return 500 as implemented than 200 for everything. So to move forward, I can mark this as lgtm.

mareklibra (approved):

PR #1317, opened 2024-03-07
fix(orchestraotr): resolved grey background appears in actions column in workflows table

batzionb:
root caused to [override of background-color in MuiDisabled](https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/src/themes/componentOverrides.ts#L59)
This override is meant for a disabled tab, but it propogates down to any element under a tab that has the MuiDisabled class
I searched in material UI for a way to localise the disabled override but couldn't find a better solution
The solution in this PR adds a custom class to a div around the table that overrides MuiDisabled class backgroundColor

### DIFF
--- a/plugins/orchestrator/src/components/ui/OverrideBackstageTable.tsx
+++ b/plugins/orchestrator/src/components/ui/OverrideBackstageTable.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+
+import {
+  Table as BackstageTable,
+  TableProps,
+} from '@backstage/core-components';
+
+import { makeStyles } from '@material-ui/core';
+
+// Workaround by issue created from overriding the tab theme in the backstage-showcase to add a gray background to disabled tabs.
+// This is achieved by overriding the global Mui-disabled class, which results in the actions column header background turning gray.
+// See https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/src/themes/componentOverrides.ts#L59
+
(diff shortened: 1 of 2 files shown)


caponetto:
@batzionb please fix the PR title

plugins/orchestrator/src/components/ui/OverrideBackstageTable.tsx
```diff
@@ -0,0 +1,31 @@
...
+  TableProps,
+} from '@backstage/core-components';
+
+import { makeStyles } from '@material-ui/core';
+
+// Workaround by issue created from overriding the tab theme in the backstage-showcase to add a gray background to disabled tabs.
+// This is achieved by overriding the global Mui-disabled class, which results in the actions column header background turning gray.
+// See https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/src/themes/componentOverrides.ts#L59
+
+const useStyles = makeStyles({
```
caponetto, line 14: @batzionb just a question.
What happens if we need something disabled in the Table?

batzionb: I don't think we will want that gray background
That's specific for tabs
The default behavior of disabled of Mui is a different font color

caponetto: Great, thanks for the info!

PR #1330, opened 2024-03-12
fix(rbac): rework condition policies to bound them to RBAC roles

AndrienkoAleksandr:

Rework condition policies to bound them to RBAC roles

### What issue is referenced in this pull request?

https://issues.redhat.com/browse/RHIDP-1526
https://issues.redhat.com/browse/RHIDP-1616

### How to test it:

You need to have two users: admin and test user.  "user:default/logarifm" is my test user and member of group:defalt/team-a. Token should be retrieved for admin user.

#### Let's create test role:

`
curl -X POST "http://localhost:7007/api/permission/roles" -d '{ "memberReferences":  [ "user:default/logarifm" ], "name": "role:default/test" }' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

#### Let's create condition to display for user logarifm only his own catalogs:

`
curl -X POST "http://localhost:7007/api/permission/roles/conditions" -d '{"result":"CONDITIONAL","roleEntityRef": "role:default/test","pluginId":"catalog","resourceType":"catalog-entity", "permissionMapping": ["read"], "conditions":{"rule":"IS_ENTITY_OWNER","resourceType":"catalog-entity","params":{"claims":["group:default/team-a"]}}}' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

OR you can use criteria "not" and hide API catalogs for user logarifm:

`
curl -X POST "http://localhost:7007/api/permission/roles/conditions" -d '{"result":"CONDITIONAL","roleEntityRef": "role:default/test","pluginId":"catalog","resourceType":"catalog-entity","permissionMapping": ["read"],"conditions":{"not":{"rule":"IS_ENTITY_KIND","resourceType":"catalog-entity","params":{"kinds":["Api"]}}}}' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

Login like user logarifm and check catalog UI.

Retrieve list of all conditions: 

`
curl -X GET "http://localhost:7007/api/permission/roles/conditions" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

Filter conditions with the query params:

`
curl -X GET "http://localhost:7007/api/permission/roles/conditions?roleEntityRef=role:default/test&&pluginId=catalog&&resourceType=catalog-entity&&actions=read" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

#### Get condition by id:

`
curl -X GET "http://localhost:7007/api/permission/roles/conditions/1" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

#### Update condition policy to display for logarifm user his own catalogs and list user groups:

`
curl -X PUT "http://localhost:7007/api/permission/roles/conditions/1" -d '{"result":"CONDITIONAL","roleEntityRef":"role:default/test", "pluginId":"catalog","resourceType":"catalog-entity","permissionMapping": ["read"],"conditions":{"anyOf":[{"rule":"IS_ENTITY_OWNER","resourceType":"catalog-entity","permissions": [{"name": "catalog.entity.read", "action": "read"}],"params":{"claims":["group:default/team-a"]}},{"rule":"IS_ENTITY_KIND","resourceType":"catalog-entity","params":{"kinds":["Group"]}}]}}' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

#### Delete condition by id:

`
curl -X DELETE "http://localhost:7007/api/permission/roles/conditions/1" -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v
`

### DIFF
--- a/plugins/rbac-backend/docs/apis.md
+++ b/plugins/rbac-backend/docs/apis.md
@@ -454,11 +454,11 @@ Returns:
 
 ## Conditions
 
-The Backstage permission framework provides conditions, and the RBAC backend plugin supports this feature. Conditions work like content filters for Backstage resources (provided by plugins). The RBAC backend plugin checks if a user has access to a particular resource. If the user has access to the resource, the RBAC backend API delegates the condition for this resource to the corresponding plugin by plugin ID.
+The Backstage permission framework provides conditions, and the RBAC backend plugin supports this feature. Conditions work like content filters for Backstage resources (provided by plugins). The RBAC backend API stores conditions assigned to the role in the database. When a user requests access to the frontend resources, the RBAC backend API searches for corresponding conditions and delegates the condition for this resource to the corresponding plugin by its plugin ID. If a user was assigned to multiple roles, and each of these roles contains its own condition, the RBAC backend merges conditions using the anyOf criteria.
 
-The corresponding plugin analyzes conditional parameters and makes a decision about which part of the content the user should see. Consequently, the user can view not all resource content but only some allowed parts. The RBAC backend plugin supports conditions on a generic level - conditions are bound to all roles but none to specific roles.
+The corresponding plugin analyzes conditional parameters and makes a decision about which part of the content the user should see. Consequently, the user can view not all resource content but only some allowed parts. The RBAC backend plugin supports conditions bounded to the RBAC role.
 
-A Backstage condition consists of a parameter or an array of parameters joined by criteria. The list of supported conditional criteria includes:
+A Backstage condition can be a simple condition with a rule and parameters. But also a Backstage condition could consists of a parameter or an array of parameters joined by criteria. The list of supported conditional criteria includes:
 
 - allOf
 - anyOf
@@ -468,12 +468,14 @@ The plugin defines the supported condition parameters. API users can retrieve th
 
 The structure of the condition JSON object is as follows:
 
-| Json field   | Description                                                           | Type   |
-| ------------ | --------------------------------------------------------------------- | ------ |
-| result       | Always has the value "CONDITIONAL"                                    | String |
-| pluginId     | Corresponding plugin ID (e.g., "catalog")                             | String |
-| resourceType | Resource type provided by the plugin (e.g., "catalog-entity")         | String |
-| conditions   | Condition JSON with parameters or array parameters joined by criteria | JSON   |
+| Json field        | Description                                                           | Type         |
+| ----------------- | --------------------------------------------------------------------- | ------------ |
+| result            | Always has the value "CONDITIONAL"                                    | String       |
+| roleEntityRef     | String entity reference to the RBAC role ('role:default/dev')         | String       |
+| pluginId          | Corresponding plugin ID (e.g., "catalog")                             | String       |
+| permissionMapping | Array permission actions (['read', 'update', 'delete'])               | String array |
+| resourceType      | Resource type provided by the plugin (e.g., "catalog-entity")         | String       |
+| conditions        | Condition JSON with parameters or array parameters joined by criteria | JSON         |
 
 ### GET </plugins/condition-rules>
 
@@ -633,7 +635,7 @@ For example, consider a condition without criteria: displaying catalogs only if 
 - criteria: in this example, criteria are not used since we need to use only one conditional parameter
 - params: from the schema, it is evident that it should be an object named "claims" with a string array. This string array constitutes a list of user or group string entity references.
 
-Based on the above schema:
+Based on the above schema condition is:
 
 ```json
 {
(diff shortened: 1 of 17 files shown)


divyanshiGupta:
@AndrienkoAleksandr is there going to be an api endpoint for getting all the conditions that were applied while role creation? Like we have for the policies `/api/permission/policies/role/${namespace}/${name}` so that we can show the existing data to users? Also since we are updating/deleting conditions based on the id we will need that id for each condition so that we can enable update/delete from UI. Maybe this id can be part of the api response for getting all conditions for a specific role?

divyanshiGupta:
@AndrienkoAleksandr after these changes will it be possible to just update the permission policy without having to update the condition that was applied after creating that policy, similar to what you have shown in the description above?

divyanshiGupta:
> @AndrienkoAleksandr after these changes will it be possible to just update the permission policy without having to update the condition that was applied after creating that policy, similar to what you have shown in the description above?

@AndrienkoAleksandr wdyt about this?

AndrienkoAleksandr:
> @AndrienkoAleksandr after these changes will it be possible to just update the permission policy without having to update the condition that was applied after creating that policy, similar to what you have shown in the description above?
@AndrienkoAleksandr wdyt about this?

I think condition should be executed without corresponding permission policy check. I fixed that in the pull request and I removed  unnecessary permission policy creation from description:

~~#### Let's provide read catalog permission:~~

~~curl -X POST "http://localhost:7007/api/permission/policies" -d '[{"entityReference": "role:default/test", "permission": "catalog-entity", "policy": "read", "effect":"allow"}]' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v~~

I remember we consider to make this  fix in the working group meeting thanks you reminded me

divyanshiGupta:
> > @AndrienkoAleksandr after these changes will it be possible to just update the permission policy without having to update the condition that was applied after creating that policy, similar to what you have shown in the description above?
> > @AndrienkoAleksandr wdyt about this?
> 
> I think condition should be executed without corresponding permission policy. I fixed that in the pull request and I removed unnecessary permission policy creation from description:
> 
> ~#### Let's provide read catalog permission:~
> 
> ~curl -X POST "http://localhost:7007/api/permission/policies" -d '[{"entityReference": "role:default/test", "permission": "catalog-entity", "policy": "read", "effect":"allow"}]' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v~
> 
> I remember we consider to make this fix in the working group meeting thanks you reminded me

@AndrienkoAleksandr you mean that current api for creating permission policies which is `curl -X POST "http://localhost:7007/api/permission/policies" -d '[{"entityReference": "role:default/test", "permission": "catalog-entity", "policy": "read", "effect":"allow"}]' -H "Content-Type: application/json" -H "Authorization: Bearer $token" -v` will remain as if and can be updated irrespective of the condition applied right? Or is there some changes in the api? Just want to confirm if permission policies and conditions are independent of each other or not.

AndrienkoAleksandr:
> @divyanshiGupta Just want to confirm if permission policies and conditions are independent of each other or not.

Yes, these APIs are independent. I didn't touch the simple permission policy API (http://some-host:port/api/permission/policies). I changed the code evaluation conditional permissions as follows:

- If a user creates a condition and doesn't have a simple permission with the resource type then the backend will simply find this condition and evaluate it (in this case, the backend sends the condition to the corresponding plugin, where the plugin analyzes the condition and renders filtered data)
- If a user has a simple permission policy with the resource type and doesn't have any conditions, then the backend will evaluate this simple permission.
- If a user has both a simple permission policy with a resource type and a conditional policy, then the backend will evaluate only the condition. In this PR, I provided higher priority for the condition compared to the simple policy.

PatAKnight (changes requested):
Looks good so far, just a couple of things. 

Can we also get some updates to the [api docs](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/apis.md#conditions)?

Also, wouldn't this be considered a breaking change? We are changing the conditions endpoints and the old ones would no longer be valid.

AndrienkoAleksandr:
I updated doc. Additionally, I made some modifications to simplify frontend requests by hiding permission names. I'm not sure if the frontend will always have this information.

> @PatAKnight : Also, wouldn't this be considered a breaking change? We are changing the conditions endpoints and the old ones would no longer be valid.

Yes, it is, but unfortunately, we couldn't make migrations for old conditions.

PatAKnight:
Do you mind updating your commit message from `fix(rbac): rework condition policies to bound them to RBAC roles` to `fix(rbac)!: rework condition policies to bound them to RBAC roles`. Since this is a breaking change. 

Also, it looks like you have a linting error which is causing a failure.

AndrienkoAleksandr:
> @PatAKnight: Do you mind updating your commit message from fix(rbac): rework condition policies to bound them to RBAC roles to fix(rbac)!: rework condition policies to bound them to RBAC roles. Since this is a breaking change.
Also, it looks like you have a linting error which is causing a failure.

Done.

PatAKnight (commented):
Last request, but can you also update the [endpoints](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/apis.md#post-condition) in the documentation?

AndrienkoAleksandr:
> Last request, but can you also update the [endpoints](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/apis.md#post-condition) in the documentation?

Done.

PatAKnight (approved):
/lgtm

PR #1331, opened 2024-03-13
feat(kiali): metrics tab

josunect:
Inbound/Outbound Metrics tab for Workloads/Services/Apps

image

Note: 

- The backstage tabs were replaced because there was not way to change the override: hidden and for the large selects from Input/outbound metrics the style was not accurate. 
- The way the API stringify the parameters was not working properly for the array parameters. the arrays are stringified as `param=value1,value2`, but kiali needs `param[]=value, param[]=value2` so there is a new method to generate the parameters, `getCustomParams`, when the optional argument `customParams` for newRequest is `true`.


leandroberetta (approved):
lgtm

PR #1352, opened 2024-03-19
feat(notifications): require users and groups to be of the form namespace/name

ydayagi:
[FLPATH-1087](https://issues.redhat.com//browse/FLPATH-1087)
https://issues.redhat.com/browse/FLPATH-1087

### DIFF
--- a/plugins/notifications-backend/src/service/handlers.ts
+++ b/plugins/notifications-backend/src/service/handlers.ts
@@ -30,31 +30,23 @@ export async function createNotification(
   let isUser = false;
 
   // validate users
-  if (Array.isArray(req.targetGroups) && req.targetGroups.length > 0) {
+  if (
+    req.targetGroups &&
+    Array.isArray(req.targetGroups) &&
+    req.targetGroups.length > 0
+  ) {
     isUser = true;
-    const promises = req.targetGroups.map(group => {
-      return catalogClient.getEntityByRef(`group:${group}`).then(groupRef => {
-        if (!groupRef) {
-          throw new Error(`group '${group}' does not exist`);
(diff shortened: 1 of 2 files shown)


plugins/notifications-backend/src/service/handlers.ts
```diff
@@ -30,31 +30,23 @@ export async function createNotification(
   let isUser = false;
 
   // validate users
-  if (Array.isArray(req.targetGroups) && req.targetGroups.length > 0) {
+  if (
+    req.targetGroups &&
```
mareklibra, line 34: I think we do not need this check since it is included in the `Array.isArray()`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

plugins/notifications-backend/src/service/handlers.ts
```diff
@@ -30,31 +30,23 @@ export async function createNotification(
...
-    });
-
-    await Promise.all(promises);
+    await validateUsersGroups(true, req.targetGroups, catalogClient);
   }
 
   // validate groups
-  if (Array.isArray(req.targetUsers) && req.targetUsers.length > 0) {
+  if (
+    req.targetUsers &&
```
mareklibra, line 44: likewise

plugins/notifications-backend/src/service/handlers.ts
mareklibra: Or is it just to make the TypeScript happy?

mareklibra (approved):

mareklibra (approved):

PR #1353, opened 2024-03-19
chore(orchestrator): set npx packages version for openapi generation

gciavarrini:
Improve the openapi generation [script](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator-common/scripts/openapi.sh#L5-L7) 

* use the `--yes` flag
so that the script doesn't need to prompt the user for the installation 
* fix a version to the packages that we are using

Solve [FLPATH-1072](https://issues.redhat.com/browse/FLPATH-1072)

### DIFF
--- a/plugins/orchestrator-common/scripts/openapi.sh
+++ b/plugins/orchestrator-common/scripts/openapi.sh
@@ -8,9 +8,9 @@ DEFINITION_FILE="./src/auto-generated/api/definition.ts"
 METADATA_FILE="./src/auto-generated/.METADATA.sha1"
 
 openapi_generate() {
-    npx openapi-typescript ${OPENAPI_SPEC_FILE} -o ${SCHEMA_FILE}
-    npx openapi-generator-cli generate -g asciidoc -i ./src/openapi/openapi.yaml -o ./src/auto-generated/docs/index.adoc
-    npx yaml2json -f ${OPENAPI_SPEC_FILE}
+    npx --yes openapi-typescript ${OPENAPI_SPEC_FILE} -o ${SCHEMA_FILE}
+    npx --yes openapi-generator-cli generate -g asciidoc -i ./src/openapi/openapi.yaml -o ./src/auto-generated/docs/index.adoc
+    npx --yes yaml2json -f ${OPENAPI_SPEC_FILE}
 
     OPENAPI_SPEC_FILE_JSON=$(tr -d '[:space:]' < "$(dirname $OPENAPI_SPEC_FILE)"/openapi.json)
     cat << EOF > ${DEFINITION_FILE}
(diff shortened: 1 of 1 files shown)


plugins/orchestrator-common/scripts/openapi.sh
```diff
@@ -8,9 +8,9 @@ DEFINITION_FILE="./src/auto-generated/api/definition.ts"
 METADATA_FILE="./src/auto-generated/.METADATA.sha1"
 
 openapi_generate() {
-    npx openapi-typescript ${OPENAPI_SPEC_FILE} -o ${SCHEMA_FILE}
-    npx openapi-generator-cli generate -g asciidoc -i ./src/openapi/openapi.yaml -o ./src/auto-generated/docs/index.adoc
-    npx yaml2json -f ${OPENAPI_SPEC_FILE}
+    npx --yes openapi-typescript@6.7.5 ${OPENAPI_SPEC_FILE} -o ${SCHEMA_FILE}
+    npx --yes @openapitools/openapi-generator-cli@v2.13.1 generate -g asciidoc -i ./src/openapi/openapi.yaml -o ./src/auto-generated/docs/index.adoc
+    npx --yes yaml2json -f ${OPENAPI_SPEC_FILE}
```
caponetto, line 13: What about a version for `yaml2json`?

jkilzi, line 13:
```suggestion
    npx --yes yaml2json@1.0.2 -f ${OPENAPI_SPEC_FILE}
```

jkilzi (commented):
WDYT?

plugins/orchestrator-common/scripts/openapi.sh
gciavarrini: It doesn't work.
that yaml2json belongs to https://www.npmjs.com/package/js-yaml-cli
I tried to set the version but i failed: i think the package is too old and doesn't have support fro this. 
I tried to change package but i didn't find anything that works (they required older versions of nodejs for example)

gciavarrini: See my comment https://github.com/janus-idp/backstage-plugins/pull/1353#discussion_r1530457156

jkilzi: Try this instead:  
`npx --yes --package=js-yaml-cli@0.6.0 -- yaml2json -f ${OPENAPI_SPEC_FILE}`

jkilzi (changes requested):

plugins/orchestrator-common/scripts/openapi.sh
gciavarrini: This works! Addressed
Thanks for the help

y0n1 (approved):

caponetto (approved):

PR #1383, opened 2024-03-25
chore(prettier): fix prettier lint issue in test check

karthikjeeyar:
**Description:**

Fix the prettier lint issue in dependabot.yml file which is causing the Test checks to fails in all PR

cc: @invincibleJai


invincibleJai:
Thanks @karthikjeeyar, for the fix

cc @nickboldt

PR #1387, opened 2024-03-25
feat(rbac): load filtered policies before enforcing

PatAKnight:
## Description

This PR aims to utilize Casbin's subset policy loading. This approach allows us to reduce the number of permission policies and roles checked during enforcement, consequently decreasing the time it takes to evaluate the enforced permissions.

## Special notes to the Reviewer

I've refactored the permission policy tests as they were failing. I've created functions for the adapter, enforcer, enforcer delegate, and RBAC permission policies to enable their reuse throughout the tests. Additionally, I've updated all adapters from `FileAdapter` to `CasbinDBAdapterFactory` to leverage the `TypeORMAdapter`. This change was necessary because the `FileAdapter` cannot handle filtered policies, unlike the `TypeORMAdapter`.

### DIFF
--- a/plugins/rbac-backend/src/__fixtures__/data/valid-csv/basic-and-resource-policies.csv
+++ b/plugins/rbac-backend/src/__fixtures__/data/valid-csv/basic-and-resource-policies.csv
@@ -1,0 +1,21 @@
+# ========== basic type permission policies ========== #
+# case 1
+p, user:default/known_user, test.resource.deny, use, deny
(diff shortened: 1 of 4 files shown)


AndrienkoAleksandr (approved):
I reviewed and tested pr - works fine. Sonar cloud displays error about duplication for non rbac-backend plugin code.

AndrienkoAleksandr (approved):

PR #1413, opened 2024-03-27
fix(orchestrator): fixes v2/instances endpoint

jkilzi:
The `start` and `end` fields are annotated as `strings` instead of `Date`. The format of the string is provided as a hint for the developer as a comment (date-time)  due to limitations in the OpenAPI code generator.


jkilzi:
The "Build all packages" task is failing due to a type-check error in our code base, unrelated to the changes I've made.
@caponetto [PTAL](https://github.com/janus-idp/backstage-plugins/actions/runs/8449188768/job/23142776579?pr=1413#step:3:585) 🙏

PR #1414, opened 2024-03-27
fix(orchestrator): fixes v2/instances endpoint

jkilzi:
The `start` and `end` fields are annotated as `strings` instead of `Date`.

Addresses:
- https://issues.redhat.com/browse/FLPATH-1111


plugins/orchestrator-common/src/openapi/openapi.yaml
```diff
@@ -430,8 +430,10 @@ components:
           $ref: '#/components/schemas/ProcessInstanceStatusDTO'
         start:
           type: string
+          format: 'date-time'
         end:
           type: string
+          format: 'date-time'
```
gciavarrini, line 436: I had to remove this because @condawrangler had this error while generating the client `FAILED api_tests/test_orchestrator.py::test_get_workflow_run_instances - ValueError: invalid literal for int() with base 10: b'2/27'` the issue may be related to the fact that the format we have doesn't represent a valid date-time format for the openapi spec.

jkilzi: This is Python specific, how is that client being generated?

jkilzi: We can remove it on our side is not going to make a big difference. It's mostly for hinting about the straing's format

gciavarrini: >  This is Python specific, how is that client being generated?
I don't have generation details, however I think we should consider this scenario. Is there a formal specification about the date-time format that we can expect? IMO, Customers should be able to generate successfully clients from openApi spec

jkilzi: https://swagger.io/docs/specification/data-models/data-types/#string:~:text=date%2Dtime%20%E2%80%93%20the%20date%2Dtime%20notation%20as%20defined%20by%20RFC%203339%2C%20section%205.6%2C%20for%20example%2C%202017%2D07%2D21T17%3A32%3A28Z

jkilzi: @gciavarrini I just copy-pasted the entire openapi.yaml into https://editor.swagger.io/ and exported a pthon client without problem (I only had to change the top openapi version from 3.1.0 to 3.0.0, but I pressume it is not really that important)

jkilzi: Removed the format: date-time from the spec file. Addressed in https://github.com/janus-idp/backstage-plugins/commit/e29ed012e7ff60dcb969ff7de590075e574fbad7

caponetto (approved):

caponetto:
I'll wait for Gloria's approval.

gciavarrini (approved):
lgtm

PR #1418, opened 2024-03-27
refactor: experimental `embed-as-dependencies` option for the backend

davidfestal:
This PR :
- adds an experimental `embed-as-dependencies` option in the `export-dynamic-plugin`  command for the backend plugins that provides an alternate way of exporting backend plugins as dynamic. In some time it should become the default way. To summarize, it includes *embedded* packages as private dependencies (ending in the plugin private `node_modules`) instead of embedding them in the generated code. This is much more robust, enable a better validation of the plugin.
- adds a `--dynamic-plugins-root` option, which can used when using the `--dev` argument to add the dynamic plugin folder into the dynmaic plugins root of a local `backstage-showcase` instance. This now works the same for frontend and backend plugins.

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/dev.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/dev.ts
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { PackageRoleInfo } from '@backstage/cli-node';
+import { ConfigReader } from '@backstage/config';
+import { loadConfig } from '@backstage/config-loader';
+
+import chalk from 'chalk';
+import { OptionValues } from 'commander';
+import * as fs from 'fs-extra';
+
+import path from 'path';
+
+import { paths } from '../../lib/paths';
+import { Task } from '../../lib/tasks';
+
+export async function applyDevOptions(
+  opts: OptionValues,
+  pkgName: string,
+  role: PackageRoleInfo,
+  target: string,
+) {
+  if (opts.dev) {
+    if (role.platform === 'node') {
+      await fs.ensureSymlink(
+        paths.resolveTarget('src'),
+        path.resolve(target, 'src'),
+        'dir',
+      );
+    }
+
+    let dynamicPluginsRootPath = opts.dynamicPluginsRoot as string | undefined;
+    let shouldSymlink = false;
+    if (!dynamicPluginsRootPath) {
+      shouldSymlink = true;
+      const appConfigs = await loadConfig({
+        configRoot: paths.targetRoot,
+        configTargets: [],
+      });
+      const fullConfig = ConfigReader.fromConfigs(appConfigs.appConfigs);
+
+      const dynamicPlugins = fullConfig.getOptional('dynamicPlugins');
+      if (
+        typeof dynamicPlugins === 'object' &&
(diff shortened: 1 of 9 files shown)


kadel:
@davidfestal this should bump minor version of the cli package

packages/cli/src/commands/export-dynamic-plugin/dev.ts
```diff
@@ -0,0 +1,98 @@
...
+        paths.resolveTarget('src'),
+        path.resolve(target, 'src'),
+        'dir',
+      );
+    }
+
+    let dynamicPluginsRootPath = opts.dynamicPluginsRoot as string | undefined;
+    let shouldSymlink = false;
+    if (!dynamicPluginsRootPath) {
+      shouldSymlink = true;
```
kadel, line 48: @davidfestal I feel like you probably mentioned this to me during the call, but why not allow symlinking when using `--dynamic-plugins-root`?

kadel:
I've tried to test new --embed-as-dependencies with a couple of packages.

```
▶ backstage-plugins/plugins/aap-backend 

▶ npx janus-cli package export-dynamic-plugin --embed-as-dependencies --dev  --dynamic-plugins-root ~/Code/backstage-showcase/dynamic-plugins-root
Building main package
  executing     yarn build ✖

Error: No declaration files found at ../../dist-types/plugins/aap-backend/src/index.d.ts, be sure to run yarn tsc to generate .d.ts files before packaging

error Command failed with exit code 1.
yarn run v1.22.22
$ backstage-cli package build
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Error: Failed to execute command 'yarn build', Error: Command failed: yarn build

Error: No declaration files found at ../../dist-types/plugins/aap-backend/src/index.d.ts, be sure to run yarn tsc to generate .d.ts files before packaging

error Command failed with exit code
```

after running `yarn tsc` it still fails

```
▶ npx janus-cli package export-dynamic-plugin --embed-as-dependencies --dev  --dynamic-plugins-root ~/Code/backstage-showcase/dynamic-plugins-root
Building main package
  executing     yarn build ✔
Packing main package
  moving @backstage/backend-common to peerDependencies
  moving @backstage/backend-plugin-api to peerDependencies
  moving @backstage/backend-tasks to peerDependencies
  moving @backstage/catalog-model to peerDependencies
  moving @backstage/config to peerDependencies
  moving @backstage/plugin-catalog-node to peerDependencies
  moving @backstage/backend-dynamic-feature-service to peerDependencies
Installing private dependencies of the main package
  executing     yarn install --production --frozen-lockfile ✔

Error: Backend plugin is not valid for dynamic loading: it should either export a BackendFeature or BackendFeatureFactory as default export, or export a const dynamicPluginInstaller: BackendDynamicPluginInstaller field as dynamic loading entrypoint
```


When using the old embedding method, everything works fine.


I'm still trying to understand all that is going on here. Is it possible that I'm using it wrong?

packages/cli/src/commands/export-dynamic-plugin/dev.ts
davidfestal: Because it might be outside of the current backstage monorepo, in which case (due to how the NodeJS module loader resolves symlinks), the dynamically-loaded plugins will not find the shared dependencies of the backstage instance you want to use.

davidfestal:
> after running `yarn tsc` it still fails

Oh, good catch indeed. That's because I added a test to verify that the plugin provides to expected entrypoints.
But I didn't test on plugins which provide new backend system entrypoint point in the `alpha` entrypoint.
And that's the case of this plugin.

I'll fix this.

davidfestal:
> > after running `yarn tsc` it still fails
> 
> Oh, good catch indeed. That's because I added a test to verify that the plugin provides to expected entrypoints. But I didn't test on plugins which provide new backend system entrypoint point in the `alpha` entrypoint. And that's the case of this plugin.
> 
> I'll fix this.

Fixed.

davidfestal:
> @davidfestal this should bump minor version of the cli package

In fact, I'd like to avoid it to finally indirectly bump the version of all plugins which have a dev dependency on the `janus-idp/cli` package.

Do you have any idea how it will behave ?

kadel:
> In fact, I'd like to avoid it to finally indirectly bump the version of all plugins which have a dev dependency on the `janus-idp/cli` package.
> 
> Do you have any idea how it will behave ?

Oh. I see. Let me test test that. I would assume that if you bump the version manually (not via semantic-release-bot) it should not update it other packages. But let me verify that first

PR #1423, opened 2024-03-28
fix: sync yarn.lock

invincibleJai:
Fixes: https://issues.redhat.com/browse/RHIDP-1801

**Description:** 

Test failures in Backstage-plugins PRs

sample failed jobs:

https://github.com/janus-idp/backstage-plugins/actions/runs/8456520631/job/23166479029#step:3:690 
https://github.com/janus-idp/backstage-plugins/actions/runs/8456565568/job/23166620071?pr=1422#step:3:886 

failure reason based on logs:

```
FAIL src/components/DynamicPluginsTable/DynamicPluginsTable.test.tsx
● Test suite failed to run
 
 Jest encountered an unexpected token
 
Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
 
Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.
 
By default "node_modules" folder is ignored by transformers.
 
Here's what you can do:
• If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
• If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
• To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
• If you need a custom transformation specify a "transform" option in your config.
• If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
```
 
**Solution:** 

`Yarn.lock` in main is out of sync and there seem to be issue with jest transformers based on the log (default jest transforms included with the `@backstage/cli`) and i am not sure what was wrong with yarn.lock updated with the commit bdd25872d24a23754d014337a4ad1ae87ed563d1 as yarn.lock is too huge, if some one figure out pls let me know :)


nickboldt (changes requested):
please provide details about this change and link to the related JIRA. https://issues.redhat.com/browse/RHIDP-1750

Since the diff is huge in the yarn lock, I have two asks:

* can you list some of the interesting/relevant changes? eg., jest related stuff?
* should we also pin test related devDependencies in the root pacakge.json so that if the yanr lock is deleted and recreated, bad things won't happen?

invincibleJai:
> https://issues.redhat.com/browse/RHIDP-1750

@nickboldt the issue you linked is different from what I attempted to fix here the issue fixed is for https://issues.redhat.com/browse/RHIDP-1801

caponetto:
@invincibleJai Could we type this PR as a `fix` instead of `chore`?
This way, the semantic releases that didn't complete can be retriggered.

nickboldt:
I'm happy with this PR and the explanation.

Opened https://issues.redhat.com/browse/RHIDP-1812 for followup work re: pinning more deps so lock files don't explode. :D

PR #1446, opened 2024-04-04
feat(rbac): add the ability to cache information

PatAKnight:
## Description

As part of our ongoing work, this PR introduces the ability to cache information for the role manager to use. This should help further alleviate the problem of constantly having to rebuild the group hierarchy graph each time there is a request to the role manager.

## Fixes
 - Fixes https://issues.redhat.com/browse/RHIDP-1377

## Special note to the reviews

There are a lot of changes to the test files. The reason for this is because I refactored some of the tests to bring out common functionality to a utils file. Now, we have a `utils.test.ts` file that stores some of the common functions that are spread throughout some of our tests.

I didn't refactor all of the tests, this was more of a starting point. There still needs to be some ongoing effort to

### DIFF
--- a/plugins/rbac-backend/src/role-manager/role-cache.ts
+++ b/plugins/rbac-backend/src/role-manager/role-cache.ts
@@ -5,9 +5,19 @@ export class RoleCache {
   private maxAge: number;
   constructor(maxEntries?: number, maxAge?: number) {
     this.maxEntries = maxEntries || 100;
-    this.maxAge = maxAge || 60 * 60 * 1000; // 1 hour (60 minutes * 60 seconds * 1000 milliseconds) <-- double check this math? I think we want is in millisecond
+    this.maxAge = maxAge || 60 * 60 * 1000; // 1 hour (60 minutes * 60 seconds * 1000 milliseconds)
   }
 
+  /**
+   * get will return the cached users based on the key that is received.
+   * We also will update the cached place in the map and update its timestamp.
+   * This is because the implemented cache is based on Least Recently used
+   * and maps hold the order that keys are placed in.
+   * Updating the place in the order will ensure that we only remove the
+   * least recently used items whenever we run out of space.
+   * @param key The user that
+   * @returns The cache if we are able to find it
+   */
   public get(key: string): Set<string> | undefined {
     const hasKey = this.cache.has(key);
     if (!hasKey) return undefined;
@@ -19,6 +29,16 @@ export class RoleCache {
     return data;
   }
 
+  /**
+   * put will add the cache using the user as the key.
+   * It will set the value to the roles that the user is directly or indirectly
+   * attached to as well as set the time for this cache.
+   *
+   * If we are running out of space, we will remove the least recently used
+   * cache which will be at the end of the cache.
+   * @param key The user to be cached
+   * @param value The roles that are attached to the user
+   */
   public put(key: string, value: Set<string>) {
     if (this.cache.size >= this.maxEntries) {
       const keyToDelete = this.cache.keys().next().value;
@@ -30,10 +50,20 @@ export class RoleCache {
     this.cache.set(key, { data: value, timestamp: currentTime });
   }
 
+  /**
+   * delete removes the cache user.
+   * @param key The user to be removed
+   */
   public delete(key: string) {
     this.cache.delete(key);
   }
 
+  /**
+   * shouldUpdate checks if the cached user's time in the cache has reached
+   * the set max age.
+   * @param key The user cache that we are checking
+   * @returns True if the time has exceeded the max age
+   */
   public shouldUpdate(key: string): boolean {
     const currentTime = Date.now();
     const hasKey = this.cache.has(key);
@@ -45,5 +75,19 @@ export class RoleCache {
     }
 
     return false;
+  }
+
+  /**
+   * deleteCacheWithRole will remove cached users based on a role.
+   * This is useful for when we want to remove a particular role and invalidate
+   * any cache that is associated with said role.
+   * @param role The role that we will be invalidating
+   */
+  public deleteCacheWithRole(role: string): void {
+    for (const [key, value] of this.cache.entries()) {
+      if (value.data.has(role)) {
+        this.cache.delete(key);
+      }
+    }
   }
 }
--- a/plugins/rbac-backend/src/role-manager/role-cache.ts
+++ b/plugins/rbac-backend/src/role-manager/role-cache.ts
@@ -1,0 +1,48 @@
+export class GroupCache {
+  private cache: Map<string, { data: string[]; timestamp: number }> = new Map();
+  private maxEntries: number;
+  private maxAge: number;
+  constructor(maxEntries?: number, maxAge?: number) {
+    this.maxEntries = maxEntries || 100;
+    this.maxAge = maxAge || 60 * 60 * 1000; // 1 hour (60 minutes * 60 seconds * 1000 milliseconds) <-- double check this math? I think we want is in millisecond
+  }
+
+  public get(key: string): string[] | undefined {
+    const hasKey = this.cache.has(key);
+    if (!hasKey) return undefined;
+
+    const { data, timestamp } = this.cache.get(key)!;
+    this.cache.delete(key);
+    this.cache.set(key, { data, timestamp });
+
+    return data;
+  }
+
+  public put(key: string, value: string[]) {
+    if (this.cache.size >= this.maxEntries) {
+      const keyToDelete = this.cache.keys().next().value;
+      this.cache.delete(keyToDelete);
+    }
+
+    const currentTime = Date.now();
+
+    this.cache.set(key, { data: value, timestamp: currentTime });
+  }
+
+  public shouldUpdate(key: string): boolean {
+    const currentTime = Date.now();
+    const hasKey = this.cache.has(key);
+    if (!hasKey) return false;
+
+    const { timestamp } = this.cache.get(key)!;
+    if (currentTime - timestamp > this.maxAge) {
+      return true;
+    }
+
+    return false;
+  }
+
+  public isEnabled(): boolean {
+    return true;
+  }
+}
--- a/plugins/rbac-backend/src/role-manager/role-cache.ts
+++ b/plugins/rbac-backend/src/role-manager/role-cache.ts
@@ -30,6 +30,10 @@ export class RoleCache {
     this.cache.set(key, { data: value, timestamp: currentTime });
   }
 
+  public delete(key: string) {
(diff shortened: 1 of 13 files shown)


plugins/rbac-backend/src/role-manager/role-cache.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.maxEntries = maxEntries || 100;
+    this.maxAge = maxAge || 60 * 60 * 1000; // 1 hour (60 minutes * 60 seconds * 1000 milliseconds) <-- double check this math? I think we want is in millisecond
+  }
+
+  public get(key: string): Set<string> | undefined {
+    const hasKey = this.cache.has(key);
+    if (!hasKey) return undefined;
+
+    const { data, timestamp } = this.cache.get(key)!;
+    this.cache.delete(key);
```
AndrienkoAleksandr, line 16: Why do you need to delete and set the same data?

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -86,47 +98,29 @@ export class BackstageRoleManager implements RoleManager {
...
-          cycles,
-        )}. Admin/(catalog owner) have to fix it to make RBAC permission evaluation correct for groups: ${JSON.stringify(
-          cycles,
-        )}`,
-      );
+    // If the cache does not exist we need to build the graph and use it
+    // Otherwise we need to use the cache instead
+    const memo = await this.buildGraph(name1);
 
+    if (!this.detectCycleError(memo, name1)) {
```
AndrienkoAleksandr, line 119: This code checks cycle dependencies only on build cache, but doesn't when cache exists.

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -227,9 +207,89 @@ export class BackstageRoleManager implements RoleManager {
...
+      this.catalogApi,
+      this.catalogDBClient,
+    );
+    await memo.buildUserGraph(memo);
+
+    memo.debugNodesAndEdges(this.log, name1);
+    return memo;
+  }
+
+  private parseRoles(userAndParentGroups: string[]): string[] {
```
AndrienkoAleksandr, line 262: Perhaps the name "matchRoles" would be better?

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -168,34 +162,20 @@ export class BackstageRoleManager implements RoleManager {
...
-            .filter(role => {
-              return role.name.startsWith('role:default');
-            })
-            .map(role => role.name);
-
-          allRoles.push(...rolesEntityRefs);
-        }
-      }
-      return Promise.resolve(allRoles);
+      const allRoles = this.parseRoles(userAndParentGroups);
```
AndrienkoAleksandr, line 177:
```suggestion
      return this.parseRoles(userAndParentGroups);
```

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -227,9 +207,89 @@ export class BackstageRoleManager implements RoleManager {
     return newRole;
   }
 
+  private removeRole(name: string): void {
```
AndrienkoAleksandr, line 210: This code doesn't handle RoleCache

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -227,9 +207,89 @@ export class BackstageRoleManager implements RoleManager {
...
+          .map(role => role.name);
+
+        allRoles.push(...rolesEntityRefs);
+      }
+    }
+    return allRoles;
+  }
+
+  private async cacheResults(name1: string): Promise<boolean> {
+    if (this.roleCache?.isEnabled()) {
```
AndrienkoAleksandr, line 280: I suppose we could use `if (this.roleCache)` instead of `if (this.roleCache?.isEnabled())` and remove the redundant boolean flag `sEnabled()` from RoleCache.

AndrienkoAleksandr:
Hello, @PatAKnight. I was pondering the necessity of caching roles. Considering that our API both provides and manages roles, and we have precise control over their application and removal, it seems we may not require an expiration mechanism to update this information. Additionally, roles are already stored in the map RoleList, as illustrated:

https://github.com/janus-idp/backstage-plugins/blob/98fbae66d28da70d7e63500562633e3d4cb02de2/plugins/rbac-backend/src/role-manager/role-manager.ts#L13

_This map, though basic, serves as a form of cache itself._ Could you kindly elaborate on the necessity for an additional cache such as RoleCache? Thank you for your insights.

AndrienkoAleksandr:
Oh, sorry, I see. You are caching roles per user to avoid rebuild group graph each time.

plugins/rbac-backend/src/role-manager/role-cache.ts
PatAKnight: I am using an LRU (least recently used) cache with a map. Maps preserve the order of the items that are placed in them. Removing the item allows me to reset its place in the map whenever we get the item.

plugins/rbac-backend/src/role-manager/role-manager.ts
PatAKnight: No the check is still happening whenever we cache, it instead will log the error and not cache the results. See this [line](https://github.com/janus-idp/backstage-plugins/blob/303801f63bf81f15eb4a67326d98655d1111b1e2/plugins/rbac-backend/src/role-manager/role-manager.ts#L284). This will still allow us to log the warning and we will return false if the role was not cached. 

The only downside to this, that I am just now noticing, is that the graph will be built a second time and the warning will be logged again.

plugins/rbac-backend/src/role-manager/role-manager.ts
PatAKnight: Sounds good with me.

plugins/rbac-backend/src/role-manager/role-manager.ts
PatAKnight: Will fix

plugins/rbac-backend/src/role-manager/role-manager.ts
PatAKnight: Yeah, I was going to do something with the `isEnabled` flag from `RoleCache` but ended up settling for a different approach with the configuration values. This is the left overs from it that I didn't remove. Will fix.

PatAKnight:
Yep, the idea is to prevent us from rebuilding the graph every single time by caching the roles using `RoleCache`. The way that I was envisioning it is that the `allRoles` cache is all of the grouping policies that the `RoleManager` is aware of, whereas the `roleCache` are roles that are verified to not have any issues when building the group hierarchy graph. Hopefully, that makes sense.

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -227,9 +212,89 @@ export class BackstageRoleManager implements RoleManager {
...
+          })
+          .map(role => role.name);
+
+        allRoles.push(...rolesEntityRefs);
+      }
+    }
+    return allRoles;
+  }
+
+  private async cacheResults(name1: string): Promise<boolean> {
```
AndrienkoAleksandr, line 284: I think we no longer need to return a boolean flag here

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -55,9 +62,16 @@ export class BackstageRoleManager implements RoleManager {
...
-   * deleteLink deletes the inheritance link between role: name1 and role: name2.
-   * aka role: name1 does not inherit role: name2 any more.
+   * deleteLink deletes the inheritance link between name1 and role: name2.
+   * aka name1 does not inherit role: name2 any more.
    * domain is a prefix to the roles.
+   * The link that is deleted is based on the defined grouping policies that
+   * are removed by the enforcer.
+   *
+   * ex. `g, name1, name2`.
+   * @param name1 User or group that will be assigned to a role
```
AndrienkoAleksandr, line 72: assigned -> removed from assignment

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -38,9 +38,16 @@ export class BackstageRoleManager implements RoleManager {
...
-   * aka role: name1 inherits role: name2.
+   * addLink adds the inheritance link between name1 and role: name2.
+   * aka name1 inherits role: name2.
    * domain is a prefix to the roles.
+   * The link that is established is based on the defined grouping policies that
+   * are added by the enforcer.
+   *
+   * ex. `g, name1, name2`.
+   * @param name1 User or group that will be assigned to a role
+   * @param name2 The role that will be created
```
AndrienkoAleksandr, line 49: created -> created or updated

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -55,9 +62,16 @@ export class BackstageRoleManager implements RoleManager {
...
-   * aka role: name1 does not inherit role: name2 any more.
+   * deleteLink deletes the inheritance link between name1 and role: name2.
+   * aka name1 does not inherit role: name2 any more.
    * domain is a prefix to the roles.
+   * The link that is deleted is based on the defined grouping policies that
+   * are removed by the enforcer.
+   *
+   * ex. `g, name1, name2`.
+   * @param name1 User or group that will be assigned to a role
+   * @param name2 The role that will be created
```
AndrienkoAleksandr, line 73: created -> deleted or updated

plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -218,6 +250,20 @@ export class BackstageRoleManager implements RoleManager {
...
+   * authorize a particular user.
+   *
+   * We iterate through the known roles within the cached tree to find
+   * groups that are attached to the group hierarchy graph.
+   *
+   * We will then compare those groups and their roles in the cached tree to the roles and
+   * the group hierarchy graph.
+   * @param memo The group hierarchy graph
+   * @param name2 The name of the role that we are authorizing with
+   * @returns True or false if the group is present in the tree and attached to the role
```
AndrienkoAleksandr, line 265: True or false if -> True if

plugins/rbac-backend/src/role-manager/role-manager.ts
AndrienkoAleksandr: Ok, I see we need it after your last changes.

PatAKnight:
Closing PR as it is outdated and we discussed that we are going to hold off on this feature for now.

PR #1478, opened 2024-04-08
fix(cli): fix the `embed-as-dependencies` option on the `export-dynamic-plugin` CLI...

davidfestal:
Fix the `embed-as-dependencies` option on the `export-dynamic-plugin` CLI when run on a wrapper plugin that doesn't use a monorepo structure.

Before this fix, this would fail, failing to resolve the wrapped plugin in the wrapper `node_modules` folder.


nickboldt (approved):
sure

PR #1486, opened 2024-04-09
feat(rbac): add the optional maxDepth feature

PatAKnight:
## Description

Introduces the optional configuration value `maxDepth` that will limit the depth that hierarchy group graphs are built. This should help reduce the amount of time building complex graphs and will help with performance.

## Fixes
- Fixes https://issues.redhat.com/browse/RHIDP-1376

## Special notes to the reviewer

I did the bare minimum for updating the tests because I have updates to tests in this [PR](https://github.com/janus-idp/backstage-plugins/pull/1446). I will rebase and fix the tests in this PR in the event the other is merged.

### DIFF
--- a/plugins/rbac-backend/src/role-manager/role-manager.ts
+++ b/plugins/rbac-backend/src/role-manager/role-manager.ts
@@ -1,6 +1,7 @@
 import { TokenManager } from '@backstage/backend-common';
 import { CatalogApi } from '@backstage/catalog-client';
 import { parseEntityRef } from '@backstage/catalog-model';
+import { Config } from '@backstage/config';
 
 import { RoleManager } from 'casbin';
 import { Knex } from 'knex';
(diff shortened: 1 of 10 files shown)


plugins/rbac-backend/src/role-manager/role-manager.ts
```diff
@@ -11,13 +12,17 @@ import { RoleList } from './role-list';
...
   constructor(
     private readonly catalogApi: CatalogApi,
     private readonly log: Logger,
     private readonly tokenManager: TokenManager,
     private readonly catalogDBClient: Knex,
+    private readonly config: Config,
   ) {
     this.allRoles = new Map<string, RoleList>();
+    const rbacConfig = this.config.getOptionalConfig('permission.rbac');
+    this.maxDepth = rbacConfig?.getOptionalNumber('maxDepth');
```
AndrienkoAleksandr, line 25: It would be nice to validate this value somewhere. It should be positive, like you described in the doc.

AndrienkoAleksandr (approved):

PR #1496, opened 2024-04-12
fix(feedback-docs): move out collapsed screenshots

yashoswalyo:
As @awanlin suggested here https://github.com/backstage/backstage/pull/24184\#discussion_r1560988756 it would be good if screenshots are not hidden


deshmukhmayur (approved):
LGTM.

PR #1512, opened 2024-04-15
chore(renovate): update renovate config

kim-tsao:
See https://issues.redhat.com/browse/RHIDP-1737

* Removed dependabot.yml so we can rely on renovate as our single config
* Migrated pinned version rules except for the ArgoCD dependency since that is a no-op in this repo (only relevant in showcase repo)
* Added integration with dependabot alerts (will need to disable security settings in repo once this is merged)


schultzp2020:
I think we should close PRs rather than maintaining individual deps in the config. (Closing PRs does the same function).

durandom:
> I think we should close PRs rather than maintaining individual deps in the config. (Closing PRs does the same function).

can you elaborate a bit? 

IUC renovate will not create another PR for a dependency update once we close the first one, which makes sense.
But then we rely on documenting the reasoning for closing the PR in the PR itself.
Also, for the next dep update, another PR would be opened.

Best case would be to have the dependency pinned or corrected, with an explanation, in the dependency file itself `package.json` - but I guess that's not always possible. 
If it's not possible or we only want to do this temporarily, I guess the `renovate.json` is a good place to capture these.
Always with the aim to keep the edits in here minimal and temporary.

Please correct me if this reasoning has flaws :)

.github/renovate.json
```diff
@@ -29,8 +34,25 @@
...
+    {
+      "description": "ignore updates to the axios to version that keycloak 18 needs", 
+      "matchCurrentVersion": "0.26.1",
+      "matchPackageName": ["axios"],
+      "enabled": "false",
+      "groupName": "Keycloak dependency"
+    }    
   ],
   "ignorePaths": ["**/dist-dynamic/**"],
-  "ignoreDeps": ["@roadiehq/backstage-plugin-argo-cd"]
```
nickboldt, line 35: why remove this? we want it pinned to 2.4.1 until we remove it in RHDH 1.4.

.github/renovate.json
```diff
@@ -18,6 +18,11 @@
       "matchManagers": ["github-actions"],
       "groupName": "GitHub Actions"
     },
+    {
+      "matchDepTypes": ["devDependencies"],
+      "matchUpdateTypes": ["patch", "minor"],
+      "groupName": ["DevDependencies (non-major)"]
```
nickboldt, line 24: can we add an automerge: true for these?

.github/renovate.json
```diff
@@ -18,6 +18,11 @@
       "matchManagers": ["github-actions"],
       "groupName": "GitHub Actions"
     },
+    {
```
nickboldt, line 21: you might also want a top-level entry
```
  "baseBranches": ["main", "/^1\\..*\\.x/"],
```
as in https://github.com/janus-idp/operator/pull/317/files#diff-f82f7d36a61d22f640c25bdb8b0435bf9cb38679d99d5f460753fa668ee7cdb3R9

.github/renovate.json
kim-tsao: It's not a dependency in this repo

.github/renovate.json
kim-tsao: yeah, I thought we wanted a "wait and see" approach

nickboldt:
> we rely on documenting the reasoning for closing the PR in the PR itself.
this is bad, things will get lost

>  renovate.json is a good place to capture these.
Yeah it's a central document for security rules, which supplements package.json and yarn.lock.

See also https://docs.renovatebot.com/dependency-pinning/#so-whats-best ... version pins + yarn.lock... for the horde!

kim-tsao:
I just disabled the dependabot security settings in this repo.  Will close up any dependabot PRs

.github/renovate.json
```diff
@@ -10,6 +10,7 @@
     ":pinOnlyDevDependencies"
   ],
   "labels": ["kind/dependency upgrade"],
+  "baseBranches": ["main", "/^1\\..*\\.x/"],
```
nickboldt, line 13: dependabot will complain about formatting and submit a new PR here, so to save a step:

```suggestion
    "baseBranches": [
    "main",
    "/^1\\..*\\.x/"
  ],
```

nickboldt (approved):

PR #1536, opened 2024-04-22
chore(deps): update to backstage 1.26.x

schultzp2020:
- Updates the plugins to 1.26
- Updates to the new backend system
- Syncs with the latest backstage frontend template.


schultzp2020:
@invincibleJai can you look into why the Quay tests are failing?

packages/backend/src/plugins/auth.ts
```diff
@@ -1,56 +0,0 @@
-import {
```
invincibleJai, line 1: @schultzp2020 are these files not needed or ?

schultzp2020: These files are no longer needed with the new backend.

PatAKnight:
I am curious on the thoughts of everyone with having this be synced with the frontend of showcase. On the one hand it is nice to be able to test for compatibility with showcase. However, having a generic backstage instance is also nice as there isn't near as much set up.

As an example, with the sync to showcase, RBAC is experiencing some issues with the TechDocs plugin and Backstage 1.25. So for me to test the problem, we need to: set up the frontend TechDocs plugin, set up the sign in page with GitHub, and set up GitHub auth in the backend. All before I can test if the changes that I have made worked.

schultzp2020:
>I am curious on the thoughts of everyone with having this be synced with the frontend of showcase. On the one hand it is nice to be able to test for compatibility with showcase. However, having a generic backstage instance is also nice as there isn't near as much set up (sometimes).

This is the generic backend. I basically copied the backend from a new backstage instance.

PatAKnight:
> This is the generic backend. I basically copied the backend from a new backstage instance. In the future it won't matter because it'll be in the community repo.

No, I see the generic backend. I am talking about the frontend not necessarily being generic. The reason that I was bringing this up was because of the process in setting up TechDocs in the frontend when testing the bug. It is not a major issue, just wondering if anyone else had an opinion on it.

schultzp2020:
> I am talking about the frontend not necessarily being generic. The reason that I was bringing this up was because of the process in setting up TechDocs in the frontend when testing the bug. It is not a major issue, just wondering if anyone else had an opinion on it.

Ah, I'm fine going either way. I thought it would be easier to make it dynamic but if not that's fine.

PatAKnight:
I am fairly flexible on either approach. We can make the frontend dynamic for now and see if anyone else has any opinions on it.

packages/app/src/components/DynamicRoot/DynamicRoot.tsx
```diff
@@ -0,0 +1,336 @@
...
+    plugin: BackstagePlugin;
+    module:
+      | React.ComponentType<{}>
+      | { [importName: string]: React.ComponentType<{}> };
+  }
+>;
+
+type EntityTabMap = Record<string, { title: string; mountPoint: string }>;
+
+export const DynamicRoot = ({
```
gashcrumb, line 39: I would not recommend going down this path of replicating the dynamic plugin implementation code in this repo.  A better approach would be to look into basing the e2e tests on using the backstage-showcase container image instead.

gashcrumb:
> I am curious on the thoughts of everyone with having this be synced with the frontend of showcase. On the one hand it is nice to be able to test for compatibility with showcase. However, having a generic backstage instance is also nice as there isn't near as much set up (sometimes).
> 
> As an example, with the sync to showcase, RBAC is experiencing some issues with the TechDocs plugin and Backstage 1.25. So for me to test the problem, we need to: set up the frontend TechDocs plugin, set up the sign in page with GitHub, and set up GitHub auth in the backend. All before I can test if the changes that I have made worked.

The backstage-showcase container could be used for this.  It'd be a matter of mounting the right files and directories in the right places.  Also maybe for the e2e tests some automation specific to the github actions may need to be created to perhaps manage the tasks that are normally done by the init container on openshift.   This would be a much better way to tackle this than replicating the frontend implementation.

gashcrumb:
One other option than the container would be to extract out the dynamic frontend implementation into it's own standalone package that could be cleanly applied to an existing backstage app instance.

schultzp2020:
@batzionb can you help look into why the Orchestrator plugin test is failing? https://github.com/janus-idp/backstage-plugins/actions/runs/8976420369/job/24653242141?pr=1536#step:3:716

packages/app/package.json
```diff
@@ -15,63 +15,58 @@
...
-    "@scalprum/react-test-utils": "0.0.4",
-    "@testing-library/jest-dom": "6.4.2",
-    "@testing-library/react": "14.2.1",
-    "@testing-library/user-event": "14.5.1",
-    "@types/node": "18.18.5",
-    "@types/react-dom": "18.2.18",
-    "cross-env": "7.0.3"
+    "@backstage/cli": "0.26.4",
+    "@backstage/test-utils": "1.5.4",
+    "@playwright/test": "1.32.3",
```
debsmita1, line 62: Is this added here on purpose ?
I ran into errors while executing the playwright tests
Here's my PR https://github.com/schultzp2020/backstage-plugins/pull/4

schultzp2020: No, this was included in the default frontend template.

PR #1571, opened 2024-04-26
chore(renovate): fix warnings, use matchDepPatterns instead

kim-tsao:
- Seeing warnings.  Looks like we are using the wrong field.  

    WARN: Use matchDepPatterns instead of matchPackagePatterns
    WARN: Error updating branch: update failure
    

- Disabled backstage core updates


nickboldt (approved):
seems legit

PR #1582, opened 2024-04-29
fix(feedback): add support for jira cloud instance

### LINKED ISSUE #1543: Feedback Plugin: add support for jira cloud instance
### Describe the bug
The current api call to get user details uses `/resp/api/2/user/search?username=<email>`.

The jira cloud instance does not support `username` query parameter.

### Expected Behavior

The plugin work without crashing

1. Configure an jira cloud instance to work with feedback plugin
2. Try to create new feedback from ui
3. The backend plugin crashes

yashoswalyo:
- fixes #1543
  Introduced a new field `integrations.jira.hostType` to determine if given jira instance is **server/cloud** 
- feat: add export-dynamic to script for RHDH
  Although I have added dynamic plugin support currently frontend plugin does not fully supports it.


plugins/feedback-backend/src/api/jiraApiService.ts
```diff
@@ -1,90 +1,105 @@
...
+ */
+export class JiraApiService {
+  private readonly hostType: string;
+  constructor(
+    private host: string,
+    private authToken: string,
+    hostType?: string,
+  ) {
+    if (!hostType) this.hostType = 'SERVER';
+    else this.hostType = 'CLOUD';
```
deshmukhmayur, line 16: This statement makes no sense, as it will use `CLOUD` even if the user specifies `hostType = 'SERVER'`due to the truthy check used.

Since the hostType is a string, why not just assign that to the `this.hostType` property?

```suggestion
    this.hostType = hostType ?? 'SERVER';
```

Or you can do the same in the contructor parameters by providing a default value:
```ts
  constructor(
    // ...
    private hostType: string = 'SERVER',
  ) {
```

deshmukhmayur (changes requested):

plugins/feedback-backend/src/api/jiraApiService.ts
yashoswalyo: Done

plugins/feedback-backend/app-config.janus-idp.yaml
```diff
@@ -3,8 +3,14 @@ feedback:
     jira:
       # Under this object you can define multiple jira hosts
       - host: ${JIRA_HOST_URL}
+        # Add your jira token along with Basic or Bearer type
         token: ${JIRA_TOKEN}
 
+        # (optional) Due to GDPR limitations on jira cloud instances
+        # set hostType: CLOUD to make some apis work
```
riginoommen, line 10:
```suggestion
        # set hostType: CLOUD to make Jira Cloud to work
```

yashoswalyo: done

riginoommen (approved):

PR #1586, opened 2024-04-30
fix(kiali): remove kiali-common from package.

aljesusg:
Remove kiali-common that was removed so time ago.


josunect (approved):
lgtm

nickboldt:
/approve

(because /lgtm is not enough to trigger openshift automerge)

PR #1590, opened 2024-04-30
fix(cli): add `resolve.fallback` for node dependencies in the `export-dynamic-plugin` CLI for frontend plugins.

davidfestal:
As spotted by the [CI workflow that exports all `@backstage/backstage@1.23.4` plugins to dynamic](https://github.com/davidfestal/backstage-plugins-dynamic-build/commit/e4c0079aac52758ca352ecf49b11767cb117f9e8#diff-e19aa0f2cb39e8dd6af3a8f7a7b998172713327f39400670e0987c48b4e15d8fR11), running the `export-dynamic-plugin` Janus CLI command on the `@backstage/api-docs` plugin produces the following error:

```
Error: Failed to compile '../../node_modules/@asyncapi/parser/esm/from.js':
  Module not found: Error: Can't resolve 'fs' in '/home/dfestal/git/backstage/upstream/backstage/node_modules/@asyncapi/parser/esm'
```

That's related to the fact that the webpack configuration created for the dynamic frontend plugin export doesn't include the  [`resolve.fallback`](https://webpack.js.org/configuration/resolve/#resolvefallback) option which is [included by the backstage CLI](https://github.com/backstage/backstage/blob/v1.23.4/packages/cli/src/lib/bundler/config.ts#L182-L198) when the whole frontend application is built through webpack.


PatAKnight (approved):
/LGTM

nickboldt:
Merged based on Patrick's comments in slack and the LGTM above.

PR #1567, opened 2024-04-26
feat(scaffolder): create custom action for scaffolder templates

debsmita1:
**Resolves:**

https://issues.redhat.com/browse/RHIDP-2192
[RHIDP-2221](https://issues.redhat.com/browse/RHIDP-2221)

**Test setup:**

- Follow the README.md to integrate the custom action plugin in the backend.

- Changes in `app-config.yaml`

```
catalog:
  location:
    # Example template
    - type: url
      target: https://github.com/debsmita1/nodejs-ex-1/blob/master/local-template.yaml
      rules:
        - allow: [Template]
```

- Changes in `App.tsx`

```
import {
  CatalogEntityPage,
  CatalogIndexPage,
  CatalogTable,
  CatalogTableRow,
  CatalogTableColumnsFunc,
} from '@backstage/plugin-catalog';


export const AppBase = () => {
  const { AppProvider, AppRouter, dynamicRoutes } =
    useContext(DynamicRootContext);
  const myCustomColumnsFunc: CatalogTableColumnsFunc = entityListContext => {
    if (entityListContext.filters.kind?.value === 'component') {
      return [
        ...CatalogTable.defaultColumnsFunc(entityListContext),
        {
          title: 'Created At',
          ender: (data: CatalogTableRow) =>
            data.entity.metadata.annotations?.['backstage.io/createdAt'],
        },
      ];
    }

    return CatalogTable.defaultColumnsFunc(entityListContext);
  };
  .....
  
  
  <Route
       path="/catalog"
       element={<CatalogIndexPage columns={myCustomColumnsFunc} />}
   />
```

PR #1591, opened 2024-04-30
feat: add custom processor module to add ScaffoldedFrom relation

Zaperex:
### Description
Adds a custom catalog processor that creates relations between catalog entities and the scaffolder templates that created them based on the `spec.scaffoldedFrom` field in the `catalog-info.yaml` of the catalog entities.

image

### What issue(s) does this fix?
Fixes: [RHIDP-2156](https://issues.redhat.com/browse/RHIDP-2156)

1. Install the module using the instructions in the README.md
2. Ingest an entity with a `spec.scaffoldedFrom` field that matches the `entityRef` of a template entity in the catalog (ex: `template:default/test-template`). You can potentially use the following example template: https://github.com/janus-idp/backstage-plugins/blob/main/plugins/scaffolder-annotator-action/examples/templates/01-scaffolder-template.yaml
3. Observe the relation graph of the entity in the catalog (or view the raw YAML/JSON of the entity and find the `scaffoldedFrom` relation in the `relations` array)


Zaperex:
Used in conjunction with https://github.com/janus-idp/backstage-plugins/pull/1567

plugins/catalog-backend-module-scaffolder-relation-processor/src/types.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Extension of the `spec` field of the entity model
+ * Used to form relations between entities and the scaffolder templates that generated them
+ *
+ * @public
+ */
+export type ScaffoldedBySpec = {
```
PatAKnight, line 7: Change to `ScaffoldedFromSpec` for consistency.

Zaperex:
Hmm that's weird, the template should have the reverse relation in the dependency graph.

Zaperex:
It should work if you manually add it though (as long as the referenced template exists)? That's how I initially tested it before having the action.

PR #1592, opened 2024-04-30
fix(cli): remove the requirement for `app-config` when exporting fronted plugins to dynamic.

davidfestal:
This PR removes the requirement of having a nearly-empty `app-config.yaml` file when exporting fronted plugins to dynamic.

In fact this requirement has been borrowed from the webpack configuration for the frontend application. But it makes no sense for dynamic plugins.


PatAKnight (approved):
/LGTM

PR #1593, opened 2024-04-30
fix(cli): allow exporting dynamic plugins from `frontend-plugin-module` packages.

davidfestal:
The `export-dynamic-plugin` command of the `@janus-idp/cli` package, when applied to frontend plugins, doesn't support packages with the `frontend-plugin-module`.

This should be allowed so that `APIFactories` can be added, for example to support additional analytics providers.

This PR drops this unnecessary limitation.


PatAKnight (approved):
/lgtm

PR #1594, opened 2024-05-01
chore(orchestrator): add yarn scripts to copy to showcase

batzionb:
added yarn tasks to orchestrator plugin to assist with developing against showcase application
the frontend watch enables making a change and seeing at after refresh of the browser
the backend watch requires rerunning "yarn start" in the showcase app after each change

### DIFF
--- a/plugins/orchestrator-backend/package.json
+++ b/plugins/orchestrator-backend/package.json
@@ -54,7 +54,9 @@
     "prepack": "backstage-cli package prepack",
     "postpack": "backstage-cli package postpack",
     "postversion": "yarn run export-dynamic",
-    "export-dynamic": "janus-cli package export-dynamic-plugin"
+    "export-dynamic": "janus-cli package export-dynamic-plugin",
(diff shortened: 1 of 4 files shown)


jkilzi (approved):

caponetto:
@batzionb 
This is very specific to the showcase. Does it need to be here?
Suppose we move our plugins upstream, this script won't make much sense.

jkilzi:
> @batzionb This is very specific to the showcase. Does it need to be here? Suppose we move our plugins upstream, this script won't make much sense.

Perhaps a best approach would be to have a dedicated repo with this kind of tools/scripts?
@caponetto, in the context of moving the plugin upstream, where is that place exactly?

batzionb:
> @batzionb This is very specific to the showcase. Does it need to be here? Suppose we move our plugins upstream, this script won't make much sense.

once we move it upstream it'll still be relevant, IIUC the showcase is the repo that is meant to stay
I can move this to a different repository

batzionb:
will put this in a different repository

PR #1598, opened 2024-05-02
fix(cli): allow retrieving scalprum config from an external file in `export-dynamic-plugin`.

davidfestal:
This PR adds the avalability, in the `export-dynamic-plugin` command of the `janus-idp/cli` package, to take the frontend plugin scalprum-specific configuration from an external JSON file instead of expecting the config to be specified in a `scalprum` field of the source plugin `package.json`.

#### Background

It should be possible to run the `export-dynamic-plugin` CLI command directly on normal & typical backstage static frontend plugin without having to change its source.

With the current behavior, before this PR, when additional scalprum configuration is needed, there is still the requirement to update the source `package.json` in order to add the `scalprum` field with additional configuration.

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/frontend.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/frontend.ts
@@ -36,12 +36,22 @@ export async function frontend(
   const {
     name,
     version,
-    scalprum: scalprumExternal,
+    scalprum: scalprumInline,
     files,
   } = await fs.readJson(paths.resolveTarget('package.json'));
 
-  let scalprum = scalprumExternal;
-  if (scalprum === undefined) {
+  let scalprum: any = undefined;
+
+  if (opts.scalprumConfig) {
+    const scalprumConfigFile = paths.resolveTarget(opts.scalprumConfig);
+    Task.log(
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/export-dynamic-plugin/frontend.ts
```diff
@@ -36,12 +36,22 @@ export async function frontend(
   const {
     name,
     version,
-    scalprum: scalprumExternal,
+    scalprum: sclaprumInline,
```
gashcrumb, line 39: type here

davidfestal: Fixed

packages/cli/src/commands/export-dynamic-plugin/frontend.ts
```diff
@@ -79,6 +89,12 @@ export async function frontend(
     }
 
     await fs.mkdirs(target);
+    await fs.writeFile(
+      path.join(target, '.gitignore'),
+      `
+*
+`,
```
gashcrumb, line 96: Is this formatting intentional?

davidfestal: Yes.

gashcrumb (approved):

PR #1599, opened 2024-05-02
feat(orchestrator): add permissions to orchestrator plugin

rgolangh:
Motivation
Workflows create assets and retrievs data from external parties. Those created assets are non-trivial by nature, cost money and time and the data retrieved should be kept for federated. For this reason we are introducing permission
on the  various route entries of the orchestrator.

Modification
All the route entries of the orchestrator backend plugin will run an authorization check 
All frontend calls to the backend will pass identity when its available

Result
Orcestrator interactions requires authorization to execute workflow and view their outcomes.

Signed-off-by: Roy Golan <rgolan@redhat.com>


jkilzi:
It is worth mentioning specifically for Podman Desktop users that they must still ensure the `/tmp/orchestrator` path is shared with the VM as a prerequisite.

batzionb (dismissed):

jkilzi (dismissed):

rgolangh:
/hold
I pushed more changed to the branch. fixing

jkilzi (commented):
@rgolangh please be aware that you are commiting the changes inside the packages/{app,backend} for everybody. Is that an intentional change?

rgolangh:
@jkilzi please see the new #1612 

I'll use this pr for the permissions changes

app-config.yaml
```diff
@@ -136,3 +166,14 @@ dynamicPlugins:
         - bindTarget: 'remotePlugins.Scaffolder.scaffolderPlugin.externalRoutes'
           bindMap:
             registerComponent: 'catalogImportPlugin.routes.importPage'
+
+orchestrator:
```
caponetto, line 170: We shouldn't add orchestrator-specific changes outside our packages.

caponetto (changes requested):
We shouldn't add orchestrator-specific changes outside our packages.

rgolangh (commented):
@caponetto @jkilzi 
In my next commit I'll exclude all non orch plugin changes

caponetto:
@rgolangh could you please address those [19 new issues](https://sonarcloud.io/project/issues?id=janus-idp_backstage-plugins&pullRequest=1599&resolved=false&sinceLeakPeriod=true) that SonarCloud has raised?

plugins/orchestrator/docs/Permissions.md
```diff
@@ -0,0 +1,33 @@
+The orchestrator plugin protects its backend endpoints with the builtin permission mechanism and combines it with 
+the RBAC plugin. The result is control over what users can see or execute.
+
+## Orchestrator Permissions
+
+| Name                                 | Resource Type           | Policy | Description                                              | Requirements |
+| ----------------------------         | -------------- | ------ | ----------------------------------------------------------------  | ------------ |
+| orchestrator.workflowInstances.read  | named resource | read   | Allows the user to read orchestrator workflows overview           |              |
+| orchestrator.workflowInstance.read   | named resource | read   | Allows the user to read the details of a single workflow instance |              |
+| orchestrator.workflowInstance.aboart | named resource | use    | Allows the user to aboart a workflow instance                     |              |
```
batzionb, line 10:
```suggestion
| orchestrator.workflowInstance.aboart | named resource | use    | Allows the user to abort a workflow instance                     |              |
```

plugins/orchestrator/src/api/OrchestratorClient.ts
```diff
@@ -155,14 +121,49 @@ export class OrchestratorClient implements OrchestratorApi {
...
+        return response
+    }
+    return await response.json();
+  }
+  // temp duplicate solution for text consumers
+  private async fetcherText(url: string, init?: RequestInit) :Promise<any> {
+    const { token: idToken } = await this.identityApi.getCredentials();
+    const baseInit = {
+      headers: {
+        'Content-Type': 'application/json',
```
batzionb, line 157:
```suggestion
        'Content-Type': 'text/plain',
```

batzionb:
@rgolangh 
PTAL at typescript errors:
https://github.com/janus-idp/backstage-plugins/actions/runs/9015191029/job/24769322509?pr=1599#step:3:653
In addition there are prettier errors, you can resolve these by running:
yarn prettier plugins/orchestrator-backend  --write
and then submit the changes

batzionb:
I think it would be useful to point to the permissions doc from the [README](https://github.com/janus-idp/backstage-plugins/tree/main/plugins/orchestrator#readme)

batzionb:
@gciavarrini  PTAL 
We will need to update v2 to have the authorization integration

plugins/orchestrator/src/api/OrchestratorClient.ts
jkilzi: I have 3 remarks about this code:
1. The `Content-Type` should be overridable by the caller.
2. It cannot be `text/plain` as @batzionb suggests because for instance in [line 53](https://github.com/janus-idp/backstage-plugins/pull/1599/files#diff-e02456b10b75ecb00d8c7c09d87ab9480374f41e4734178ba89625aed2da1d20R53) you are sending a JSON payload.

I suggest to destructure the `init.headers` at the end of the object literal, this will allow you to define some defaults and allow the users to override them when necessary.

plugins/orchestrator-backend/static/generated/envelope/combined.bundle.js
jkilzi, line 1: All files under static/generated/* should not be changed

jkilzi (changes requested):
Please note that all files in orchestrator-backend/static/generated/* should not be changed manually. Consider reverting these changes 🙏

plugins/orchestrator/src/api/OrchestratorClient.ts
rgolangh: I intentionally made auth headers and content type unchanable by the caller. At this stage I don't see
any other content type we need to support, and all calls should include identity information, if present. 
Unless I'm missing a use-case, why should we allow those behaviours if are not needed or wanted?

rgolangh:
ack

plugins/orchestrator/src/api/OrchestratorClient.ts
jkilzi: If ["a GET request should contain no body"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET#:~:text=Note%3A%20Sending%20body/payload%20in%20a%20GET%20request%20may%20cause%20some%20existing%20implementations%20to%20reject%20the%20request%20%E2%80%94%20while%20not%20prohibited%20by%20the%20specification%2C%20the%20semantics%20are%20undefined.%20It%20is%20better%20to%20just%20avoid%20sending%20payloads%20in%20GET%20requests.), then specifing a content-type in this kind of operation is redundant.

jkilzi (changes requested):
Just fix 1 last nit and we are good to go :)

jkilzi (approved):

plugins/orchestrator-backend/package.json
```diff
@@ -60,6 +60,7 @@
     "@backstage/backend-app-api": "^0.7.2",
     "@backstage/backend-common": "^0.21.7",
     "@backstage/backend-dynamic-feature-service": "^0.2.9",
+    "@backstage/errors": "^1.2.4",
```
caponetto, line 63: Please run `yarn export-dynamic` on `plugins/orchestrator-backend` folder so this dependency is added to `dist-dynamic/package.json` file.

plugins/orchestrator/docs/Permissions.md
```diff
@@ -0,0 +1,33 @@
+The orchestrator plugin protects its backend endpoints with the builtin permission mechanism and combines it with
```
caponetto, line 1:
```suggestion
The Orchestrator plugin protects its backend endpoints with the builtin permission mechanism and combines it with
```

plugins/orchestrator/docs/rbac-policy.csv
```diff
@@ -0,0 +1,11 @@
+p, role:default/workflowViewer, orchestrator.workflowInstances.read, read, allow
```
caponetto, line 1: Do we need this file? We already have a sample in the `Permissions.md` file.

plugins/orchestrator-backend/package.json
rgolangh: ahh easy to miss out. thanks for the heads up.
any reasonable way to automatically enforce/warn against this?

plugins/orchestrator/docs/rbac-policy.csv
rgolangh: easy to get started for developers.

batzionb (approved):

caponetto (approved):

BillKlineVT:
I updated to the latest modules (`@janus-idp/backstage-plugin-orchestrator-backend` `1.9.0` and `@janus-idp/backstage-plugin-orchestrator-common` `1.8.0` that have this change and was hopeful that it would fix my issue I wrote [here](https://github.com/janus-idp/backstage-plugins/issues/1700), but I am still getting the same 401 error on the frontend when it calls the backend.

It looks like as a part of this PR you did the backend auth migration mentioned [here](https://backstage.io/docs/tutorials/auth-service-migration/)

Can you confirm you have this working with a recent version of Backstage (1.24.0 or later)?  I'm on the latest 1.27.2.

PR #1610, opened 2024-05-06
Audit Log: Track RBAC changes

AndrienkoAleksandr:

Implement audit logging for RBAC backend

### Referenced issue:
https://issues.redhat.com/browse/RHIDP-2004


plugins/rbac-backend/src/audit-log/audit-logger.ts
```diff
@@ -0,0 +1,291 @@
...
+    request: PolicyQuery,
+    result?: AuthorizeResult,
+    condition?: ConditionalPolicyDecision,
+  ) {
+    const resourceType = (request.permission as ResourcePermission)
+      .resourceType;
+
+    const logMsg: LogMsgWithEvaluationInfo = {
+      level,
+      message,
```
Zaperex, line 237:
```suggestion
```

plugins/rbac-backend/src/audit-log/audit-logger.ts
```diff
@@ -0,0 +1,291 @@
...
+    const logMsg: LogMsg = {
+      level: 'error',
+      message: msg,
+      isAuditLog: true,
+      entityRef,
+      source: 'rest',
+      modifiedBy,
+      time,
+    };
+    this.logger.error(logMsg);
```
Zaperex, line 221:
```suggestion
    const logMsg: LogMsg = {
      isAuditLog: true,
      entityRef,
      source: 'rest',
      modifiedBy,
      time,
    };
    this.logger.error(msg, logMsg);
```

plugins/rbac-backend/src/audit-log/audit-logger.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+    if (resourceType) {
+      logMsg.resourceType = resourceType;
+    }
+    if (condition) {
+      logMsg.condition = JSON.stringify(condition);
+    }
+    logMsg.time = new Date().toUTCString();
+
+    this.logger.log(logMsg);
```
Zaperex, line 257:
```suggestion
    if (level === 'error'){
      this.logger.error(message, logMsg);
    } else {
      this.logger.info(message, logMsg);
    }
```

plugins/rbac-backend/src/audit-log/audit-logger.ts
```diff
@@ -0,0 +1,291 @@
...
+        roleEntityRefs,
+      )}'. Cause: ${e.message}. Stack trace: ${e.stack}`,
+      isAuditLog: true,
+      entityRef: roleEntityRefs,
+      source,
+      modifiedBy,
+      time: new Date().toUTCString(),
+    };
+    this.logger.error(msg);
+  }
```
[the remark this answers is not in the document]
Zaperex: Just sending the `msg` via `logger.error` would result in the entire `LogMsg` object being stuffed into the `message` field of the log. If you want to have the same behavior as `logger.log`, you'll need to provide all the non-message fields of the log in the [`meta` field](https://github.com/backstage/backstage/blob/99305a09da8b0ab628421c66d9d3742cb125362d/packages/backend-plugin-api/src/services/definitions/LoggerService.ts#L25) of the function.

Zaperex (changes requested):
Upstream backstage is moving towards using their [`LoggerService`](https://github.com/backstage/backstage/blob/4ea48f4e7a9579c3570fe1c560b93a82b5f2f5ae/packages/backend-plugin-api/src/services/definitions/LoggerService.ts#L24) (also refer to the [implementation](https://github.com/backstage/backstage/blob/4ea48f4e7a9579c3570fe1c560b93a82b5f2f5ae/packages/backend-app-api/src/logging/WinstonLogger.ts#L47)) instead of the winston Logger. The `LoggerService` does not contain this `log` function so we should be using the `info` and `error` functions instead to make migration to LoggerService in the future easier.

To get the same behavior as using `log`, you'll need to put the non message fields in the [metadata argument](https://github.com/backstage/backstage/blob/4ea48f4e7a9579c3570fe1c560b93a82b5f2f5ae/packages/backend-app-api/src/logging/WinstonLogger.ts#L172-L182) of the `info` or `error` function. Otherwise, the object containing your audit log message will be logged as an object in the message field. This would result in `[object Object]` being printed if the logger format isn't JSON.

plugins/rbac-backend/src/audit-log/audit-logger.ts
```diff
@@ -0,0 +1,289 @@
...
+    const logMsg: LogMsg = {
+      isAuditLog: true,
+      entityRef: metadata.roleEntityRef,
+      source: metadata.source,
+      modifiedBy: metadata.modifiedBy,
+      time: new Date().toUTCString(),
+    };
+
+    this.logger.info(
+      `${this.fmtToPastTime(operation)} '${metadata.roleEntityRef}'`
```
Zaperex, line 72:
```suggestion
      `${this.fmtToPastTime(operation)} '${metadata.roleEntityRef}'`,
```

plugins/rbac-backend/src/audit-log/audit-logger.ts
```diff
@@ -0,0 +1,289 @@
...
+      this.toLogMsgWithRoleInfo(logMsg, metadata)
+    );
```
Zaperex, line 74:
```suggestion
    this.logger.info(
      `${this.fmtToPastTime(operation)} '${metadata.roleEntityRef}'`,
      this.toLogMsgWithRoleInfo(logMsg, metadata)
    );
```

AndrienkoAleksandr:
close on favour: https://github.com/janus-idp/backstage-plugins/pull/1694

PR #1613, opened 2024-05-06
feat(rbac): add support for rejectUnauthorized in PG SSL options

boris154:
Backstage's database configurations support various options that are being passed to the DB connection, example:
```yaml
ssl:
  require: true
  rejectUnauthorized: false
```

This PR adds support for rejectUnauthorized.

### DIFF
--- a/plugins/rbac-backend/src/database/casbin-adapter-factory.ts
+++ b/plugins/rbac-backend/src/database/casbin-adapter-factory.ts
@@ -10,7 +10,7 @@ import { TlsOptions } from 'tls';
 const DEFAULT_SQLITE3_STORAGE_FILE_NAME = 'rbac.sqlite';
 
 type DbSSLOptions = {
-  ca: string;
+  ca?: string;
   rejectUnauthorized?: boolean;
 };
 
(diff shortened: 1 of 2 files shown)


plugins/rbac-backend/src/database/casbin-adapter-factory.ts
PatAKnight, line 1: It looks like there is some overlap between the knex client and typeORMAdapter. So we can update to instead do something like the below. Which would allow us to no longer have to use custom code to resolve ssl.

```
adapter = await TypeORMAdapter.newAdapter({
  type: 'postgres',
  username: databaseConfig?.getString('connection.user'),
  password: databaseConfig?.getString('connection.password'),
  ...knexClient.client.config.connection
});
```

What do you think?

PatAKnight (commented):
Hello, thank you for this PR. Sorry that we are just now getting around to reviewing it. Had one comment that I wanted to get your opinion on.

plugins/rbac-backend/src/database/casbin-adapter-factory.ts
AndrienkoAleksandr: @PatAKnight, sorry, I have some concerns about mixing these two different types. The SSL type used in Backstage is:

`boolean | ConnectionOptions | undefined`

while the ORM adapter expects:

`boolean | TlsOptions | undefined`

These types aren’t interchangeable, so it might lead to issues if we try to mix them.

AndrienkoAleksandr:
@PatAKnight, I cherry-picked and rebased code from this pr and provided some improvements: https://github.com/janus-idp/backstage-plugins/pull/2122 . Can you take a look, please.

AndrienkoAleksandr:
@boris154 we merged fix, but in the another pr https://github.com/janus-idp/backstage-plugins/pull/2122. Thank you very much for your contribution!

PR #1618, opened 2024-05-07
fix(argocd): fix argocd configurations visibility

karthikjeeyar:
Change visibility of the argocd appLocatorMethods configuration and explicitly mask the visibility of username and password fields.


rohitkrai03 (approved):
/lgtm

PR #1628, opened 2024-05-09
feat(rbac): support for updating/deleting conditional permissions

divyanshiGupta:
Story: https://issues.redhat.com/browse/RHIDP-1878

This PR adds support for updating/deleting existing conditions.

https://github.com/janus-idp/backstage-plugins/assets/20724543/ffbeb886-9656-4b23-a38c-0f23985c291b

### DIFF
--- a/plugins/rbac/src/components/CreateRole/RoleForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/RoleForm.tsx
@@ -20,8 +20,11 @@ import { rbacApiRef } from '../../api/RBACBackendClient';
 import { MemberEntity, PermissionsData, RoleError } from '../../types';
 import {
   getConditionalPermissionPoliciesData,
+  getNewConditionalPolicies,
   getPermissionPoliciesData,
+  getRemovedConditionalPoliciesIds,
   getRoleData,
+  getUpdatedConditionalPolicies,
   validationSchema,
 } from '../../utils/create-role-utils';
 import {
@@ -82,6 +85,15 @@ export const RoleForm = ({
     try {
       const newData = getRoleData(values);
       const newPermissionsData = getPermissionPoliciesData(values);
+      const newConditions = getNewConditionalPolicies(values);
+      const deleteConditions = getRemovedConditionalPoliciesIds(
+        values,
+        initialValues,
+      );
+      const updateConditions = getUpdatedConditionalPolicies(
+        values,
+        initialValues,
+      );
 
       const oldData = getRoleData(initialValues);
       const res = await rbacApi.updateRole(oldData, newData);
@@ -126,6 +138,56 @@ export const RoleForm = ({
             );
           }
         }
+
+        if (deleteConditions.length > 0) {
+          const promises = deleteConditions.map(cid =>
+            rbacApi.deleteConditionalPolicies(cid),
+          );
+
+          const cppRes: (Response | RoleError)[] = await Promise.all(promises);
+          const cpErr = cppRes
+            .map(r => (r as unknown as RoleError).error?.message)
+            .filter(m => m);
+
+          if (cpErr.length > 0) {
+            throw new Error(
+              `Unable to remove conditions from the role. ${cpErr.join('\n')}`,
+            );
+          }
+        }
+
+        if (newConditions.length > 0) {
+          const promises = newConditions.map(cpp =>
+            rbacApi.createConditionalPermission(cpp),
+          );
+
+          const cppRes: (Response | RoleError)[] = await Promise.all(promises);
+          const cpErr = cppRes
+            .map(r => (r as unknown as RoleError).error?.message)
+            .filter(m => m);
+
+          if (cpErr.length > 0) {
+            throw new Error(
+              `Unable to add conditions to the role. ${cpErr.join('\n')}`,
+            );
+          }
+        }
+
+        if (updateConditions.length > 0) {
+          const promises = updateConditions.map(({ id, updateCondition }) =>
+            rbacApi.updateConditionalPolicies(id, updateCondition),
+          );
+
+          const cppRes: (Response | RoleError)[] = await Promise.all(promises);
+          const cpErr = cppRes
+            .map(r => (r as unknown as RoleError).error?.message)
+            .filter(m => m);
+
+          if (cpErr.length > 0) {
+            throw new Error(`Unable to update conditions. ${cpErr.join('\n')}`);
+          }
+        }
+
         navigateTo(`${name} updated`);
       }
     } catch (e) {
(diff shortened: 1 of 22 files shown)


divyanshiGupta:
cc: @ShiranHi

ShiranHi:
@divyanshiGupta looks good to me!

I've noticed that the edit and delete icons are different colors. Could we please take a look at that?
icons

Also, I'm concerned about the shade of red used here. Could we ensure that both the icon and the button use the [PF red color](https://www.patternfly.org/design-foundations/colors/#status-and-state-colors)?
deletion

divyanshiGupta:
> @divyanshiGupta looks good to me!
> 
> I've noticed that the edit and delete icons are different colors. Could we please take a look at that? icons
> 
> Also, I'm concerned about the shade of red used here. Could we ensure that both the icon and the button use the [PF red color](https://www.patternfly.org/design-foundations/colors/#status-and-state-colors)? deletion

Thanks @ShiranHi 

Looks like these are some theme related issues which might get fixed once we have the new theme in place. For now can you log tickets for these issues in jira?

plugins/rbac/src/utils/create-role-utils.ts
```diff
@@ -202,6 +202,74 @@ export const getConditionalPermissionPoliciesData = (
   );
 };
 
+export const getUpdatedConditionalPolicies = (
```
invincibleJai, line 205: good to add unit tests for these utils

plugins/rbac/src/components/CreateRole/RoleForm.tsx
```diff
@@ -126,6 +138,56 @@ export const RoleForm = ({
             );
           }
         }
+
+        if (deleteConditions.length > 0) {
```
invincibleJai, line 142: do we need tests to cover the below scenarios?

invincibleJai:
if I understand correctly permission policies and conditional policies are treated differently. So if I only define conditional policies then on the overview screen I see the error as below (earlier it was not possible to create a role without policies but with conditional policies it is)

image

this doesn't seem good and if policies are not there IMO we should not show errors for 404, @divyanshiGupta WDYT, can address this as part of this PR.

and on the list/overview screen, we are not showing conditional policies. I don't see this in Figma as well so is this based on the design @ShiranHi , we need to think about how can we improve this further

image

and i observed this issue https://issues.redhat.com/browse/RHIDP-1757 and this needs to be fixed for 1.2

am not sure below why is giving 500

http://localhost:7007/api/permission/authorize

Error:

```
backend:start: 2024-05-14T08:45:26.598Z backstage error Request failed with status 500 Entity reference "catalog-entity" had missing or empty kind (e.g. did not start with "component:" or similar) type=errorHandler stack=Error: Entity reference "catalog-entity" had missing or empty kind (e.g. did not start with "component:" or similar)
backend:start:     at Object.parseEntityRef (/Users/jakumar/Workspaces/backstage-plugins/node_modules/@backstage/catalog-model/src/entity/ref.ts:110:11)
backend:start:     at <anonymous> (/Users/jakumar/Workspaces/backstage-plugins/node_modules/@backstage/plugin-catalog-backend/src/service/CatalogBuilder.ts:550:49)
backend:start:     at Array.map (<anonymous>)
backend:start:     at getResources (/Users/jakumar/Workspaces/backstage-plugins/node_modules/@backstage/plugin-catalog-backend/src/service/CatalogBuilder.ts:549:33)
backend:start:     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
backend:start:     at <anonymous> (/Users/jakumar/Workspaces/backstage-plugins/node_modules/@backstage/plugin-permission-node/src/integration/createPermissionIntegrationRouter.ts:460:27)
```

@AndrienkoAleksandr / @PatAKnight is this expected?

plugins/rbac/src/components/CreateRole/RoleForm.tsx
divyanshiGupta: There are unit tests for the RoleForm component but not form save functionalities such as (create/update) but these are covered by e2e. Maybe we can have a task for adding unit tests for these if needed.

divyanshiGupta: I see there is no e2e for testing role deletion. We can have one more task for imrproving e2e as well.

divyanshiGupta:
@invincibleJai I am now not showing error if API returns 404.
Screenshot 2024-05-14 at 5 15 13 PM

plugins/rbac/src/components/CreateRole/RoleForm.tsx
invincibleJai: in dev mode e2e might not make sense, in showcase we should add  so should be fine for now for this PR

debsmita1:
@divyanshiGupta 
I created a role with conditional policies for managing catalog entities. While editing these policies, I deleted the ones for the 
 catalog and configured access for scaffolder-templates. However, upon saving the changes, an error occurred

Screenshot 2024-05-14 at 6 55 36 PM

divyanshiGupta:
> @divyanshiGupta I created a role with conditional policies for managing catalog entities. While editing these policies, I deleted the ones for the catalog and configured access for scaffolder-templates. However, upon saving the changes, an error occurred
> 
> Screenshot 2024-05-14 at 6 55 36 PM

I dont see this issue.

https://github.com/janus-idp/backstage-plugins/assets/20724543/73a438c0-dffb-4d82-9870-8c9cbb5a1240

invincibleJai:
one observation with tab switch on sidebar - it's not persisting the saved values



https://github.com/janus-idp/backstage-plugins/assets/5129024/6b1ab416-854d-4e4d-87eb-fb7193fdeb03



should this be handled?

divyanshiGupta:
> one observation with tab switch on sidebar - it's not persisting the saved values
> 
> 
> 
> 
> 
> 
> 
> https://github.com/janus-idp/backstage-plugins/assets/5129024/6b1ab416-854d-4e4d-87eb-fb7193fdeb03
> 
> 
> 
> 
> 
> 
> 
> should this be handled?

@invincibleJai the value is not persisted in the sidebar as I am using the same model as backend wherein there can be only one criteria at a time. Once the user changes the criteria a new object is added. But if you haven't saved the new criteria you can go back to the previous by opening the sidebar again as the values are only saved in the main form on clicking save.

ShiranHi:
> and on the list/overview screen, we are not showing conditional policies. I don't see this in Figma as well so is this based on the design @ShiranHi , we need to think about how can we improve this further

Thank you @invincibleJai for pointing this. I think we can add a "Conditional" column write "x rules" if specified

PR #1631, opened 2024-05-10
chore(owners): remove OWNERS files from orchestrator plugins

kim-tsao:
- Confirmed with the @janus-idp/orchestrator team that they would prefer to use CODEOWNERS over OWNERS.  Deleting OWNERS to avoid dup assignments
- Update CODEOWNERS to add the missing `/plugins/orchestrator-swf-editor-envelope` plugin


caponetto (approved):
Thank you!

PR #1632, opened 2024-05-10
fix(rbac)!: remove token manager for auth service

PatAKnight:
## Description

While upgrading to the new backend system, I missed one other place that required the new `auth` backend service. This PR fixes that mistake as well as removes the token manager since it is no longer in use.

## Special notes for reviewer

I tested this with both `admin` and `superUser` with and without the database.

### DIFF
--- a/plugins/rbac-backend/README.md
+++ b/plugins/rbac-backend/README.md
@@ -62,7 +62,6 @@ export default async function createPlugin(
       discovery: env.discovery,
       identity: env.identity,
       permissions: env.permissions,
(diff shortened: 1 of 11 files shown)


PatAKnight:
@divyanshiGupta This should fix the issue that you were having with testing your PR #1628

divyanshiGupta (approved):
@PatAKnight thanks for the PR. It works as expected.

/approve

divyanshiGupta:
Removing lgtm in case @AndrienkoAleksandr also wants to review it once.

PR #1636, opened 2024-05-10
chore(owners): remove OWNERS files from devex and rhtap plugins

kim-tsao:
* got confirmation from the devex and rhtap teams that these files can be removed from their plugins


yashoswalyo (approved):

yashoswalyo:
LGTM

PR #1641, opened 2024-05-13
Rhidp 1502

schultzp2020:


packages/cli/src/commands/metadata/command.ts
```diff
@@ -208,10 +183,39 @@ export function updatePackageMetadata(
...
+    newKeywords.unshift(optsLifecycle ?? oldLifecycle ?? '');
+  }
+
+  // if there is a support keyword, push to the beginning of the array
+  if (oldSupport || optsSupport) {
+    newKeywords.unshift(optsSupport ?? oldSupport ?? '');
+  }
 
-  /* all done! */
+  // dedup new keywords
```
nickboldt, line 217:
```suggestion
  // dedupe new keywords
```

packages/cli/src/commands/metadata/command.ts
```diff
@@ -208,10 +183,39 @@ export function updatePackageMetadata(
...
   packageJSON.bugs = new URL(opts.bugs);
 
-  // add keywords (replace some), then uniquify
-  addNewKeywords(packageJSON, opts);
+  // initialize empty string array if not already present
+  if (!packageJSON.keywords) {
+    packageJSON.keywords = [];
+  }
+
+  // if already have keywords, replace lifecycle and support with new values (if defined)
```
nickboldt, line 191: does sonarlint not complain about this method being too complex?

schultzp2020: Nope

PR #1642, opened 2024-05-13
fix(kiali): removing unnecessary afterAll hook

ScriptingShrimp:
the `afterAll` hook in test file is not needed. It causing troubles when another file is executed after specific file with the ` await browser.close();` method. From the playwright docs
> The [Browser](https://playwright.dev/docs/api/class-browser) object itself is considered to be disposed and cannot be used anymore.

This is simplest fix for now. If test starts to flake again, lets reiterate and create more robust bootstrap and cleanup methods. Tested, working with multiple test files.


aljesusg (approved):
LGFM

PR #1647, opened 2024-05-14
fix(argocd): make refreshInterval configuration as optional

karthikjeeyar:
**Description:**

Since we have a fallback value for refreshInterval config, marking it as optional field.


rohitkrai03 (approved):
/lgtm

PR #1649, opened 2024-05-14
chore(orchestrator): add openapi client generation support

gciavarrini:
Introduce functionality to generate a client based on the OpenAPI specification available in the `orchestrator-common` module. 

The main features:
* _Client Generation_:
Utilizes  `@hey-api/openapi-ts` to create client from the OpenAPI YAML file available in `orchestrator-common`.
* _Change Detection_: Implements a mechanism to detect changes in the OpenAPI specification or auto-generated files during the `build`, prompting regeneration if necessary.

Although the generated client is not currently in use, subsequent PRs will focus on integrating this client and phasing out the current v1 endpoint in the orchestrator module.


gciavarrini:
@batzionb @caponetto PTAL

schultzp2020:
@gciavarrini Just a heads up, some of the other plugins in janus-idp/backstage-plugins already support OpenAPI client generation using `@hey-api/openapi-ts`. For example, check out [`servicenow-actions`](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/servicenow-actions/package.json#L35). It might be helpful to see how it's implemented there. My main concern is that you're managing dependencies [outside of the `package.json` file](https://github.com/janus-idp/backstage-plugins/pull/1649/files#diff-a6f4686eab29be3fa065a633624a95e42ecef380deb3f844aef0ab8dd359ceaaR40).

gciavarrini:
> @gciavarrini Just a heads up, some of the other plugins in janus-idp/backstage-plugins already support OpenAPI client generation using `@hey-api/openapi-ts`. For example, check out [`servicenow-actions`](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/servicenow-actions/package.json#L35). It might be helpful to see how it's implemented there. My main concern is that you're managing dependencies [outside of the `package.json` file](https://github.com/janus-idp/backstage-plugins/pull/1649/files#diff-a6f4686eab29be3fa065a633624a95e42ecef380deb3f844aef0ab8dd359ceaaR40).

@schultzp2020 thank you for your review!
I updated the PR following your suggestions. I wasn't aware of `@hey-api/openapi-ts` :)

plugins/orchestrator/openapitools.json
```diff
@@ -0,0 +1,7 @@
+{
```
schultzp2020, line 1: Is this file still required?

plugins/orchestrator/package.json
```diff
@@ -28,14 +28,16 @@
   ],
   "scripts": {
     "start": "backstage-cli package start",
-    "build": "backstage-cli package build",
+    "build": "yarn openapi:check && backstage-cli package build",
```
schultzp2020, line 31: I would suggest creating a dependency in turbo.json.

e.g. https://github.com/janus-idp/backstage-plugins/blob/main/plugins/servicenow-actions/turbo.json

plugins/orchestrator/package.json
```diff
@@ -28,14 +28,16 @@
...
     "export-dynamic": "janus-cli package export-dynamic-plugin",
     "tsc": "tsc",
     "lint": "backstage-cli package lint",
     "test": "backstage-cli package test --passWithNoTests --coverage",
     "clean": "backstage-cli package clean",
     "prepack": "backstage-cli package prepack",
-    "postpack": "backstage-cli package postpack"
+    "postpack": "backstage-cli package postpack",
+    "openapi:generate:client": "./scripts/openapi.sh generate",
```
schultzp2020, line 39:
```suggestion
    "generate": "./scripts/openapi.sh generate",
```

To maintain consistency, I believe it's beneficial to use the same script name as other plugins.

plugins/orchestrator/src/generated/core/ApiError.ts
```diff
@@ -0,0 +1,21 @@
+import type { ApiRequestOptions } from './ApiRequestOptions';
```
schultzp2020, line 1: Do we need to include the automatically generated files in the commits? Other plugins typically exclude them. Since the OpenAPI client generation process is idempotent, I believe it's unnecessary.

plugins/orchestrator/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "@backstage/cli/config/tsconfig.json",
   "include": ["src", "dev"],
-  "exclude": ["node_modules"],
+  "exclude": ["node_modules", "src/generated"],
```
schultzp2020, line 4:
```suggestion
  "exclude": ["node_modules"],
```

I don't think you need to exclude `src/generated`. e.g. https://github.com/janus-idp/backstage-plugins/blob/main/plugins/servicenow-actions/tsconfig.json#L4

plugins/orchestrator/src/generated/core/ApiError.ts
gciavarrini: I'm not a TypeScript expert, so I may be missing something. However, we need to use the generated code in [OrchestratorClient](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator/src/api/OrchestratorClient.ts). If I don't include the generated code in the commits, how can we ensure that it's available for use?

batzionb: +1

plugins/orchestrator/package.json
batzionb: @gciavarrini 
My instinct when cloning the repo and trying to use it is to go to the backend to generate the client, since this isn't an operation related to the UI
IIUC the purpose is to enable the user to consume the backend easily.
Do you think it can move to the backend?

gciavarrini: I currently have two different tasks: `openapi:generate:client` and `openapi:check`. 
While I agree that consistency is beneficial, using names like `generate` and `check` might be too generic and could lead to confusion. Perhaps a good compromise would be to rename them to `openapi:generate` and `openapi:check` to simplify the names while still maintaining clarity. What do you think?

batzionb:
@gciavarrini 
I think this capability should move to orchestrator-backend
My instince when cloning the repo was to go there to run it
The goal of this capability is to enable running the backend, regardless of the UI even existing

plugins/orchestrator/openapitools.json
gciavarrini: Ops, leftover from the previous tool :)

plugins/orchestrator/scripts/openapi.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+pwd
+set -e
+
+OPENAPI_SPEC_FILE="../orchestrator-common/src/openapi/openapi.yaml"
+CLIENT_FOLDER="./src/generated"
```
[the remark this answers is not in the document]
batzionb: This code definitely shouldn't be added to the UI src code, which is supposed to contain only the code that generates the UI to run on the browser
But I think it should also not be included in the backend src code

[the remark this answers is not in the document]
gciavarrini: @batzionb 
> The src folder of the backend is supposed to contain the code required to run the backend itself, 

why are you talking about "src folder of the backend"?
If you're reffering to `CLIENT_FOLDER` it is inside `orchestrator` package itself, nothing read/write from/to `orchestrator-backend`.

> I suggest that the destination be configured by the user
I don't agree. First, why does the developer should change this dir? Second, the code is generated and imported in `OrchestratorClient.ts`. If the the developer changes the generation folder than it should update all the imports. Third, the `generated` folder is widely used in other plugins

> This code definitely shouldn't be added to the UI src code, which is supposed to contain only the code that generates the UI to run on the browser
But this code generates files to be used only in UI. Need clarification here :)

schultzp2020:
> > @gciavarrini
> > ~I think this capability should move to orchestrator-backend My instince when cloning the repo was to go there to run it The goal of this capability is to enable running the backend, regardless of the UI even existing~
> > @gciavarrini thanks for your explanations, I understand now that the UI will be the main consumer of the client. We might want to consider other possible consumers, which is why it would be a good idea to check out the backstage best practices for openapi client https://backstage.io/docs/openapi/01-getting-started/
> 
> IIUC they suggest to put the client in a separate plugin that can be consumed by the UI and others WDYT?

This tutorial is still in an experimental stage. Therefore, I recommend refraining from using it until it is fully ready.

plugins/orchestrator/src/generated/core/ApiError.ts
schultzp2020: Apologies, I chose to ask a leading question rather than explicitly explaining my approach. The answer is no, we do not need to include the automatically generated files in the commit. Running `build` from the root directory will invoke `tsc`, which in turn calls `openapi:generate`.

plugins/orchestrator/package.json
schultzp2020: Would you be willing to update the other `generate` script references to `openapi:generate` in the repository in a separate PR?

gciavarrini:
> > @gciavarrini
> > ~I think this capability should move to orchestrator-backend My instince when cloning the repo was to go there to run it The goal of this capability is to enable running the backend, regardless of the UI even existing~
> > @gciavarrini thanks for your explanations, I understand now that the UI will be the main consumer of the client. We might want to consider other possible consumers, which is why it would be a good idea to check out the backstage best practices for openapi client https://backstage.io/docs/openapi/01-getting-started/
> 
> IIUC they suggest to put the client in a separate plugin that can be consumed by the UI and others WDYT?

@batzionb I moved the client generation into `orchestrator-common` but I still prefer @schultzp2020 approach (that is using `@hey-api/openapi-ts`). As @schultzp2020 pointed out, the "backstage approach" is experimental and `@hey-api/openapi-ts` offers some advantages: it is easy to use, it is already used in other plugins, and it provides `CancellablePromise`

plugins/orchestrator/src/generated/core/ApiError.ts
gciavarrini: @schultzp2020  Ok, now I see your point, thanks for clarifying. 

To better understand: if I don't commit the generated code, when a developer checks out the project they will initially see import errors. However, building the project will generate the necessary code, fixing these errors. Additionally, when creating the `orchestrator` package, the generated code will be included, so everything will work as expected. Correct?

The downside is that developers will see a lot of uncommitted changes. Should I add the generated folder to .gitignore to prevent this? Or is there a better approach?

gciavarrini:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=janus-idp_backstage-plugins&pullRequest=1649) **Quality Gate failed**
> 
> Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=janus-idp_backstage-plugins&pullRequest=1649&resolved=false&sinceLeakPeriod=true)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=janus-idp_backstage-plugins&pullRequest=1649)

@schultzp2020 do you know how to solve/ignore this? I think you faced the same issue in servicenow plugin

schultzp2020:
> > ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=janus-idp_backstage-plugins&pullRequest=1649) **Quality Gate failed**
> > Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=janus-idp_backstage-plugins&pullRequest=1649&resolved=false&sinceLeakPeriod=true)
> > [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=janus-idp_backstage-plugins&pullRequest=1649)
> 
> @schultzp2020 do you know how to solve/ignore this? I think you faced the same issue in servicenow plugin

It's fixed

plugins/orchestrator/package.json
gciavarrini: Do you mean in other plugins as well (e.g. `servicenow-actions`)?  If so, sure, I will update those in a separate PR

batzionb (approved):

plugins/orchestrator/src/generated/core/ApiError.ts
schultzp2020: Yes to both questions. https://github.com/janus-idp/backstage-plugins/blob/main/plugins/servicenow-actions/.gitignore

gciavarrini: I see you used .gitignore :) I'll apply the changes

batzionb (approved):

plugins/orchestrator/package.json
gciavarrini: Open PR https://github.com/janus-idp/backstage-plugins/pull/1708

PR #1650, opened 2024-05-14
fix(orchestrator): remove the need of react dev dependencies

caponetto:
Accessing the Workflow Editor in the Orchestrator plugin throws an error after the [Backstage upgrade to 1.26](https://github.com/janus-idp/backstage-plugins/pull/1536) and this PR fixes it.


caponetto:
@schultzp2020 PTAL

PR #1681, opened 2024-05-16
fix(orchestrator): fixes many security-related issues

jkilzi:
Addresses:
- https://issues.redhat.com/browse/FLPATH-1316

### DIFF
--- a/plugins/orchestrator-backend/src/service/router.ts
+++ b/plugins/orchestrator-backend/src/service/router.ts
@@ -239,12 +239,15 @@ function setupInternalRoutes(
     const {
       params: { workflowId },
     } = req;
-    await routerApi.v1
-      .getWorkflowSourceById(workflowId)
-      .then(result => res.status(200).send(result))
-      .catch(error => {
-        res.status(500).send(error.message || INTERNAL_SERVER_ERROR_MESSAGE);
-      });
+    try {
+      const result = await routerApi.v1.getWorkflowSourceById(workflowId);
+      res.status(200).contentType('text/plain').send(result);
+    } catch (error) {
+      res
+        .status(500)
+        .contentType('text/plain')
+        .send((error as Error)?.message || INTERNAL_SERVER_ERROR_MESSAGE);
+    }
   });
 
   // v2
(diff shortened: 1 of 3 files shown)


batzionb (approved):

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -239,12 +239,15 @@ function setupInternalRoutes(
...
     } = req;
-    await routerApi.v1
-      .getWorkflowSourceById(workflowId)
-      .then(result => res.status(200).send(result))
-      .catch(error => {
-        res.status(500).send(error.message || INTERNAL_SERVER_ERROR_MESSAGE);
-      });
+    try {
+      const result = await routerApi.v1.getWorkflowSourceById(workflowId);
+      res.status(200).contentType('text/plain').send(result);
```
rgolangh, line 244: don't think we should reply with text/plain?

rgolangh: scratch that, this is for the source

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -269,12 +275,15 @@ function setupInternalRoutes(
...
-      .catch(error => {
-        res.status(500).send(error.message || INTERNAL_SERVER_ERROR_MESSAGE);
-      });
+    try {
+      await routerApi.v1.abortWorkflow(instanceId);
+      res.status(200).send();
+    } catch (error) {
+      res
+        .status(500)
+        .contentType('plain/text')
```
rgolangh, line 284: why plain/text on the abort response?

jkilzi: Because, when an error occurs (Replying with status 500), we are delivering to the client a plain text message, not an object ([Line 285](https://github.com/janus-idp/backstage-plugins/pull/1681/files#diff-afc3e7a4a5383fae292402adaf30b3ac1af05826dee9a024a693cd95d247ff15R285)).

rgolangh: > Because, when an error occurs (Replying with status 500), we are delivering to the client a plain text message, not an object ([Line 285](https://github.com/janus-idp/backstage-plugins/pull/1681/files#diff-afc3e7a4a5383fae292402adaf30b3ac1af05826dee9a024a693cd95d247ff15R285)).

don't need a body at all I guess. but +1

batzionb (approved):

PR #1699, opened 2024-05-21
fix(rbac): conditional access form validation

ciiay:
[RHIDP-2345](https://issues.redhat.com/browse/RHIDP-2345)

Screen recording

https://github.com/janus-idp/backstage-plugins/assets/26255262/2d1416d7-7b04-4a37-b0f3-e7a22b504fcc

### DIFF
--- a/plugins/rbac/src/__fixtures__/mockConditionRules.ts
+++ b/plugins/rbac/src/__fixtures__/mockConditionRules.ts
@@ -91,7 +91,8 @@ export const mockConditionRules = [
           properties: {
             kinds: {
               type: 'array',
-              items: { type: 'string' },
+              items: { type: 'string' }, // TODO: should be enum?
+              minItems: 1,
               description: 'List of kinds to match at least one of',
             },
           },
@@ -110,6 +111,7 @@ export const mockConditionRules = [
             claims: {
               type: 'array',
               items: { type: 'string' },
+              minItems: 1,
               description:
                 'List of claims to match at least one on within ownedBy',
             },
--- a/plugins/rbac/src/components/ConditionalAccess/CustomArrayField.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/CustomArrayField.tsx
@@ -35,8 +35,9 @@ export const CustomArrayField = (props: FieldProps) => {
         label={name}
         value={fieldVal}
         onChange={e => {
-          setFieldVal(e.target.value);
-          onChange(e.target.value.split(',').map(val => val.trim()));
+          const value = e.target.value;
+          setFieldVal(value);
+          onChange(value ? value.split(',').map(val => val.trim()) : []);
         }}
         className={classes.bgPaper}
         required={required}
--- a/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRowFields.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRowFields.tsx
@@ -123,14 +123,14 @@ export const ConditionsFormRowFields = ({
         {schema ? (
           <Form
             schema={paramsSchema}
-            formData={oldCondition?.params ?? null}
+            formData={oldCondition?.params || {}}
             validator={validator}
             uiSchema={uiSchema}
             fields={customFields}
             onChange={data =>
               handleConditionChange({
                 ...oldCondition,
-                params: data.formData ?? {},
+                params: data.formData || {},
               })
             }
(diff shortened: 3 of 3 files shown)


plugins/rbac/src/__fixtures__/mockConditionRules.ts
```diff
@@ -91,7 +91,8 @@ export const mockConditionRules = [
           properties: {
             kinds: {
               type: 'array',
-              items: { type: 'string' },
+              items: { type: 'string' }, // TODO: should be enum?
```
divyanshiGupta, line 94: @ciiay this actually depends on the plugin devs/maintainer as what condition rules they add to their plugins resource-types.

plugins/rbac/src/components/ConditionalAccess/CustomArrayField.tsx
```diff
@@ -35,8 +35,9 @@ export const CustomArrayField = (props: FieldProps) => {
         label={name}
         value={fieldVal}
         onChange={e => {
-          setFieldVal(e.target.value);
-          onChange(e.target.value.split(',').map(val => val.trim()));
+          const value = e.target.value;
+          setFieldVal(value);
+          onChange(value ? value.split(',').map(val => val.trim()) : []);
```
divyanshiGupta, line 40: So if a `minItems` attribute is available in the paramsSchema of the array param then an empty array will return a validation error. But when the array param for example kind is just listed as required this will not return any error as we are returning [] and its an acceptable value. Looks good to me.

ciiay: If it returns an empty array, the item account will be 0 which doesn't meet the `minItem: 1` requirement and it throws an error.

divyanshiGupta: @AndrienkoAleksandr I am able to add below condition on catalog-entity :
```
        {
            "rule": "IS_ENTITY_KIND",
            "resourceType": "catalog-entity",
            "params": {
                "kinds": []
            }
         }
```
On applying above the corresponding role members cannot access any type of catalog-entity kinds. Does this looks correct to you?

plugins/rbac/src/__fixtures__/mockConditionRules.ts
ciiay: Will remove the comment.

plugins/rbac/src/components/ConditionalAccess/CustomArrayField.tsx
AndrienkoAleksandr: We can deny such conditions with help of validator in the frame of the  https://issues.redhat.com/browse/RHIDP-2157 . But maybe it can be feature...

divyanshiGupta: But here https://docs.google.com/document/d/1XMEmSlwVog1cOD0X5ADwg53Vnm8PL-HH_8LZriUM7vs/edit?disco=AAABNnL7BR8 kinds is just required which mean kinds cannot be undefined but it can be [] also it doesnt have `minItems: 1` to restrict an empty array. So can this might be what plugin conidition rules author expects or its a backend bug?

divyanshiGupta: @AndrienkoAleksandr ^^

divyanshiGupta (approved):
@ciiay PR looks good from UI side as we are checking if `minItems` is present in param schema then array field uses it to validate the input. But if schema just says that an array is required in which case `[]` this is also a valid value and plugin maintainers might have intentionally left out `minItems` restriction. In UI we can only work with what is provided to us in the schema.

> I am able to add below condition on catalog-entity :

        {
            "rule": "IS_ENTITY_KIND",
            "resourceType": "catalog-entity",
            "params": {
                "kinds": []
            }
         }
On applying above the corresponding role members cannot access any type of catalog-entity kinds. Does this look correct to you?

@AndrienkoAleksandr If you think above is a backend bug, please log a bug for it.

/lgtm

plugins/rbac/src/components/ConditionalAccess/CustomArrayField.tsx
AndrienkoAleksandr: > So can this might be what plugin condition rules author expects or its a backend bug?

I believe we are not responsible for such arrays. The plugin author designs and controls this aspect.

PR #1702, opened 2024-05-22
fix(rbac): show configure-access cta for existing simple permission policies in edit form

debsmita1:
**Resolves:**
https://issues.redhat.com/browse/RHIDP-2346

### DIFF
--- a/plugins/rbac/src/utils/rbac-utils.ts
+++ b/plugins/rbac/src/utils/rbac-utils.ts
@@ -108,12 +108,23 @@ export const getMembersFromGroup = (group: GroupEntity): number => {
   return membersList || 0;
 };
 
-export const getPluginId = (
+export const getPluginInfo = (
   permissions: PermissionPolicy[],
-  permission: string | undefined,
-): string =>
-  permissions.find(p => p.policies?.find(pol => pol.permission === permission))
-    ?.pluginId || '-';
+  permissionName?: string,
+): { pluginId: string; isResourced: boolean } =>
+  permissions.reduce(
+    (acc: { pluginId: string; isResourced: boolean }, p: PermissionPolicy) => {
+      const policy = p.policies?.find(pol => pol.permission === permissionName);
+      if (policy) {
+        return {
+          pluginId: p.pluginId || '-',
+          isResourced: policy?.isResourced || false,
+        };
+      }
+      return acc;
+    },
+    { pluginId: '-', isResourced: false },
+  );
 
 const getPolicy = (str: string) => {
   const arr = str.split('.');
@@ -174,12 +185,15 @@ export const getPermissionsData = (
           const policiesSet = new Set<{ policy: string; effect: string }>();
           acc.push({
             permission: policy.permission || '-',
-            plugin: getPluginId(permissionPolicies, policy?.permission) || '-',
+            plugin: getPluginInfo(permissionPolicies, policy?.permission)
+              .pluginId,
             policyString: policyString.add(policyTitleCase || 'Use'),
             policies: policiesSet.add({
               policy: policyTitleCase || 'Use',
               effect: policy.effect,
             }),
+            isResourced: getPluginInfo(permissionPolicies, policy?.permission)
+              .isResourced,
           });
         }
       }
--- a/plugins/rbac/src/types.ts
+++ b/plugins/rbac/src/types.ts
@@ -39,6 +39,7 @@ export type PermissionsDataSet = {
   permission: string;
   policies: Set<RowPolicy>;
   policyString?: Set<string>;
+  isResourced?: boolean;
 };
 
 export type PermissionsData = {
--- a/plugins/rbac/src/utils/rbac-utils.test.ts
+++ b/plugins/rbac/src/utils/rbac-utils.test.ts
@@ -21,7 +21,7 @@ import {
   getMembersFromGroup,
   getPermissions,
   getPermissionsData,
-  getPluginId,
+  getPluginInfo,
   getPoliciesData,
 } from './rbac-utils';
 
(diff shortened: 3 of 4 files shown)


ciiay:
I wasn't able to test this out on `backstage-plugins` repo, as the policies API returns empty array when I ran `yarn start:backstage` in the plugins repo. Here's the screen recording I took from `backstage-showcase` local run with exported this PR branch. Working as expected.

https://github.com/janus-idp/backstage-plugins/assets/26255262/4e4b25fb-3253-4b6c-b6d7-3fe3be9cd80b

plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
```diff
@@ -126,6 +126,7 @@ export const PermissionPoliciesForm = ({
 
   const onAddConditions = (index: number, conditions?: ConditionsData) => {
     setFieldValue(`permissionPoliciesRows[${index}].conditions`, conditions);
+    setFieldValue(`permissionPoliciesRows[${index}].isResourced`, true);
```
divyanshiGupta, line 129: @debsmita1 ideally `isResourced` should be set while we are generating data for edit form similar to what is done for create form because we get `isResourced` as part of `RoleBasedPolicy` api. You can set this here https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac/src/utils/rbac-utils.ts#L146.

plugins/rbac/src/components/CreateRole/PermissionPoliciesFormRow.tsx
```diff
@@ -149,39 +149,37 @@ export const PermissionPoliciesFormRow = ({
...
+            {!!conditionRulesData?.[`${permissionPoliciesRowData.plugin}`]?.[
+              `${permissionPoliciesRowData.permission}`
+            ]?.rules.length && (
+              <IconButton
+                title=""
+                color="primary"
+                hidden={
+                  !permissionPoliciesData?.pluginsPermissions[
+                    permissionPoliciesRowData.plugin
+                  ]?.policies[permissionPoliciesRowData.permission]?.isResourced
```
divyanshiGupta, line 161: Once above comment is resolved you can simple use this `permissionPoliciesRowData.isResourced`. And better to add this above as firstly the resource type needs to be resourced then only it might have rules.

divyanshiGupta (changes requested):

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -180,6 +180,7 @@ export const getPermissionsData = (
               policy: policyTitleCase || 'Use',
               effect: policy.effect,
             }),
+            isResourced: true,
```
divyanshiGupta, line 183: @debsmita1 this value should come from `RoleBasedPolicy`. 
```suggestion
            isResourced: policy.isResourced,
```

divyanshiGupta (changes requested):

plugins/rbac/src/utils/rbac-utils.ts
debsmita1: Thank you for bringing this to my attention. I noticed the isResourced property in the permission policies API response while addressing a different RBAC issue.

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -108,12 +108,20 @@ export const getMembersFromGroup = (group: GroupEntity): number => {
...
 };
 
-export const getPluginId = (
+export const getPluginInfo = (
   permissions: PermissionPolicy[],
-  permission: string | undefined,
-): string =>
-  permissions.find(p => p.policies?.find(pol => pol.permission === permission))
-    ?.pluginId || '-';
+  permissionName: string | undefined,
```
divyanshiGupta, line 113:
```suggestion
  permissionName?: string,
```

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -108,12 +108,20 @@ export const getMembersFromGroup = (group: GroupEntity): number => {
...
+): { pluginId: string; isResourced: boolean } => {
+  const permission = permissions.find(
+    p => p.policies?.find(pol => pol.permission === permissionName),
+  );
+  const pluginId = permission?.pluginId || '-';
+  const isResourced =
+    permission?.policies?.find(p => p.permission === permissionName)
+      ?.isResourced || false;
+
+  return { pluginId, isResourced };
```
divyanshiGupta, line 123: @debsmita1 I will suggest that you use reduce here instead of using find 3 times.

divyanshiGupta (changes requested):

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -108,12 +108,26 @@ export const getMembersFromGroup = (group: GroupEntity): number => {
...
-export const getPluginId = (
+export const getPluginInfo = (
   permissions: PermissionPolicy[],
-  permission: string | undefined,
-): string =>
-  permissions.find(p => p.policies?.find(pol => pol.permission === permission))
-    ?.pluginId || '-';
+  permissionName?: string,
+): { pluginId: string; isResourced: boolean } => {
+  const { pluginId, isResourced } = permissions.reduce(
```
divyanshiGupta, line 115:
```suggestion
return permissions.reduce(
```

plugins/rbac/src/utils/rbac-utils.ts
```diff
@@ -108,12 +108,26 @@ export const getMembersFromGroup = (group: GroupEntity): number => {
...
+          pluginId: p.pluginId || '-',
+          isResourced: policy?.isResourced || false,
+        };
+      }
+      return acc;
+    },
+    { pluginId: '-', isResourced: false },
+  );
+
+  return { pluginId, isResourced };
```
divyanshiGupta, line 129:
```suggestion
```

divyanshiGupta (changes requested):

PR #1703, opened 2024-05-22
fix(rbac): fix to enable create and edit role buttons on having correct permissions

divyanshiGupta:
Fixes: https://issues.redhat.com/browse/RHIDP-1757

Currently `catalog.entity.read` permission auth was incorrect as it expects a complete resource ref for eg  `component:default/test` but `catalog-entity` was used instead and on applying any conditions to the admin role the create button was getting disabled as /api/permission/authorize returns 500 with this error for catalogEntityReadPermission :
"Entity reference \"catalog-entity\" had missing or empty kind (e.g. did not start with \"component:\" or similar)"

To fix this I have remove this permission check. If the user is authorised to read users and groups this /api/catalog/entities?filter=kind=user&filter=kind=group will return a list of users and groups otherwise it returns an empty array. So I have uitilised this to check if the user is authorised or not.

Now if the user doesnt have `catalog-entity, read, allow` or a condition is applied like below

Screenshot 2024-05-22 at 8 25 01 PM

then the create button will be disabled

Screenshot 2024-05-22 at 8 29 11 PM

Error page for route /rbac/role/new when user is not authorised to access users and groups
Screenshot 2024-05-22 at 8 29 48 PM

Error page for route /rbac/role/role/default/div  when user is not authorised to access users and groups
Screenshot 2024-05-22 at 8 29 28 PM

### DIFF
--- a/plugins/rbac/README.md
+++ b/plugins/rbac/README.md
@@ -30,6 +30,8 @@ p, role:default/team_a, policy-entity, create, allow
 g, user:default/<login-id/user-name>, role:default/team_a
 ```
 
+> Note: Even after applying above permissions if the create button is still disabled then please contact your administrator as you might be conditionally restricted to access the create button.
+
 - To fetch the permissions from other plugins like `Kubernetes` and `Jenkins` in the Role Form as mentioned [here](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/permissions.md), add the following configuration in your `app-config.yaml`:
 
 ```yaml title="app-config.yaml"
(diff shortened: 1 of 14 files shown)


divyanshiGupta:
cc: @invincibleJai

invincibleJai:
@divyanshiGupta One observation on the edit icon, in the card `Users and groups` it's disabled but for card `Permission Policies` it is enabled and on click it shows 404

image

invincibleJai:
on admin nav click observed warning message briefly and then the screen, so good to handle it (on slow 3G it's easier to reproduce)


https://github.com/janus-idp/backstage-plugins/assets/5129024/15f67236-49fd-45f7-8ce6-c770fd14f628

invincibleJai:
/lgtm
/approve

Verified the changes works as expected

PR #1704, opened 2024-05-22
fix(orchestrator): fix the common package reference version

caponetto:
The `common` package version was bumped by the bot but its references have not been updated.


gciavarrini (approved):
lgtm

PR #1712, opened 2024-05-23
fix(rbac): enable save on remove-all button click

divyanshiGupta:
Fixes: https://issues.redhat.com/browse/RHIDP-2369

https://github.com/janus-idp/backstage-plugins/assets/20724543/704af55f-1a46-4c7a-9dbd-6f06794eb525

### DIFF
--- a/plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
@@ -126,6 +126,8 @@ export const PermissionPoliciesForm = ({
 
   const onAddConditions = (index: number, conditions?: ConditionsData) => {
     setFieldValue(`permissionPoliciesRows[${index}].conditions`, conditions);
+    if (!conditions)
+      setFieldValue(`permissionPoliciesRows[${index}].id`, undefined);
   };
 
   const onRowRemove = (index: number) => {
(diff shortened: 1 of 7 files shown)


plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
```diff
@@ -126,6 +126,8 @@ export const PermissionPoliciesForm = ({
 
   const onAddConditions = (index: number, conditions?: ConditionsData) => {
     setFieldValue(`permissionPoliciesRows[${index}].conditions`, conditions);
+    if (!conditions)
+      setFieldValue(`permissionPoliciesRows[${index}].id`, undefined);
```
ciiay, line 130: I don't think we should intentionally set value to `undefined`. I know it's been used everywhere, and we don't have a coding standard to follow. Just want to put it out there for discussion.

ciiay (approved):
Verified working great on `backstage-plugins`'s local run with `yarn start:backstage`. Just left a comment for discussion.

plugins/rbac/src/components/CreateRole/PermissionPoliciesForm.tsx
divyanshiGupta: @ciiay thanks for the feedback but why should we not set a value to undefined if its not needed for some case?

PR #1714, opened 2024-05-24
fix(rbac): fix role list view permission policies column value

ciiay:
For [RHIDP-2431](https://issues.redhat.com/browse/RHIDP-2431)

Screen recording

https://github.com/janus-idp/backstage-plugins/assets/26255262/b130e344-ed3a-4edb-be41-95cefdbc556a

Screenshot for error on fetching role conditions
image

### DIFF
--- a/plugins/rbac/src/components/RolesList/RolesList.tsx
+++ b/plugins/rbac/src/components/RolesList/RolesList.tsx
@@ -2,8 +2,7 @@ import React from 'react';
 
 import { Table, WarningPanel } from '@backstage/core-components';
 
-import { Box, makeStyles } from '@material-ui/core';
-import { Alert } from '@material-ui/lab';
+import { makeStyles } from '@material-ui/core';
 
 import { useLocationToast } from '../../hooks/useLocationToast';
 import { useRoles } from '../../hooks/useRoles';
@@ -45,27 +44,42 @@ export const RolesList = () => {
     setRoles(searchResults.length);
   };
 
+  const getErrorWarning = () => {
+    const errorTitleBase = 'Something went wrong while fetching the';
+    const errorWarningArr = [
+      { message: error?.rolesError, title: `${errorTitleBase} roles` },
+      {
+        message: error?.policiesError,
+        title: `${errorTitleBase} permission policies`,
+      },
+      {
+        message: error?.roleConditionError,
+        title: `${errorTitleBase} role conditions`,
+      },
+    ];
+
+    return (
+      errorWarningArr.find(({ message }) => message) || {
+        message: '',
+        title: '',
+      }
+    );
+  };
+
+  const errorWarning = getErrorWarning();
+
   return (
     <>
       <SnackbarAlert toastMessage={toastMessage} onAlertClose={onAlertClose} />
-      {error?.roleConditionError && (
-        <Box>
-          <Alert severity="warning">{error?.roleConditionError}</Alert>
-        </Box>
-      )}
       <RolesListToolbar
         createRoleAllowed={createRoleAllowed}
         createRoleLoading={createRoleLoading}
       />
-      {(error?.rolesError || error?.policiesError) && (
+      {errorWarning.message && (
         <div style={{ paddingBottom: '16px' }}>
           <WarningPanel
-            message={error.rolesError || error.policiesError}
-            title={
-              error.rolesError
-                ? 'Something went wrong while fetching the roles'
-                : 'Something went wrong while fetching the permission policies'
-            }
+            message={errorWarning.message}
+            title={errorWarning.title}
             severity="error"
           />
         </div>
--- a/plugins/rbac/src/components/RolesList/RolesList.tsx
+++ b/plugins/rbac/src/components/RolesList/RolesList.tsx
@@ -2,7 +2,8 @@ import React from 'react';
 
 import { Table, WarningPanel } from '@backstage/core-components';
 
-import { makeStyles } from '@material-ui/core';
+import { Box, makeStyles } from '@material-ui/core';
+import { Alert } from '@material-ui/lab';
 
 import { useLocationToast } from '../../hooks/useLocationToast';
 import { useRoles } from '../../hooks/useRoles';
@@ -47,6 +48,11 @@ export const RolesList = () => {
   return (
     <>
       <SnackbarAlert toastMessage={toastMessage} onAlertClose={onAlertClose} />
+      {error?.roleConditionError && (
+        <Box>
+          <Alert severity="warning">{error?.roleConditionError}</Alert>
+        </Box>
+      )}
       <RolesListToolbar
         createRoleAllowed={createRoleAllowed}
         createRoleLoading={createRoleLoading}
--- a/plugins/rbac/src/hooks/useRoles.ts
+++ b/plugins/rbac/src/hooks/useRoles.ts
@@ -84,21 +84,22 @@ export const useRoles = (
     resourceRef: policyEntityUpdatePermission.resourceType,
   });
 
-  const conditionsPromises: Promise<RoleWithConditionalPoliciesCount>[] =
-    (Array.isArray(roles) &&
-      roles.map(async role => {
-        const conditionalPolicies = await rbacApi.getRoleConditions(role.name);
-
-        return {
-          ...role,
-          conditionalPoliciesCount: Array.isArray(conditionalPolicies)
-            ? conditionalPolicies.length
-            : 0,
-        };
-      })) ||
-    [];
-
   React.useEffect(() => {
+    const conditionsPromises: Promise<RoleWithConditionalPoliciesCount>[] =
+      (Array.isArray(roles) &&
+        roles.map(async role => {
+          const conditionalPolicies = await rbacApi.getRoleConditions(
+            role.name,
+          );
+
+          return {
+            ...role,
+            conditionalPoliciesCount: Array.isArray(conditionalPolicies)
+              ? conditionalPolicies.length
+              : 0,
+          };
+        })) ||
+      [];
     const fetchAllPermissionPolicies = async () => {
       if (Array.isArray(roles)) {
         try {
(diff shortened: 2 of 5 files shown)


ShiranHi:
@ciiay I don't see any issue here. The conditional policy is part of the original policy, functioning as an additional configuration rather than a separate condition. If the intention was to count a conditional policy separately, then we should not consider it as a new policy. In addition, we're going to improve this view to make it clearer which plugins are part of each role, so this column probably will be removed

ciiay:
> @ciiay I don't see any issue here.

Hi @ShiranHi , do you mean the screen recording has no issue, or the description from [RHIDP-2431](https://issues.redhat.com/browse/RHIDP-2431) has no issue? I understand they should be considered as same policy from the user's view, but implementation-wise they are retrieved separately.

If there's no issue with the current behavior I could just close this PR.

ShiranHi:
Hey @ciiay ,It's a bit confusing, so let me clarify :)
In the Jira ticket, it appears that users see "0" in the main table list when they add a conditional policy. However, in the video attached to this PR, everything seems to be working fine. Was the screen recording added as a solution to the Jira ticket, or was it intended to showcase the problem that described in the ticket? If this is the solution - then great, it looks good

ShiranHi:
Thank you @ciiay for the explanation!

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -70,42 +75,76 @@ export const useRoles = (
...
+                role.name,
+              );
+
+              return {
+                ...role,
+                conditionalPolicies: Array.isArray(conditionalPolicies)
+                  ? conditionalPolicies.length
+                  : 0,
+              };
+            }),
```
divyanshiGupta, line 95: @ciiay I will suggest that we dont do these computations inside `Promise.all` to reduce code complexity and increase readability. You can create an array of promises like this and then use `Promise.all`:
```
conditionsPromises = roles.map(role =>  rbacApi.getRoleConditions(role.name));
const conditions = await Promise.all(conditionsPromises);
```
After this you can have another function to map conditionsCount to respective roles using `roleEntityRef` param from the condition object. This will make the code a bit more linear as it would be divided into isolated steps.

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
divyanshiGupta, line 1: @ciiay since here we are importing `React` you dont need to import `useEffect` separately.

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -70,42 +77,79 @@ export const useRoles = (
...
+              ? conditionalPolicies.length
+              : 0,
+          };
+        })) ||
+      [];
+    const fetchAllPermissionPolicies = async () => {
+      if (Array.isArray(roles)) {
+        try {
+          const updatedRoles: RoleWithConditionalPoliciesCount[] =
+            await Promise.all(conditionsPromises);
```
divyanshiGupta, line 101: FYI Promise.all will never throw an error because of how the api calls are handled above as in case one of the api calls fail there will always still be an object just the conditionalPoliciesCount will be 0. It should be okay to keep the count as 0 for the api call which failed but the current code is not correct as try and catch will not be used here as the code expects. Also if we keep the count as 0 user will not know what happened so ideally user should know for which role this api call failed and still calculate the count for others who passed. Also I think you should not make the roles data generation dependent upon this function as for example if the code actually throws an error once you fix it the roles will never set and hence even the roles data will not be generated.

divyanshiGupta (changes requested):

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -70,42 +77,79 @@ export const useRoles = (
     permission: policyEntityUpdatePermission,
     resourceRef: policyEntityUpdatePermission.resourceType,
   });
+
+  React.useEffect(() => {
+    const conditionsPromises: Promise<RoleWithConditionalPoliciesCount>[] =
+      (Array.isArray(roles) &&
```
divyanshiGupta, line 83: You can move this inside of fetchAllPermissionPolicies function.

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -70,42 +77,79 @@ export const useRoles = (
...
+          return {
+            ...role,
+            conditionalPoliciesCount: Array.isArray(conditionalPolicies)
+              ? conditionalPolicies.length
+              : 0,
+          };
+        })) ||
+      [];
+    const fetchAllPermissionPolicies = async () => {
+      if (Array.isArray(roles)) {
```
divyanshiGupta, line 98: Not required as you have already added this check here https://github.com/janus-idp/backstage-plugins/pull/1714/files#diff-de9e9b3435ef119eaf0550b2d4ff253ec247e8b76f386f0d5fadd8e111218aa7R109

divyanshiGupta (changes requested):

plugins/rbac/src/hooks/useRoles.ts
ciiay: @divyanshiGupta What is the best way to handle any role condition fetching failures? I can still return the role but if the condition fetching failed, do we need to let the user know at meanwhile show the role data in the list just missing condition data?

divyanshiGupta: So if you see here https://github.com/janus-idp/backstage-plugins/blob/b83a315113ef69c5a85f0a100d730a99a4590b48/plugins/rbac/src/hooks/useRoles.ts#L172 error is being set. You can have another param for conditions-error
If you follow this:
```
conditionsPromises = roles.map(role =>  rbacApi.getRoleConditions(role.name));
const conditions = await Promise.all(conditionsPromises);
```
the conditions will be an array of RoleConditionalPolicyDecision<PermissionAction>[] | Response. You will need extract out the Responses and use its statusText for the error message.
```
const err = conditions
            .map(c => (c as unknown as Response)?.statusText)
            .filter(m => m);

          if (err.length > 0) {
            throw new Error(
              `Unable to retrieve conditions. ${err.join('\n')}`,
            );
          }
```
And set the error in state in catch and use it here https://github.com/janus-idp/backstage-plugins/blob/b83a315113ef69c5a85f0a100d730a99a4590b48/plugins/rbac/src/hooks/useRoles.ts#L172. (edited)

ciiay: I have added an alert warning when some role condition fetching fails. I made it a warning message as it won't block roles data listing. Let me know how do you think.
image

plugins/rbac/src/components/RolesList/RolesList.tsx
```diff
@@ -47,6 +48,11 @@ export const RolesList = () => {
   return (
     <>
       <SnackbarAlert toastMessage={toastMessage} onAlertClose={onAlertClose} />
+      {error?.roleConditionError && (
+        <Box>
+          <Alert severity="warning">{error?.roleConditionError}</Alert>
+        </Box>
+      )}
```
divyanshiGupta, line 55: @ciiay there is already a warning panel below RolesListToolbar for simple policies error why dont you use that instead of adding a new alert?

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -138,6 +186,7 @@ export const useRoles = (
         (typeof policies === 'object'
           ? (policies as any as Response)?.statusText
           : '')) as string,
+      roleConditionError: fetchRoleConditionsError,
```
divyanshiGupta, line 189: @ciiay can we pass this in the error object here https://github.com/janus-idp/backstage-plugins/blob/b83a315113ef69c5a85f0a100d730a99a4590b48/plugins/rbac/src/hooks/useRoles.ts#L172 instead of creating a new param for it.

divyanshiGupta (changes requested):

plugins/rbac/src/components/RolesList/RolesList.tsx
ciiay: Updated screenshot
image

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -76,39 +86,77 @@ export const useRoles = (
...
+          );
+          return {
+            ...role,
+            conditionalPoliciesCount: Array.isArray(conditionalPolicies)
+              ? conditionalPolicies.length
+              : 0,
+          };
+        } catch (error) {
+          setRoleConditionError(
+            `Error fetching role conditions for role ${role.name}, please try again later.`,
```
divyanshiGupta, line 106: Append the statusText here.

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -76,39 +86,77 @@ export const useRoles = (
...
     resourceRef: policyEntityUpdatePermission.resourceType,
   });
+
+  React.useEffect(() => {
+    const fetchAllPermissionPolicies = async () => {
+      if (!Array.isArray(roles)) return;
+      const conditionPromises = roles.map(async role => {
+        try {
+          const conditionalPolicies = await rbacApi.getRoleConditions(
+            role.name,
```
divyanshiGupta, line 96: @ciiay you need to manually throw the error here because of you see the api file the apis doenst throw error so we are returning the error response. You need to use that error response to get the error msg and throw it manually.

```
if((conditionaPolicies as unknown as Response)?.statusText) throw new Error((conditionaPolicies as unknown as Response).statusText);
```

divyanshiGupta (changes requested):

plugins/rbac/src/hooks/useRoles.ts
```diff
@@ -76,39 +86,77 @@ export const useRoles = (
     permission: policyEntityUpdatePermission,
     resourceRef: policyEntityUpdatePermission.resourceType,
   });
+
+  React.useEffect(() => {
+    const fetchAllPermissionPolicies = async () => {
+      if (!Array.isArray(roles)) return;
```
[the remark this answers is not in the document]
ciiay: Removing this condition causes compiler error says "'roles' is possibly 'undefined'."
image

divyanshiGupta:
@ciiay CI is failing.

PR #1727, opened 2024-05-28
fix(orchestrator): upgrade to mui v5

caponetto:
Adapting the code to upgrade the plugin to mui v5 as the bot in https://github.com/janus-idp/backstage-plugins/pull/1655 cannot handle it.

Required changes:
- Use full path for imported components
- Use `styled` in favor of `makeStyles`
- Remove `ref` from Dialogs


batzionb:
Thanks @caponetto!

PR #1728, opened 2024-05-28
fix(rbac): add proper empty page for RBAC plugin

ciiay:
For [RHIDP-2344](https://issues.redhat.com/browse/RHIDP-2344): Show empty state when rbac frontend is not available
(Updated on 5/30)
```
      <ErrorPage
        status="404"
        statusMessage="Enable the RBAC backend plugin to use this feature."
        additionalInfo="To enable RBAC, set `permission.enabled` to `true` in the app-config file."
      />
```

@ShiranHi @divyanshiGupta Let me know if I need to update the `statusMessage` or the `addtionalInfo` content.

Screenshot(updated on 5/30):
image

### DIFF
--- a/plugins/rbac/src/components/Router.test.tsx
+++ b/plugins/rbac/src/components/Router.test.tsx
@@ -32,6 +32,10 @@ jest.mock('./CreateRole/EditRolePage', () => ({
   EditRolePage: () => <div>EditRole</div>,
 }));
 
+jest.mock('@backstage/core-components', () => ({
+  ErrorPage: jest.fn().mockImplementation(() => <div>Mocked ErrorPage</div>),
+}));
+
 jest.mock('@backstage/plugin-permission-react', () => ({
   RequirePermission: jest
     .fn()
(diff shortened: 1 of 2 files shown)


ShiranHi:
@ciiay , yes, we have design for this screen, please take a look [here](https://www.figma.com/design/E6C9X3OSVzlWkW7ZsbxN7s/Administration?node-id=4507-4016&t=7HgmiULcWqQrmFyV-1). 

Can we do these changes?
1. Remove the small text on top "ERROR 404: RBAC-backend plugin..."
2. Change the title "Looks like someone drop the mic!" to be "404: Enable the RBAC backend plugin to use this feature"
3. Instead of "Go back... or please..." can we write "To enable RBAC, set ‘permission.enabled’ to ‘true’ in the  app-config file."
4. And adding a "RBAC backend plugin" button that will send users to [this page](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/README.md)
5. Can we change the default image there?

ciiay:
Hi @ShiranHi , thanks for the review. In this case will this new Error Page be used universally across our app? The content part will be editable in different cases. For this PR I just reused the existing ErrorPage component from upstream Backstage. If we need to make a new one for us then I can update the ticket to add a criteria such as "create a new universal Error Page".
cc. @divyanshiGupta

ShiranHi:
> In this case will this new Error Page be used universally across our app?
Yes, we want to create new error pages for our RHDH plugins, you can find the design [here](https://www.figma.com/design/GeVb7rc390TkxeQNnJIcBD/RHDH-Theme-(pages)?node-id=110%3A395&t=gHtMURhMtSLuLO4B-1).
If we can include it in this PR, that would be great. However, if you think it's too much work and requires a dedicated feature, that's fine too.

divyanshiGupta:
@ciiay @ShiranHi if we want to update the error pages universally across the app then it makes sense to have a separate story for it. For now we can go with the backtsage default error page for consistency.

ciiay:
>if we want to update the error pages universally across the app then it makes sense to have a separate story for it. For now we can go with the backtsage default error page for consistency.

In this case, I have updated the error message on the current backstage ErrorPage for RBAC plugin, @ShiranHi can you confirm the updated screenshot again? And @divyanshiGupta this is ready for review. Thank you both.

ShiranHi:
@ciiay looks good, thank you

PR #1726, opened 2024-05-28
feat(rbac): add audit log for RBAC backend

AndrienkoAleksandr:
### What does this pull request do

Implement audit logging for RBAC backend

### Referenced issue:
https://issues.redhat.com/browse/RHIDP-2004

PR #1622, opened 2024-05-08
feat(audit-log): add common audit-log package

Zaperex:
### Description
Add initial common package for the audit logging
Fixes: [RHIDP-2287](https://issues.redhat.com/browse/RHIDP-2287)

PR #1730, opened 2024-05-28
fix(audit-log): deep copy objects passed into audit logger so redactor will not modify objects and remove `auditErrorLog` function

Zaperex:
### Description
Fixes a bug where objects passed into the audit logger may potentially be modified if they contain values that would be redacted by the logger service redactor.

Also removes the `auditErrorLog` function in favor of making the `auditLog` function more generic.


Zaperex:
https://github.com/janus-idp/backstage-plugins/pull/1622#discussion_r1617951723
> Removed the `auditErrorLog` function in https://github.com/janus-idp/backstage-plugins/pull/1730/commits/35eb36afdbce0cb4e8fb5434d00c0e92a1edd7cc in favor of using the `auditLog` function for everything for simplicity. 
> 
> This would require the following changes:
> - All usages of `auditErrorLog` to be changed to `auditLog`
> - A `status` field must be provided to all `auditLog` function calls
> - A `level: 'error'` field should be provided to all `auditLog` calls converted from `auditErrorLog`

Zaperex:
Considering how the logger redactor will redact anything in logged through the root logger, I think there's a concern that important things like the `actorId`, `ip`, or other details of the audit log MIGHT get redacted. Would that be much of a concern? I personally don't think it's likely to happen.

AndrienkoAleksandr:
Looks good to me.

PatAKnight:
@Zaperex @AndrienkoAleksandr is this something that could potentially wait until after 1.2 so that we can get this [PR](https://github.com/janus-idp/backstage-plugins/pull/1726) is merged?

Zaperex:
@PatAKnight this PR fixes a bug where the audit log modifies values passed into it if it gets redacted by the log redactor, so I think it's quite important to have? I can potentially move the structure modification to a different PR if needed. But the package has not been published yet due to the pipeline being broken so https://github.com/janus-idp/backstage-plugins/pull/1726 would not exactly work due to it already trying to import the package.

schultzp2020 (approved):
/lgtm
/approve

PR #1734, opened 2024-05-29
feat(kiali): sanitized input

josunect:


josunect:
Done in https://github.com/janus-idp/backstage-plugins/pull/1786

PR #1739, opened 2024-05-29
fix(rbac): do not disable already selected rule for allOf/anyOf

divyanshiGupta:
Fixes: https://issues.redhat.com/browse/RHIDP-2453

Screenshot 2024-05-29 at 8 18 52 PM


ciiay (approved):
Verified working as expected.
/lgtm

PR #1741, opened 2024-05-29
fix(cli): fix entrypoint validation on backstage 1.24.0 and above.

davidfestal:
Starting with backstage release 1.24.0, the use of `ts-node` to manage the typescript compilation has been replaced by the use of the `@backstage/cli/config/nodeTransform.cjs` script based on `pirates` and `@swc/core`  (see commit https://github.com/backstage/backstage/commit/2c2f705fe169f72fdac4da84c154d7118598d06f).

We should adapt to this in the the validation of the entrypoints part of the Janus CLI `expot-dynamic-plugin` command new implementation for backend plugins.

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/backend-embed-as-dependencies.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/backend-embed-as-dependencies.ts
@@ -796,21 +796,41 @@ function validatePluginEntryPoints(target: string): string {
     Task.log(
       `  adding typescript extension support to enable entry point validation`,
     );
-    let tsNode: string | undefined;
+
+    let nodeTransform: string | undefined;
     try {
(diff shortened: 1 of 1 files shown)


nickboldt:
code could use some inline comments (it's unclear to me what you're doing and why) but otherwise lgtm. 

why do we need support here for 1.24 and earlier if this is only for 1.26+ ?

davidfestal:
> why do we need support here for 1.24 and earlier if this is only for 1.26+ ?

I still wanted to keep the previous approach, because we want this command to also work on environments setup by third-parties that would still be based on backstage `v1.23.4`, which is the version backing RHDH 1.1.

We want the same CLI to support exporting dynamic plugins from third-party repositories for both RHDH 1.1 and 1.2.

gashcrumb (approved):

PR #1745, opened 2024-05-30
chore(deps): fix dependency issues on main

jerolimov:
This PR just merges #1744 (+ an updated `yarn.lock` from a single `yarn install` run) and #1740

/cc @nickboldt @ciiay


divyanshiGupta:
Thanks @jerolimov. I will be merging this to unblock other PRs for now.

/lgtm

PR #1755, opened 2024-05-31
fix(rbac): improve criteria toggle button readability on dark themes

jerolimov:
The criteria toggle button (ButtonGroup or ToggleButtonGroup) sets a custom style for the inactive options with a too low contrast.

When removing the color, these button colors were blue by default because all buttons have a blue color by default.

To remove that, I switched to the semantic, more matching ToggleButtonGroup and ToggleButton here so that it matches almost automatically our design.

Here is the sidebar design change (with the latest theme already):

| Before | After |
|---|---|
| backstage-light-before | backstage-light-after |
| backstage-dark-before | backstage-dark-after |
| rhdh-light-before | rhdh-light-after |
| rhdh-dark-before | rhdh-dark-after |

/cc @divyanshiGupta @debsmita1


debsmita1:
tested this locally
/lgtm

Issue #1492, opened 2024-04-10
RBAC Role Data Out-of-Sync When Scaling Horizontally


zthfmh:
### Describe the bug

Hello, we've discovered some inconsistent behavior from the [rbac-backend](https://github.com/janus-idp/backstage-plugins/tree/main/plugins/rbac-backend) plugin when scaling the backstage deployment to 2 pods/containers with identical instance. The use case we had issue with is when using the API endpoint `PUT </api/permission/roles/:kind/:namespace/:name>` (https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/apis.md#put-role) to update membership of a role. When having two Kubernetes pods running, the API request was received and processed by pod A, in which itself also updated the DB with the new role definition. Everything seemed good at this point until someone started querying the endpoint `GET </api/permission/roles/:kind/:namespace/:name>` (https://github.com/janus-idp/backstage-plugins/blob/main/plugins/rbac-backend/docs/apis.md#get-role) to retrieve the role - the updated role data was returned when a request hit pod A, however if the same request was received by pod B, we'd still get the old role definition. 

Since the DB appears to be successfully updated, we are wondering if the plugin utilizes some kind of in memory cache that caused this out-of-sync behavior. Our backstage deployment has been configured to using Redis as the cache backend and it's seemingly not taking advantage of that. Does anyone have any thoughts on this? Thanks in advance.

### Expected Behavior

When having a high availability deployment setup, API endpoints served by the RBAC backend plugin should provide consistent data across all instances.

### What are the steps to reproduce this bug?

1. Horizontally scale the backstage deployment to have 2 replicas
2. Choose any existing RBAC role, and retrieve its current definition with the endpoint `GET </api/permission/roles/:kind/:namespace/:name>` 
3. Modify the retrieved definition by adding a new member to the role. Feed the endpoint `PUT </api/permission/roles/:kind/:namespace/:name>` with the new definition to update the role
4. Querying `GET </api/permission/roles/:kind/:namespace/:name>`, you will see that the data returned varies depending on the pod/container being hit 

### Versions of software used and environment

@backstage/app-defaults	1.5.3
@backstage/backend-app-api	0.5.14, 0.6.2
@backstage/backend-common	0.19.9, 0.20.1, 0.21.6
@backstage/backend-dev-utils	0.1.4
@backstage/backend-openapi-utils	0.1.9
@backstage/backend-plugin-api	0.6.16
[...]

PR #1757, opened 2024-05-31
fix(rbac): add support for scaling

PatAKnight:
## Description

We were using an in memory cache of all of the roles that the RBAC Backend plugin was aware of for quick look up and evaluation. We are realizing that this approach is not sufficient in the event that a user scales Backstage out horizontally. The problem is that whenever one instance would add, update, or delete a role, the in memory cache of the other instances would not be updated.

This PR introduces the ability to query the RBAC database table `casbin_rule` for the roles that are currently stored ensuring that roles between each instance match. This is a first approach to get something working when scaling until we have a chance to evaluate the cache system of Backstage.

## Fixes
- Fixes: [RHIDP-1994](https://issues.redhat.com/browse/RHIDP-1994)
- Fixes: #1492


plugins/rbac-backend/src/role-manager/member-list.ts
```diff
@@ -0,0 +1,124 @@
...
+    userAndGroups: string[],
+    client: Knex,
+  ): Promise<void> {
+    try {
+      const roles: string[] = await client
+        .table('casbin_rule')
+        .whereIn('v0', userAndGroups)
+        .pluck('v1');
+
+      roleMemberList.addRoles(roles);
```
AndrienkoAleksandr, line 111: Sorry, but roles and members can be duplicated in the array. Are we going to handle that?

AndrienkoAleksandr: I hope that our validator is good enough

AndrienkoAleksandr (approved):

PR #1786, opened 2024-06-05
fix(kiali): sanitize input

josunect:
- Sanitize input before appendChild
- Remove 'x-powered-by' from test express


aljesusg (approved):
LGFM

PR #1790, opened 2024-06-05
feat(kiali): add tests for the overview page

matejnesuta:
I added some tests for the header and also for the Overview cards. Links to other pages and graphs cannot be tested yet, as they do not work properly.

### DIFF
--- a/plugins/kiali/tests/kialiPage.spec.ts
+++ b/plugins/kiali/tests/kialiPage.spec.ts
@@ -71,17 +71,8 @@ test.describe('Kiali page', () => {
       );
     });
 
-    test('Debug info can be opened', async () => {
-      await page.click('data-test=help-button');
-      await page.click('text=View Debug Info');
-      await expect(
-        page.locator('css=[aria-describedby="Debug information"]'),
-      ).toBeVisible();
-    });
-
     test('About page can be opened', async () => {
       await page.click('data-test=help-button');
-      await page.click('text=About');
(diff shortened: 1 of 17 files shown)


plugins/kiali/tests/kialiPage.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+      page.reload();
+    });
+
+    test('Home cluster is visible', async () => {
+      await expect(page.locator('data-test=home-cluster')).toHaveText(
+        CONFIG.clusters.Kubernetes.name,
+      );
+    });
+
+    test('Debug info can be opened', async () => {
```
josunect, line 74: That feature was removed, so this test should be removed

plugins/kiali/tests/kialiPage.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+      await page.click('data-test=help-button');
+      await page.click('text=View Debug Info');
+      await expect(
+        page.locator('css=[aria-describedby="Debug information"]'),
+      ).toBeVisible();
+    });
+
+    test('About page can be opened', async () => {
```
josunect, line 82: Now the about page can be opened with just one click, so this test would need to be redone (Some steps less)

josunect (changes requested):
Looks good, except for a couple of changes

josunect (approved):
lgmt

PR #1794, opened 2024-06-06
fix(kiali): add csrf in express (test)

josunect:


PatAKnight:
Hello @josunect,

I see this has been opened for some time. Is this something that is still needed or should we go ahead and close it out?

PR #1798, opened 2024-06-07
fix(feedback): update EntityPage component

yashoswalyo:
- changed button variant to contained
  image

- change secondary color to error for Create Modal
  image

- Moved jira note above in modal to make it more noticable
  image


deshmukhmayur (approved):
LGTM👍

PR #1800, opened 2024-06-07
fix(topology): remove check for catalog entity permission

PatAKnight:
## Description

There is an issue where using a conditional policy that involves the catalog entity read permission will cause an error to be thrown when attempting to access the Topology page. This PR aims to remove the catalog entity read permission check from the Topology page which in turn will fix the error. 

# Fixes

- Fixes: [RHIDP-2594](https://issues.redhat.com/browse/RHIDP-2594)


PatAKnight:
cc @divyanshiGupta @AndrienkoAleksandr

AndrienkoAleksandr (approved):

PR #1804, opened 2024-06-10
fix(orchestrator): fix error handling in case data index failed to start

batzionb:
https://issues.redhat.com/browse/FLPATH-1381
resolved the issue by calling the data index also when the cache is empty, so it'll fail in case data index isn't available

### DIFF
--- a/plugins/orchestrator-backend/src/service/OrchestratorService.test.ts
+++ b/plugins/orchestrator-backend/src/service/OrchestratorService.test.ts
@@ -141,17 +141,13 @@ describe('OrchestratorService', () => {
     });
 
     it('should throw error when data index returns error', async () => {
-      let err: Error | null = null;
       const errMsg = 'Failed to load instances';
       dataIndexServiceMock.fetchInstances = jest.fn().mockImplementation(() => {
         throw new Error(errMsg);
       });
-      try {
-        await orchestratorService.fetchInstances({});
-      } catch (err_) {
-        err = err_ as Error;
-      }
-      expect(err?.message).toEqual(errMsg);
+
+      const promise = orchestratorService.fetchInstances({});
+      await expect(promise).rejects.toThrow(errMsg);
     });
 
     it('should execute the operation', async () => {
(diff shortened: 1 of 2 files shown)


plugins/orchestrator-backend/src/service/OrchestratorService.test.ts
```diff
@@ -178,26 +197,30 @@ describe('OrchestratorService', () => {
...
+        .fn()
+        .mockImplementation(() => {
+          throw new Error(errMsg);
+        });
+      try {
+        await orchestratorService.fetchWorkflowOverviews({});
+      } catch (err_) {
+        err = err_ as Error;
+      }
+      expect(err?.message).toEqual(errMsg);
```
gciavarrini, line 220: Is it possible to simplify this into  `expect(promise).rejects.toEqual(errMsg)`?
Eventually, this applies to other sections in this PR.

plugins/orchestrator-backend/src/service/OrchestratorService.ts
```diff
@@ -54,19 +54,13 @@ export class OrchestratorService {
   public async fetchInstances(args: {
     pagination?: Pagination;
   }): Promise<ProcessInstance[]> {
-    if (this.workflowCacheService.isEmpty()) {
-      return [];
-    }
```
gciavarrini, line 59: I 'd like to understand if there can be consequences in removing these sections

batzionb: The updated process works as follows:

We always call the APIs and send the list of IDs that are in the cache.
Before the fix, we didn't call the API if the cache was empty thus hiding the error in case the DataIndex fails.
After the fix, we now call the API even when the cache is empty, sending an empty list.
If the API call fails, an error will be displayed.

plugins/orchestrator-backend/src/service/OrchestratorService.test.ts
batzionb: https://github.com/janus-idp/backstage-plugins/pull/1804/commits/6645148eda0872f0c26635e92aace8936fd0f846

gciavarrini (approved):
lgtm

Issue #1709, opened 2024-05-23
Users without email breaking all of the backend part and backstage fully crashes


jonatanask:
### Describe the bug

When imported users from gitlab doesn't have emails because their accounts are private, the system fully crashes and backstage stops working

`TypeError: Cannot read properties of undefined (reading 'email')
    at /app/node_modules/@janus-idp/backstage-plugin-feedback-backend/dist/cjs/plugin-DV3-jMW_.cjs.js:375:160
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)`

### Expected Behavior

If there is no email, just don't send email. Or possibility to not send emails and to not use jira. Just created feedbacks in database

### What are the steps to reproduce this bug?

1. When creating user kind, do not add email
```
apiVersion: backstage.io/v1alpha1
kind: User
metadata:
  name: guest
spec:
  memberOf: [guests]
  profile:
    displayName: Guest
```
2. Create a feedback with that user, for example guest

yashoswalyo:
Made email as optional, update to latest version of backend plugin

Issue #1679, opened 2024-05-16
Feedback Plugin


boueida:
### What do you want to improve?
Do a JIRA integration without having to do the email integration and vice versa. Also be able to create feedback without having to send an email. The email integration is sometimes not needed.

### What is the current behavior?
Need to create the email backend for other parts like JIRA and feedback to work

### What will the new behavior be?
Be able to specify only JIRA, or only email in the feedback plugin.

yashoswalyo:
Made email as optional, update to latest version of backend plugin and try now

PR #1809, opened 2024-06-12
fix(feedback): fixed feedback backend plugin crashing

yashoswalyo:
Fixed feedback plugin crashing when integrations are not found
Fixes:
- #1709 
- #1679

### DIFF
--- a/plugins/feedback-backend/src/service/router.ts
+++ b/plugins/feedback-backend/src/service/router.ts
@@ -5,7 +5,7 @@ import {
 } from '@backstage/backend-common';
 import { AuthService, LoggerService } from '@backstage/backend-plugin-api';
 import { CatalogClient } from '@backstage/catalog-client';
-import { Entity } from '@backstage/catalog-model';
+import { Entity, UserEntityV1alpha1 } from '@backstage/catalog-model';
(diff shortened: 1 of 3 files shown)


plugins/feedback-backend/src/service/router.ts
```diff
@@ -108,31 +109,38 @@ export async function createRouter(
...
           host = serviceConfig.getString('host');
           const authToken = serviceConfig.getString('token');
           const hostType = serviceConfig.getOptionalString('hostType');
 
           const projectKey = entityRef.metadata.annotations['jira/project-key'];
           const jiraService = new JiraApiService(host, authToken, hostType);
-          const jiraUsername =
-            await jiraService.getJiraUsernameByEmail(reporterEmail);
+          const jiraUsername = reporterEmail
+            ? await jiraService.getJiraUsernameByEmail(reporterEmail)
```
invincibleJai, line 142: should this await be under try catch block?

yashoswalyo: it can be under **try-catch**, but its not required as the function won't be called if reporterEmail is `undefined`

PR #1814, opened 2024-06-14
feat(rbac): nested condition

ciiay:
For [RHIDP-2553](https://issues.redhat.com/browse/RHIDP-2553): Add support for creating nested conditions

Screen recording(updated on 7/1):

https://github.com/janus-idp/backstage-plugins/assets/26255262/7127740e-e988-43e2-a5da-df621e2d0b1d

Screenshot for multiple level nested condition warning message(updated on 8/5):
image

<details>
<summary>Demo: (uploaded on 8/2)</summary>

https://github.com/user-attachments/assets/f13211db-59c3-4253-8fd7-8726b7196b0e

</details>

### DIFF
--- a/plugins/rbac/src/components/ConditionalAccess/AddNestedConditionButton.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/AddNestedConditionButton.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+
+import { Box } from '@material-ui/core';
+import Tooltip from '@material-ui/core/Tooltip';
+import HelpOutlineIcon from '@material-ui/icons/HelpOutline';
+
+export const AddNestedConditionButton = () => {
+  const tooltipTitle = () => (
+    <div>
+      <p style={{ textAlign: 'center' }}>
+        Nested conditions are <b>1 layer rules within a main condition</b>. It
+        lets you allow appropriate access by using detailed permissions based on
+        various conditions. You can add multiple nested conditions.
+      </p>
+      <p style={{ textAlign: 'center' }}>
+        For example, you can allow access to all entity types in the main
+        condition and use a nested condition to limit the access to entities
+        owned by the user.
+      </p>
+    </div>
+  );
+  return (
+    <Box
+      style={{
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+      }}
+    >
+      <span>Add Nested Condition</span>
+      <Tooltip title={tooltipTitle()} placement="top">
+        <HelpOutlineIcon fontSize="inherit" style={{ marginLeft: '0.25rem' }} />
+      </Tooltip>
+    </Box>
+  );
+};
--- a/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
@@ -81,7 +81,9 @@ export const ConditionsFormRow = ({
       case criterias.not:
         if (
           conditionRow.not &&
-          criteriaTypes.includes(Object.keys(conditionRow.not)[0])
+          criteriaTypes.includes(
+            Object.keys(conditionRow.not)[0] as keyof ConditionsData,
+          )
         ) {
           nestedConditions.push(conditionRow.not);
           setNotConditionType(NotConditionType.NestedCondition);
@@ -94,7 +96,7 @@ export const ConditionsFormRow = ({
     setNestedConditionRow(nestedConditions);
   }, [conditionRow, criteria]);
 
-  const handleCriteriaChange = (val: string) => {
+  const handleCriteriaChange = (val: keyof ConditionsData) => {
     setCriteria(val);
     setErrors(resetErrors(val));
 
@@ -510,28 +512,26 @@ export const ConditionsFormRow = ({
       {criteria !== criterias.condition && (
         <Box>
           {criteria !== criterias.not &&
-            (
-              conditionRow[
-                criteria as keyof ConditionsData
-              ] as PermissionCondition[]
-            )?.map((c, srIndex) => (
-              <ComplexConditionRow
-                key={`${criteria}-simple-condition-${srIndex}`}
-                conditionRow={conditionRow}
-                nestedConditionRow={nestedConditionRow}
-                criteria={criteria}
-                onRuleChange={onRuleChange}
-                updateRules={updateRules}
-                setErrors={setErrors}
-                setRemoveAllClicked={setRemoveAllClicked}
-                conditionRulesData={conditionRulesData}
-                notConditionType={notConditionType}
-                classes={classes}
-                currentCondition={c}
-                ruleIndex={srIndex}
-                activeCriteria={criteria as 'allOf' | 'anyOf'}
-              />
-            ))}
+            (conditionRow[criteria] as PermissionCondition[])?.map(
+              (c, srIndex) => (
+                <ComplexConditionRow
+                  key={`${criteria}-simple-condition-${srIndex}`}
+                  conditionRow={conditionRow}
+                  nestedConditionRow={nestedConditionRow}
+                  criteria={criteria}
+                  onRuleChange={onRuleChange}
+                  updateRules={updateRules}
+                  setErrors={setErrors}
+                  setRemoveAllClicked={setRemoveAllClicked}
+                  conditionRulesData={conditionRulesData}
+                  notConditionType={notConditionType}
+                  classes={classes}
+                  currentCondition={c}
+                  ruleIndex={srIndex}
+                  activeCriteria={criteria as 'allOf' | 'anyOf'}
+                />
+              ),
+            )}
           {criteria === criterias.not && (
             <RadioGroup
               className={classes.radioGroup}
@@ -541,12 +541,12 @@ export const ConditionsFormRow = ({
               }
             >
               <FormControlLabel
-                value="simple-condition"
+                value={NotConditionType.SimpleCondition}
                 control={<Radio color="primary" />}
                 label="Add rule"
                 className={classes.radioLabel}
               />
-              {notConditionType === 'simple-condition' && (
+              {notConditionType === NotConditionType.SimpleCondition && (
                 <ConditionsFormRowFields
                   oldCondition={
                     conditionRow.not ?? getDefaultRule(selPluginResourceType)
@@ -568,7 +568,7 @@ export const ConditionsFormRow = ({
                 />
               )}
               <FormControlLabel
-                value="nested-condition"
+                value={NotConditionType.NestedCondition}
                 control={<Radio color="primary" />}
                 label={<AddNestedConditionButton />}
                 className={classes.radioLabel}
--- a/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
@@ -2,67 +2,36 @@ import React from 'react';
 
 import { PermissionCondition } from '@backstage/plugin-permission-common';
 
-import { IconButton, makeStyles, useTheme } from '@material-ui/core';
-import ToggleButton from '@material-ui/lab/ToggleButton';
-import ToggleButtonGroup from '@material-ui/lab/ToggleButtonGroup';
+import {
+  FormControlLabel,
+  IconButton,
+  Radio,
+  RadioGroup,
+  useTheme,
+} from '@material-ui/core';
 import AddIcon from '@mui/icons-material/Add';
 import RemoveIcon from '@mui/icons-material/Remove';
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
+import ToggleButtonGroup from '@mui/material/ToggleButtonGroup';
 
+import {
+  extractNestedConditions,
+  getDefaultRule,
+  nestedConditionButtons,
+  ruleOptionDisabled,
+  useConditionsFormRowStyles,
+} from '../../utils/conditional-access-utils';
+import { AddNestedConditionButton } from './AddNestedConditionButton';
 import { ConditionsFormRowFields } from './ConditionsFormRowFields';
 import { conditionButtons, criterias } from './const';
-import { ConditionsData, RuleParamsErrors, RulesData } from './types';
-
-const useStyles = makeStyles(theme => ({
-  conditionRow: {
-    padding: '20px',
-    border: `1px solid ${theme.palette.border}`,
-    borderRadius: '4px',
-    backgroundColor: theme.palette.background.default,
-    '& input': {
-      color: `${theme.palette.textContrast}!important`,
-      '&:-internal-autofill-selected, &:-webkit-autofill, &:-webkit-autofill:hover, &:-webkit-autofill:focus, &:-webkit-autofill:active':
-        {
-          WebkitBoxShadow: `0 0 0px 1000px ${theme.palette.background.paper} inset`,
-          WebkitTextFillColor: `${theme.palette.textContrast}!important`,
-          caretColor: `${theme.palette.textContrast}!important`,
-        },
-    },
-    '& button': {
-      textTransform: 'none',
-    },
-  },
-  criteriaButtonGroup: {
-    backgroundColor: theme.palette.background.paper,
-    width: '80%',
-  },
-  criteriaButton: {
-    width: '100%',
-    padding: `${theme.spacing(1)}px !important`,
-  },
-  addRuleButton: {
-    color: theme.palette.primary.light,
-    marginTop: theme.spacing(1),
-  },
-  removeRuleButton: {
-    color: theme.palette.grey[500],
-    flexGrow: 0,
-    alignSelf: 'baseline',
-    marginTop: theme.spacing(3.3),
-  },
-}));
-
-type ConditionFormRowProps = {
-  conditionRulesData?: RulesData;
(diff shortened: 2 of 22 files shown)


divyanshiGupta:
Thanks @ciiay for the PR.

I am not sure if you have seen this https://docs.google.com/document/d/1UtQeeTwPvUndxEInjvEt9JgKIA5Pnrkl5ImFmGsAjR8/edit?disco=AAABPVRBMts but there are plans for adding conditions via a file and it can have any levels of nesting. From our end we should make sure if this happens the UI doesnt break as we are only supporting 1 level of nesting for now. For testing this you can create multi level nested conditions using the CLI and add necessary checks in UI so that it doesnt break in case a multi level nested condition is present.

ciiay:
Hi @divyanshiGupta , thanks for the tips. I have tested with third level conditional permission policies with mocked data in dev env and it shows empty rule rows as shown in the below screenshot. As the nested conditions don't have a `rule` property, it only shows empty data in rule rows, and it doesn't break anything. My question is, do we have a design for such situation? What needs to be showed in this case to indicate that there are more levels of conditional permission policies? And is it within the scope of this story or out of this story?
cc. @ShiranHi 
image

ShiranHi:
@ciiay looks good, thank you! Few comments:
1. Can we add the nested condition tooltip ([Figma](https://www.figma.com/design/E6C9X3OSVzlWkW7ZsbxN7s/Administration?node-id=4580-5377&t=0FTIezImwDPJ38SY-1))?

2. Can we increase the padding between the "Add rule" button and the field above it? it seems too small
image01

3. This is really small but I think it will be better if the border of nested condition will be aligned with the beginning of the "Add nested condition" button ([Figma](https://www.figma.com/design/E6C9X3OSVzlWkW7ZsbxN7s/Administration?node-id=5054-4645&t=0FTIezImwDPJ38SY-1))
image02

4. Can we show this error only after users click on "Add rule" or on the side drawer "Save" button?
image03

5. The bottom buttons of the side drawer "Save" and "Cancel" should be sticky, I didn't see that from the video so I write this to confirm we have it
  
#My question is, do we have a design for such situation? What needs to be showed in this case to indicate that there are more levels of conditional permission policies? And is it within the scope of this story or out of this story?

We agreed to limit ourselves to just one level of nested conditions for now, despite the backend's capability to support multiple levels. In case users use the CLI to create multiple levels we will show them one below each other, so I think the nested condition fields should not be empty. Does this make sense?

ciiay:
Hi @ShiranHi , thanks for the review. Here's updated screenshot.
image

image

About the form validation, it has consistent behavior as Plugin drop list and Resource type input field, which is from the form component we implemented with. I don't see a very straight forward solution here. @divyanshiGupta Any thoughts?
image

>In case users use the CLI to create multiple levels we will show them one below each other, so I think the nested condition fields should not be empty

Do you mean we don't allow them to edit the multiple level nested conditions which is more than 1 level, but we need to display the nested condition? It's empty because the nested condition in a nested condition doesn't have a `rule` field to display in the `Rule` input field. I can filter the simple conditions in a nested condition which have a `Rule` field, but not sure what the design is for the nested conditions. Hopefully this screenshot explains what I mean:
image

The mocked condition I used is
```
      conditions: {
        anyOf: [
          {
            rule: 'IS_ENTITY_OWNER',
            resourceType: 'catalog-entity',
            params: {
              claims: ['user:default/ciiay'],
            },
          },
          {
            rule: 'IS_ENTITY_KIND',
            resourceType: 'catalog-entity',
            params: { kinds: ['Group'] },
          },
          {
            allOf: [
              {
                rule: 'IS_ENTITY_OWNER',
                resourceType: 'catalog-entity',
                params: {
                  claims: ['user:default/ciiay'],
                },
              },
              {
                rule: 'IS_ENTITY_KIND',
                resourceType: 'catalog-entity',
                params: {
                  kinds: ['User'],
                },
              },
              {
                not: {  // this is 2nd level nested condition
                  rule: 'HAS_LABEL',
                  resourceType: 'catalog-entity',
                  params: { label: 'temp' },
                },
              },
              {
                anyOf: [  // this is another 2nd level nested condition
                  {
                    rule: 'HAS_TAG',
                    resourceType: 'catalog-entity',
                    params: { tag: 'dev' },
                  },
                  {
                    rule: 'HAS_TAG',
                    resourceType: 'catalog-entity',
                    params: { tag: 'test' },
                  },
                ],
              },
            ],
          },
        ],
      },
```

I'm thinking maybe we could give some message like this
image

ShiranHi:
> About the form validation, it has consistent behavior as Plugin drop list and Resource type input field, which is from the form component we implemented with. I don't see a very straight forward solution here. @divyanshiGupta Any thoughts?
> image

I believe we should display error messages only after users have left the field empty, rather than immediately. If this behavior appears in other areas, I recommend making the same adjustment.

> Do you mean we don't allow them to edit the multiple level nested conditions which is more than 1 level, but we need to display the nested condition? It's empty because the nested condition in a nested condition doesn't have a rule field to display in the Rule input field. I can filter the simple conditions in a nested condition which have a Rule field, but not sure what the design is for the nested conditions.

So, in case there are multiple level nested conditions, I suggest showing only the first level as in the screenshot and show a message like [this one](https://www.figma.com/design/E6C9X3OSVzlWkW7ZsbxN7s/Administration?node-id=4580-5700&t=IG3W0nV5wYCmgC5V-1).

If we can provide a link explaining how to use the CLI for this, we can add it to the yellow notification.
And let's remove the red text "Detailed condition information is hard to display..."

ShiranHi:
> I have uploaded the updated screen recording along with the screenshot of the multiple-level nested condition sidebar display.

Thank you!

> Since this PR focuses on reusing the existing form component to implement the nested condition logic, we can track this improvement in a different ticket. What do you think?

Sure, it would be great to change that everywhere.

> That was only a note on the screenshot to explain the situation, not in the code 😄

Good to know haha, thanks for clarifying it!

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
```diff
@@ -24,26 +34,65 @@ const useStyles = makeStyles(theme => ({
       backgroundColor: `${theme.palette.background.paper}!important`,
     },
   },
+  nestedConditionRow: {
+    padding: '20px',
+    marginLeft: '1.5rem',
```
divyanshiGupta, line 39: Can this be replaced with a spacing value from the theme?

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
```diff
@@ -24,26 +34,65 @@ const useStyles = makeStyles(theme => ({
...
     alignSelf: 'baseline',
     marginTop: theme.spacing(3.3),
   },
+  removeNestedRuleButton: {
+    color: theme.palette.grey[500],
+    flexGrow: 0,
+    alignSelf: 'baseline',
+  },
+  radioGroup: {
+    margin: '0.5rem',
```
divyanshiGupta, line 91: Can this be replaced with a spacing value from the theme?

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
```diff
@@ -24,26 +34,65 @@ const useStyles = makeStyles(theme => ({
...
+  },
+  radioLabel: {
+    marginTop: '0.5rem',
```
divyanshiGupta, line 94: Can this be replaced with a spacing value from the theme?

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
divyanshiGupta, line 1: @ciiay the size of the file has become huge so it makes sense to move out independent pieces of code for example some utility functions. You can create a new file in utils folder for this component and move out utils that are not directly dependent on this component.

divyanshiGupta (changes requested):

plugins/rbac/src/components/ConditionalAccess/AddNestedConditionButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Tooltip from '@material-ui/core/Tooltip';
+import HelpOutlineIcon from '@material-ui/icons/HelpOutline';
+
+export const AddNestedConditionButton = () => {
+  const tooltipTitle = () => (
+    <div>
+      <p style={{ textAlign: 'center' }}>
+        Nested conditions are <b>1 layer rules within a main condition</b>. It
+        lets you allow appropriate access by using detailed permissions based on
+        various conditions. You can add multiple nested conditions.
```
divyanshiGupta, line 13: IMO nested condition definition can be improved. We should explain what a nested condition is and then add that the UI currently supports only 1 level of nesting i.e multiple criteria based conditions can only be added to the upper most criteria.

@AndrienkoAleksandr @parvathyvr any suggestions on the definition?

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
```diff
@@ -2,17 +2,36 @@ import React from 'react';
...
-import { ConditionsData, RuleParamsErrors, RulesData } from './types';
+import { CriteriaToggleButton } from './CriteriaToggleButton';
+import {
+  Condition,
+  ConditionFormRowProps,
+  ConditionsData,
+  NotConditionType,
+} from './types';
 
 const useStyles = makeStyles(theme => ({
```
divyanshiGupta, line 36: @ciiay since the styles have increased we can also think about moving it to a separate file.

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
```diff
@@ -24,108 +43,540 @@ const useStyles = makeStyles(theme => ({
...
+      setNotConditionType('simple-condition');
+    }
+
+    const ruleChange = ruleMap[val];
+    if (ruleChange) {
+      onRuleChange(ruleChange);
+    }
+  };
+
+  const updateRules = (updatedNestedConditionRow: Condition[] | Condition) => {
```
divyanshiGupta, line 171: This can be moved out

plugins/rbac/src/components/ConditionalAccess/AddNestedConditionButton.tsx
ShiranHi: > IMO nested condition definition can be improved. We should explain what a nested condition is and then add that the UI currently supports only 1 level of nesting i.e multiple criteria based conditions can only be added to the upper most criteria.
> 
> @AndrienkoAleksandr @parvathyvr any suggestions on the definition?

I find this text fairly clear in explaining what a nested condition is and discussing your suggestion. It mentions the limitation of having only 1 layer and how it has been incorporated into the main condition. I'm unsure about what you feel is still unclear or missing. Could you please explain what is missing/unclear?

divyanshiGupta: `Nested conditions are <b>1 layer rules within a main condition`
@ShiranHi I find above line confusing as nested conditions can have multiple levels. It looks like we are stating that in general nested conditions can have only 1 level. But its just a suggestion, if you think it looks correct then I am fine with it :)

divyanshiGupta: cc: @invincibleJai

ShiranHi: > I find above line confusing as nested conditions can have multiple levels. It looks like we are stating that in general nested conditions can have only 1 level whereas its the limitation of the UI we are providing.

Got it, I see your point. I think that if we allow for one level of nested conditions on the UI, this is the story we should communicate to users, unless it will be confusing. I've tried some other wording but it feels that this sentence is the best option for now. Unless you have a better suggestion, let's stick with this version, as it was suggested by our UX copywriter.

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
ciiay: @divyanshiGupta This function uses `criteria`, `conditionRow` and updates form using `onRuleChange` all from `ConditionsFormRow`. It makes more sense to leave it in its context to me. How do you think?

divyanshiGupta:
@ciiay Please rebase your PR and resolve the conflicts.

divyanshiGupta:
@ciiay I see there are a lot of issues reported by sonarcloud. Please try to fix these.

divyanshiGupta:
https://github.com/user-attachments/assets/57d998fe-35fd-44a7-ac29-ffcad9430aeb

@ciiay If I remove any simple rule row, entire form data is getting removed.

divyanshiGupta:
https://github.com/user-attachments/assets/3bccc5a6-dab1-4bfd-a9ca-491c604b02db

@ciiay I configured 5 rules but I see its saying 4 rules instead of 5.

divyanshiGupta:
https://github.com/user-attachments/assets/574913f7-28d5-42ab-8708-18f822adb94e

@ciiay since allOf/anyOf is an array of PermissionCriteria<PermissionCondition> the user should be able to create a nested condition without having to add a simple rule. But I noticed that in the current flow the user cannot add a nested condition without a simple rule.

PatAKnight:
cc @AndrienkoAleksandr

plugins/rbac/src/utils/conditional-access-utils.ts
```diff
@@ -0,0 +1,129 @@
...
+export const calculateConditionIndex = (
+  conditionRow: ConditionsData,
+  nestedConditionIndex: number,
+  criteria: string,
+): number => {
+  const simpleRulesCount =
+    criteria === criterias.not
+      ? 0
+      : (conditionRow[criteria as keyof Condition] as Condition[]).filter(
+          (e: Condition) => 'rule' in e,
```
AndrienkoAleksandr, line 125: @ciiay, maybe: 
```suggestion
          (e: Condition) => e && 'rule' in e,
```

because, I faced with bug here: 

https://github.com/user-attachments/assets/aff4268c-79fd-4a4c-ac11-736cdacf143a

AndrienkoAleksandr: Possible workaround is "?? []": : (conditionRow[criteria as keyof Condition] as Condition[] ?? []).filter(

plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
divyanshiGupta, line 1: @ciiay this file needs refactor as I see it has more than 800 lines of code now. This will cause following issues in future:

- Difficut to review and increases chances of introducing bugs

-  Difficult to maintain and add new code to

- Bug prone and if a bug occurs in this file it will be difficult to identify and fix as the file is super huge and the code does not look streamlined

Easiest way to reduce the no. of line in this file is to convert it into a wrapper component `ConditionsFormRowWrapper.tsx` and move out the render part into a new file and call it `ConditionsFormRow.tsx`. There are other ways also we can reduce the no. of lines in this file. We should always look for how we can create small isolated code blocks that can be separated out into multiple files/components so that one file/component does not get super huge. LMK if you are comfortable in doing that in this PR or a separate refactor PR.

plugins/rbac/src/components/ConditionalAccess/const.ts
```diff
@@ -18,3 +20,6 @@ export const conditionButtons = [
   { val: criterias.anyOf, label: criteriasLabels[criterias.anyOf] },
   { val: criterias.not, label: criteriasLabels[criterias.not] },
 ];
+
+export const rbacAPIDocLink =
+  'https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#ref-rbac-rest-api-endpoints_assembly-rhdh-integration-aks';
```
invincibleJai, line 25: this link has a version hardcoded in it so with each release this would need an update and it doesn't take me to RBAC doc as well , if we just need to point to product doc then can use https://docs.redhat.com/en/documentation/red_hat_developer_hub

ciiay: @ShiranHi  @invincibleJai , this link is for the CLI keyword in the message "Multiple levels of nested conditions are not supported. Only one level is displayed. Please use the [CLI](url) to view all nested conditions". If it's just a generic link to the product doc, do we still need to put a link? If we want to point to a more specific CLI related page, it will have version in the URL which is hardcoded.

ShiranHi: @ciiay do we cover this in the documentation? I couldn't find it. If we do, we can use the link Jai mentioned. If not, I suppose we can just remove it since users initially configured the nested condition using the CLI, so they should know where to go back and check. Is that correct?

ciiay: @ShiranHi Agreed. In the documentation, it mentions "nested condition" in an example, but nowhere mentions the "multiple levels of nested condition are not supported" limitation. So we can just simply remove it as you suggested.

divyanshiGupta (approved):
Thanks for the PR @ciiay. Tested it locally and create flow is working as expected. Found some issues in the edit flow but these can be discussed and fixed in your edit flow PR.

/approve

AndrienkoAleksandr (commented):
/lgtm

AndrienkoAleksandr (approved):

PR #1818, opened 2024-06-17
feat(orchestrator): disable buttons based on permissions

gabriel-farache:
In the Orchestrator plugin UI, a given user shall only see the workflows it is allowed to see/execute

The runs of the workflow shall also be hidden to the not allowed users


gabriel-farache:
@rgolangh @batzionb

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -82,8 +82,10 @@ const authorize = async (
   return decision;
 };
 
-declare class UnauthorizedError extends NotAllowedError {
-  message: 'Unauthorized';
+export class UnauthorizedError extends NotAllowedError {
```
rgolangh, line 85: use the the rbac-common Unauthrorized one

plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -204,7 +209,7 @@ export const WorkflowInstancePage = ({
                     <Button
                       variant="contained"
                       color="primary"
-                      disabled={!canAbort}
+                      disabled={!permittedToExecute.allowed || !canAbort}
```
rgolangh, line 212: we should probably use the orchestratorWorkflowInstanceAbortPermission and same for line 222

plugins/orchestrator/src/components/WorkflowDefinitionViewerPage/WorkflowDefinitionViewerPage.tsx
```diff
@@ -61,13 +67,16 @@ export const WorkflowDefinitionViewerPage = () => {
...
-                onClick={handleExecute}
-              >
-                Run
-              </Button>
+              !loadingPermission && (
+                <Button
+                  variant="contained"
+                  color="primary"
+                  onClick={handleExecute}
+                  disabled={!canRun}
```
batzionb, line 75: while it's loading it's pretty easy to add a spinner to the button 
see the SubmitButton used in the Execute page for reference
https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator/src/components/SubmitButton.tsx#L31

batzionb:
The user might not understand why he is not allowed to click the buttons
In other applications, such as assisted installer, we added a tooltip explaining why the button is disabled
Perhaps we can add a tooltip "Unauthorized" when the buttons are disabled
WDYT?

plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -201,39 +212,54 @@ export const WorkflowInstancePage = ({
...
+                      title="user not authorized to execute workflow"
+                      disableHoverListener={permittedToExecute.allowed}
                     >
-                      Retrigger
-                    </Button>
+                      <Button
+                        variant="contained"
+                        color="primary"
+                        disabled={!permittedToExecute.allowed || !canRerun}
+                        onClick={canRerun ? handleRerun : undefined}
```
gabriel-farache, line 223: not sure about the permission change, but for me, rerun == execution, not abort

batzionb: makes sense

rgolangh: I confused it with something I did in the permission check. Anyhow it is closer to execute, right, but a bit more than that -  a rerun is a call to get the parameters of a certain execution and then an execute with that. 
+1

plugins/orchestrator/src/components/WorkflowDefinitionViewerPage/WorkflowDefinitionViewerPage.tsx
```diff
@@ -61,13 +68,21 @@ export const WorkflowDefinitionViewerPage = () => {
             {loading ? (
```
batzionb, line 68:
```suggestion
            {loading || loadingPermission ? (
```

rgolangh:
```
test: manifests generate envtest vet fmt test-api ## Run tests.
```

plugins/orchestrator/src/components/WorkflowDefinitionViewerPage/WorkflowDefinitionViewerPage.tsx
```diff
@@ -65,7 +65,7 @@ export const WorkflowDefinitionViewerPage = () => {
         )}
         <Grid container item justifyContent="flex-end" spacing={1}>
           <Grid item>
-            {loading ? (
+            {loading || loadingPermission ? (
               <Skeleton variant="text" width="5rem" />
             ) : (
               !loadingPermission && (
```
batzionb, line 71: you can remove !loadingPermission from here

plugins/orchestrator/src/components/WorkflowsTable.tsx
```diff
@@ -130,12 +139,14 @@ export const WorkflowsTable = ({ items }: WorkflowsTableProps) => {
...
 
   return (
-    <OverrideBackstageTable<FormattedWorkflowOverview>
-      title="Workflows"
-      options={options}
-      columns={columns}
-      data={data}
-      actions={actions}
-    />
+    permittedToReadWorkflows && (
```
batzionb, line 142: I think we need some kind of message if the user is not permitted to read the workflows otherwise it won't be clear what's going on

batzionb: I think we can just leave this as is
If the user can't access he'll see an error message on top, it's the same effect only the user will know what's going on

gabriel-farache: Yeah, agree, the big "403" error in red should be enough

batzionb:
@gabriel-farache 
please remove the changes in yarn.lock, they are added after running "yarn install" with the diff, but do not need to go in
yes, I know it's a pain :(

gabriel-farache:
@batzionb But if I remove the changes, then I have the same things as for #1820 (https://github.com/janus-idp/backstage-plugins/actions/runs/9599728928/job/26478472007?pr=1820) :
```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

plugins/orchestrator/src/components/WorkflowsTable.tsx
batzionb: When I said leave it as is I meant without the change in this PR
With the change in this PR there won't be any error message and nothing will be seen

gabriel-farache: ah, I see what you mean, I'll remote the condition

batzionb:
@gciavarrini FYI for v2

batzionb:
@gabriel-farache 

what's left is to remove the global yarn.lock

and include plugins/orchestrator-backend/dist-dynamic/yarn.lock in the PR

gabriel-farache:
> what's left is to remove the global yarn.lock

and include plugins/orchestrator-backend/dist-dynamic/yarn.lock in the PR

@batzionb ok, now I feel stupid :) 

Done

batzionb (approved):

PR #1822, opened 2024-06-19
fix(matomo): add default export for new backend system

yashoswalyo:
- add default export for new backend system
- remove material-ui deps
- add `configSchema` field to package.json


riginoommen (approved):

deshmukhmayur (approved):
LGTM.

PR #1823, opened 2024-06-19
fix(orchestrator): change log level of cache messages to be debug

batzionb:
The cache messages were generating a large volume of logs, which was overwhelming and obscuring other important log messages. This made it difficult to debug issues effectively. By changing the log level to debug, we can reduce noise in the logs and make it easier to identify and address other issues.


batzionb:
created from wrong base branch

PR #1825, opened 2024-06-19
feat(keycloak): expose keycloak user/group transformer extension point

JessicaJHee:
Resolves [RHIDP-2644](https://issues.redhat.com/browse/RHIDP-2644): Expose extension points for the keycloak-backend plugin

To test this change locally, [this](https://github.com/JessicaJHee/backstage-showcase/tree/testing-keycloak-transformers) may be useful - it adds a new module that uses the extension point to append `_foo` and `_bar` to the group and user name respectively.

### DIFF
--- a/plugins/keycloak-backend/README.md
+++ b/plugins/keycloak-backend/README.md
@@ -243,11 +243,60 @@ yarn workspace backend add @janus-idp/backstage-plugin-keycloak-backend
    backend.start();
    ```
 
----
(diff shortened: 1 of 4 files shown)


kim-tsao:
I was going to ask where the [unit tests](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/src/lib/read.test.ts) were but it looks like they were added before. 

/lgtm

nickboldt:
I'm not clear in 1.2.x branch of showcase where this change was picked up https://github.com/janus-idp/backstage-showcase/commits/1.2.x/ 

please clarify if this is being cherrypicked to 1.2.2 or only for 1.3.

PR #1831, opened 2024-06-20
fix(rbac): simplify db logic

AndrienkoAleksandr:

Drop policy source database table, move this data to the cabin db table.

https://issues.redhat.com/browse/RHIDP-1528


plugins/rbac-backend/src/service/enforcer-delegate.ts
```diff
@@ -70,14 +60,11 @@ export class EnforcerDelegate {
     const trx = externalTrx ?? (await this.knex.transaction());
 
     try {
-      await this.policyMetadataStorage.createPolicyMetadata(
-        source,
-        policy,
-        trx,
-      );
-      const ok = await this.enforcer.addPolicy(...policy);
+      const ok = await this.enforcer.addPolicy(...policy, source);
```
PatAKnight, line 63: I have been thinking about this some more. Do we need to store the source information any more with the policies if we are just going to reference the role metadata for the source?

PatAKnight (commented):
Looking pretty good so far, had a question that I wanted to get your opinion on.

AndrienkoAleksandr:
> I have been thinking about this some more. Do we need to store the source information any more with the policies if we are just going to reference the role metadata for the source?

I also thought about that, which is why I spent two days preparing an alternative PR: https://github.com/janus-idp/backstage-plugins/pull/1842 on top of the current changes. Please take a look.

AndrienkoAleksandr:
Close on favour: https://github.com/janus-idp/backstage-plugins/pull/1842

PR #1833, opened 2024-06-21
feat(orchestrator): remove unneeded orchestrator jira integration and endpoint

anludke:
The jira integration and /webhook/jira endpoint are no longer necessary. These were temporary solutions used during the poc and initial work on the orchestrator. The orchestrator workflows, including escalation workflow, support jira integration where needed, making the previous workaround obsolete.


gciavarrini (approved):
lgtm

PR #1835, opened 2024-06-21
feat(kiali): include Kiali external URL as a parameter

josunect:
```
    kiali:
      # Required. Kiali endpoint
      url: ${KIALI_BASE_URL}
      # Optional. Kiali public URL to redirect to standalone Kiali. When not specified, url will be used.
      urlExternal: ''
```

This URL is also displayed in the about window, it can be used to verify: 

image

- Added some test coverage for resources card.


aljesusg (approved):
LGFM

PR #1838, opened 2024-06-25
Update package.json - bump to new version in main to align with downstream versioning (3.3 ~= 1.3)

nickboldt:
…tream versioning (3.3 ~= 1.3)


nickboldt:
note: in future this should be done automatically -- see https://issues.redhat.com/browse/RHIDP-931

PR #1842, opened 2024-06-26
fix(rbac): simplify db logic

AndrienkoAleksandr:

Drop policy source.

https://issues.redhat.com/browse/RHIDP-1528


plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
```diff
@@ -399,10 +378,7 @@ export class CSVFileWatcher {
           roleMetadata.roleEntityRef,
         );
 
-        await this.enforcer.addOrUpdateGroupingPolicy(
-          groupPolicy,
-          roleMetadata,
-        );
+        await this.enforcer.addGroupingPolicy(groupPolicy, roleMetadata);
```
PatAKnight, line 381: It looks like on start up, legacy roles that have a matching role in the CSV file are not updated until a change is made to that role.

plugins/rbac-backend/src/service/enforcer-delegate.ts
```diff
@@ -92,18 +77,24 @@ export class EnforcerDelegate {
...
-    source: Source,
     externalTrx?: Knex.Transaction,
   ): Promise<void> {
     const trx = externalTrx || (await this.knex.transaction());
+
+    const policiesToAdd: string[][] = await policies.reduce<
+      Promise<string[][]>
+    >(async (accP: Promise<string[][]>, p: string[]) => {
+      const acc = await accP;
+      const hasPolicy = await this.enforcer.hasPolicy(...p);
```
PatAKnight, line 88: Noticed we are duplicating the `hasPolicy` check a few times. We have it here, in the `processPolicies` right [before](https://github.com/janus-idp/backstage-plugins/blob/0d7ab0691878579c35b3d75a2cbd3bc1a5919f3b/plugins/rbac-backend/src/service/policies-rest-api.ts#L298C44-L298C59) the `addPolicies` call in `policies-rest-api.ts`, and again we are calling `processPolicies` [before](https://github.com/janus-idp/backstage-plugins/blob/0d7ab0691878579c35b3d75a2cbd3bc1a5919f3b/plugins/rbac-backend/src/service/policies-rest-api.ts#L378) the `updatePolicies` call that will end up calling `addPolicies`.

I think we should find a way to simplify it so that we are not performing duplicate checks.

plugins/rbac-backend/src/service/enforcer-delegate.ts
```diff
@@ -179,15 +169,19 @@ export class EnforcerDelegate {
   ): Promise<void> {
     const trx = externalTrx ?? (await this.knex.transaction());
 
+    const policiesToAdd: string[][] = await policies.reduce<
+      Promise<string[][]>
+    >(async (accP: Promise<string[][]>, gp: string[]) => {
+      const acc = await accP;
+      const hasPolicy = await this.enforcer.hasGroupingPolicy(...gp);
```
PatAKnight, line 176: Same as the comment about `addPolicies`, we have multiple repeated checks of `hasGroupingPolicy` leading up to `addGroupingPolicies`. In `policies-rest-api.ts` we have the same call [here](https://github.com/janus-idp/backstage-plugins/blob/0d7ab0691878579c35b3d75a2cbd3bc1a5919f3b/plugins/rbac-backend/src/service/policies-rest-api.ts#L502) before `addGroupingPolicies`

PatAKnight (commented):
Looking good so far, I have testing this with making changes with the UI and CSV file and can confirm

From UI -> unable to create / edit / delete policies and roles that originate from config file and csv file

From CSV -> unable to create / edit / delete policies and roles that originate from rest api and config file

Still need to finish testing legacy policies and roles and when using rest api. But wanted to at least start a review.

PatAKnight (commented):
Went through and tested some of the legacy options:

Legacy updates to config if there is a match
Legacy updates to CSV if there is a match on start up
Legacy updates to REST if there is a match during PUT request

The only thing that does not happen is if I add a role to the CSV file that matches a legacy role in the database. In this scenario, the source stays as legacy.

Also, `yarn tsc` is failing because `CSV_PERMISSION_FILE_AUTHOR` is declared in the `csv-file-watcher.test.ts` but never used.

plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
AndrienkoAleksandr: I think I handled this comment.

AndrienkoAleksandr:
> The only thing that does not happen is if I add a role to the CSV file that matches a legacy role in the database. In this scenario, the source stays as legacy.
> Also, yarn tsc is failing because CSV_PERMISSION_FILE_AUTHOR is declared in the csv-file-watcher.test.ts but never used.

Fixed.

PatAKnight (approved):
/LGTM

PR #1868, opened 2024-07-05
feat(rbac): edit nested conditions

ciiay:
For [RHIDP-2554](https://issues.redhat.com/browse/RHIDP-2554)

Screen recording:

https://github.com/user-attachments/assets/938034c4-fad5-481c-98c4-4e023a637920

### DIFF
--- a/plugins/rbac/src/__fixtures__/mockConditions.ts
+++ b/plugins/rbac/src/__fixtures__/mockConditions.ts
@@ -111,6 +111,26 @@ export const mockConditions: RoleConditionalPolicyDecision<PermissionAction>[] =
               },
             ],
           },
+          {
+            not: {
+              allOf: [
+                {
+                  rule: 'IS_ENTITY_OWNER',
+                  resourceType: 'catalog-entity',
+                  params: {
+                    claims: ['user:default/xyz'],
+                  },
+                },
+                {
+                  rule: 'IS_ENTITY_KIND',
+                  resourceType: 'catalog-entity',
+                  params: {
+                    kinds: ['User'],
+                  },
+                },
+              ],
+            },
+          },
         ],
       },
       roleEntityRef: 'role:default/rbac_admin',
--- a/plugins/rbac/src/components/ConditionalAccess/ConditionsForm.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/ConditionsForm.tsx
@@ -178,18 +178,25 @@ export const ConditionsForm = ({
         const nestedConditionCriteria = Object.keys(
           firstLevelNestedCondition,
         )[0];
-        return (
+        if (
           Array.isArray(
             firstLevelNestedCondition[
               nestedConditionCriteria as keyof Condition
             ],
-          ) &&
-          (
+          )
+        ) {
+          return (
             firstLevelNestedCondition[
               nestedConditionCriteria as keyof Condition
             ] as Condition[]
-          ).some((con: Condition) => !('rule' in con))
-        );
+          ).some((con: Condition) => !('rule' in con));
+        }
+
+        return !Object.keys(
+          firstLevelNestedCondition[
+            nestedConditionCriteria as keyof Condition
+          ] as Condition[],
+        ).includes('rule');
       });
   };
 
--- a/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
+++ b/plugins/rbac/src/components/ConditionalAccess/ConditionsFormRow.tsx
@@ -436,32 +436,34 @@ export const ConditionsFormRow = ({
                 }
               />
             ))}
-          {selectedNestedConditionCriteria === criterias.not && (
-            <ConditionsFormRowFields
-              oldCondition={
-                (nc as ConditionsData).not ??
-                getDefaultRule(selPluginResourceType)
-              }
-              onRuleChange={onRuleChange}
-              conditionRow={conditionRow}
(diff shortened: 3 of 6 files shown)


divyanshiGupta:
@ciiay I tried created the following multi-level nested condition using cli:
```
curl -X POST "http://localhost:7007/api/permission/roles/conditions" -d '{"result":"CONDITIONAL","roleEntityRef": "role:default/dev","pluginId":"catalog","resourceType":"catalog-entity","permissionMapping": ["update"],"conditions": {
     "allOf":[
      {
         "rule":"IS_ENTITY_OWNER",
         "resourceType":"catalog-entity",
         "params":{
            "claims":[
               "user:default/divyanshiGupta"
            ]
         }
      },
      {
        "allOf":[
          {
            "rule":"IS_ENTITY_OWNER",
            "resourceType":"catalog-entity",
            "params":{
                "claims":[
                  "user:default/xyz2"
                ]
            }
          },
          {
            "rule":"IS_ENTITY_KIND",
            "resourceType":"catalog-entity",
            "params":{
                "kinds":[
                  "Group"
                ]
            }
          }
        ]
      },
      {
         "anyOf":[
            {
               "rule":"IS_ENTITY_OWNER",
               "resourceType":"catalog-entity",
               "params":{
                  "claims":[
                     "user:default/divs258"
                  ]
               }
            },
            {
               "rule":"IS_ENTITY_KIND",
               "resourceType":"catalog-entity",
               "params":{
                  "kinds":[
                     "User"
                  ]
               }
            }
         ]
      },
      {
         "not":{
            "allOf":[
              {
                "rule":"IS_ENTITY_OWNER",
                "resourceType":"catalog-entity",
                "params":{
                    "claims":[
                      "user:default/xyz"
                    ]
                }
              },
              {
                "rule":"IS_ENTITY_KIND",
                "resourceType":"catalog-entity",
                "params":{
                    "kinds":[
                      "User"
                    ]
                }
              }
           ] 
          }
      }
    ]
  }
}' -H "Content-Type: application/json" -H "Authorization: Bearer token" -v
```
Above `not` has a multui-level nesting but instead of seeing the warning alert I see an empty `not` row.

divyanshiGupta (changes requested):
@ciiay Overall the PR looks good and works as expected. I just found a small issue.
Now I can see the alert for the multi-level nested 'not' criteria but I also see a row without the fields. It happens when there is multi-level nested conditions inisde a criteria without any simple rules. When there are simple rules we have to show the criteria row but when there are no simple rules and just nested conditions inside the criteria I am not sure showing the criteria row without the fields or option to add more rules is good user experience. If we are showing the row then we should also gove `Add rule` option so that the user can add more simple rules if they want.

Not row 
Screenshot 2024-08-09 at 5 44 14 PM

Allof row 
Screenshot 2024-08-09 at 5 58 23 PM

divyanshiGupta:
Noticed another issue, wherein I tried to update the actions but instead of getting updated I got an error.


https://github.com/user-attachments/assets/2e81c8d9-f0e7-46a9-ae1f-21515d4bc154

ciiay:
Hi @divyanshiGupta @ShiranHi 
>Now I can see the alert for the multi-level nested 'not' criteria but I also see a row without the fields. 

I actually left the empty nested condition block on purpose so that it's clear where is the hidden second level nested condition. If the user wants to add/edit the empty nested condition block, they can just toggle criteria to get an empty input row. Let me know if I need to do any change on this behavior.

ciiay:
Hi @divyanshiGupta 
> Noticed another issue, wherein I tried to update the actions but instead of getting updated I got an error.

Yeah, I've seen this error before but I thought I put conflicted permissions which caused the error... This should be a bug unrelated to nested condition, check out this test flow with only simple rules. We need to either restrict the order of the update or put a better warning message to let the user know what went wrong and how to fix it with the right order to update the existing permissions.
 
https://github.com/user-attachments/assets/06150634-7e11-4f19-b70e-ba2ef54eafc2

ShiranHi:
> When there are simple rules along with nested condition inside a criteria then we have to show the criteria row with just the simple rules but when there are no simple rules and just nested conditions inside the criteria I am not sure showing the criteria row without the fields or option to add more rules is good user experience. If we are showing the row then we should also give `Add rule` option so that the user can add more simple rules if they want. Please confirm this with @ShiranHi.

+1 on this. Just displaying empty rows doesn't seem logical to me. I'm fine with simply hiding it and presenting the error message in this case. I'm curious to see how many users will encounter this situation, but I don't want to over-engineer it. If we'll find that this happens frequently, we can come up with a better UI solution later. For now, let's just hide it.

divyanshiGupta:
> Noticed another issue, wherein I tried to update the actions but instead of getting updated I got an error.
> 
> 
> 
> 
> 
> https://github.com/user-attachments/assets/2e81c8d9-f0e7-46a9-ae1f-21515d4bc154
> 
> 
> 
> Looks like this issue might be in main as well @ciiay can you confirm. Looks like its happening because it tries to update one policy and validates it against policies which are meant to be updated but not yet updated and hence throws the error. Maybe backend should do a dry run to update all the incoming policies first and then validate to see if the respective actions are already being used or not. If action is still being used then throw error otherwise continue with updation. @AndrienkoAleksandr wdyt?
> 
> 
> 
> 

@AndrienkoAleksandr did you get a chance to look at this?

divyanshiGupta (approved):
/lgtm

PR #1880, opened 2024-07-11
fix(orchestrator): returned scrolling bars to instance page cards

batzionb:
Following upgrade of material UI packages, styling needed to move from InfoCard to CardContent
resolves [FLPATH-1480](https://issues.redhat.com/browse/FLPATH-1480)


gciavarrini (approved):
lgtm

PR #1891, opened 2024-07-15
feat(deps): update to backstage 1.28

schultzp2020:
fixes https://issues.redhat.com/browse/RHIDP-2844


karthikjeeyar (approved):
Verified tekton, quay and argocd plugins. 

Quay plugin uses the proxy backend and all the GET requests is attached the backstage user credentials, so we are not impacted by the breaking changes that backstage 1.28 brings in wrt to proxy backend.



/lgtm

nickboldt:
note: this change has caused https://issues.redhat.com/browse/RHIDP-3236 because BS added new required metadata backstage.pluginId field in their CLI, which prevents a MSR plugin release.

PR #1904, opened 2024-07-18
docs: fix argocd naming

jerolimov:
nit: Argocd is spelled Argo CD


karthikjeeyar:
@jerolimov In Readme file there are multiple instances of the name, should we update that too ?

jerolimov:
> @jerolimov In Readme file there are multiple instances of the name, should we update that too ?

Fixed some other places as well. PTAL.

PR #1906, opened 2024-07-18
fix(orchestrator): resolve broken dynamic plugin publish

batzionb:
Issue was conflicting dependencies that caused dist-dynamic to contain @backstage dependencies that should be taken from host application.
Moving @janus-idp dependencies to peerDependencies resolved the issue.


gciavarrini (approved):
lgtm

PR #1908, opened 2024-07-18
fix(rbac): add additional validation for permission policies

### LINKED ISSUE #996: [RBAC plugin] Add additional validation for the permission policies
### What do you want to improve?

Add additional validation for the permission policies

### What is the current behavior?

Permission policies within the RBAC plugin have the ability to set the policy section to anything. 
Permission policy format: { entityReference, permission, policy, effect }
```
p, role:default/test, catalog-entity, write, allow
p, role:default/test, catalog.entity.create, execute, allow
```

This allows for permissions to be created that are not actually valid (write and execute) for the RBAC plugin and the Backstage permission framework

### What will the new behavior be?

Add validation to the policy section of permission policies. The only allowed terms that can be used are `create`, `read`, `update`, `delete`, and our RBAC specific keyword `use`.

dzemanov:
Adds validation to the policy section of permission policies. The only allowed terms that can be used are create, read, update, delete, and our RBAC specific keyword use.

Closes #996.

### DIFF
--- a/plugins/rbac-backend/src/service/policies-rest-api.test.ts
+++ b/plugins/rbac-backend/src/service/policies-rest-api.test.ts
@@ -1240,7 +1240,7 @@ describe('REST policies api', () => {
               effect: 'allow',
             },
           ],
-          newPolicy: [{ permission: 'policy-entity', policy: 'write' }],
+          newPolicy: [{ permission: 'policy-entity', policy: 'create' }],
         });
 
       expect(result.statusCode).toEqual(400);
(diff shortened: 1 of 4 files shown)


AndrienkoAleksandr (commented):
I tested this pr - works like expected. And I'm fine with current pr code. There is only one small minus - we have duplication permission policy action values. @dzemanov I can propose small improvements to fix that: https://github.com/janus-idp/backstage-plugins/compare/rbac-policy-validation...cloud-eda:backstage-plugins:rbac-policy-valivation-alternative?expand=1 . But it is optional and up to you and these changes require one more release of the common package. @PatAKnight what do you think?

dzemanov:
@AndrienkoAleksandr yes, was thinking about this as well. If it is ok to release new version, I would go with your proposed changes.

AndrienkoAleksandr (approved):

PR #1913, opened 2024-07-19
docs(rbac): add curl request examples

dzemanov:
Adds example cURL command for nested conditional policies.


AndrienkoAleksandr (commented):
Looks good to me. @PatAKnight, what do you think? Perhaps we should include an additional curl command example for creating a role at the beginning of the chapter?

PatAKnight:
@AndrienkoAleksandr yes, I agree that an example for adding a role would be nice. Do you mind adding the example @dzemanov?

dzemanov:
Sure!

AndrienkoAleksandr (approved):

PR #1923, opened 2024-07-23
fix(rbac): don't start transaction if there no group policies

AndrienkoAleksandr:

Fix regression after: https://github.com/janus-idp/backstage-plugins/pull/1842


dzemanov:
Tried locally, fixes the `KnexTimeoutError`.

PatAKnight (approved):
/lgtm

PR #1983, opened 2024-07-30
fix(orchestrator): remove default pagination on v2 endpoints

gciavarrini:
If the user doesn't set pagination info, then retrieve all the information without pagination.
This affects workflows overviews and instances endpoints.


plugins/orchestrator-backend/src/service/api/v2.test.ts
```diff
@@ -109,6 +109,7 @@ describe('getWorkflowOverview', () => {
     ).mockResolvedValue(mockOverviewsV1.items);
 
     // Act
+    console.log('BUILD PAG:: ', buildPagination(mockRequest));
```
batzionb, line 112: Is this log needed?

batzionb (dismissed):

plugins/orchestrator-backend/src/service/api/v2.test.ts
gciavarrini: no :) removed!

gciavarrini:
@batzionb need to wait for this fix https://github.com/janus-idp/backstage-plugins/pull/1987

batzionb (approved):

PR #1987, opened 2024-07-31
fix(argocd): fix argocd-common plugin version

karthikjeeyar:
Fix incorrect argocd-common plugin reference.

cc: @invincibleJai


invincibleJai (approved):
/approve

PR #1988, opened 2024-07-31
feat(lightspeed): introduce lightspeed backend

dzemanov:
## Description:
Introduces a new backend plugin that can act as a proxy to LLM models

## Fixes:
[RHIDP-2949](https://issues.redhat.com/browse/RHIDP-2949)


plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -0,0 +1,74 @@
...
+  router.use(express.json());
+
+  router.get('/health', (_, response) => {
+    logger.info('PONG!');
+    response.json({ status: 'ok' });
+  });
+
+  router.post(
+    '/chat/completions',
+    validateCompletionsRequest,
```
dzemanov, line 32: Maybe we can validate only via openai's `APIError`, or via using `openapi.yaml` schema and `openapi-backend`

plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -0,0 +1,74 @@
...
+      });
+
+      try {
+        const stream = await openai.chat.completions.create({
+          messages: messages,
+          model: model,
+          stream: true,
+        });
+
+        for await (const chunk of stream) {
```
rohitkrai03, line 50: We would need the streaming support in the API as well so that we can show streaming data in the UI.

plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -0,0 +1,74 @@
...
+      const baseURL = config.getString('lightspeed.baseURL');
+      const apiKey = config.getString('lightspeed.apiKey');
+
+      const openai = new OpenAI({
+        baseURL: baseURL,
+        apiKey: apiKey,
+      });
+
+      try {
+        const stream = await openai.chat.completions.create({
```
rohitkrai03, line 44: Wouldn't it be better if we just do proxying of requests from UI to the API server? That way we can utilize any OpenAI's complaint API from the UI directly. We can build other APIs as needed into the backend like the chat history or something like that but I am not sure if we should wrap the completion API like this.

rohitkrai03 (changes requested):

plugins/lightspeed-backend/src/service/router.ts
dzemanov: Yes, I tried this approach in frontend plugin to read streamed data from backend (request has headers set to `'Content-Type': 'text/event-stream',`):
```
const reader = response.body!.pipeThrough(new TextDecoderStream())
      .getReader();

while (true) {
      const { value, done } = await reader.read();
      if (done) break;
      console.log('Received: ', value);`
}
```
Which logs data as they are streamed from backend.

plugins/lightspeed-backend/src/service/router.ts
dzemanov: Yes, that could be another approach. And then sending messages to save in separate API calls. We should discuss this and then redefine [RHIDP-2949](https://issues.redhat.com/browse/RHIDP-2949) if needed.

plugins/lightspeed-backend/README.md
```diff
@@ -0,0 +1,78 @@
+# Lightspeed Backend
+
+This is the lightspeed backend plugin that enables you to interact with any LLM server running a model with OpenAI's API compatibility.
```
nickboldt, line 3: New plugins should be contributed to https://github.com/backstage/community-plugins  as this repo is deprecated.

nickboldt (changes requested):
New plugins should be contributed to https://github.com/backstage/community-plugins as this repo is deprecated.

PR #1989, opened 2024-07-31
fix(argocd): use correct version argocd common lib

AndrienkoAleksandr:
### What does this pr do:

Use correct version argocd common lib

### DIFF
--- a/plugins/argocd/package.json
+++ b/plugins/argocd/package.json
@@ -39,7 +39,7 @@
     "@backstage/core-plugin-api": "^1.9.3",
     "@backstage/plugin-catalog-react": "^1.12.2",
     "@backstage/plugin-permission-react": "^0.4.24",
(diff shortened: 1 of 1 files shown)


PatAKnight:
It looks like what is happening is that the release process is updating the plugin whether it exists or not. Here is the same [PR](https://github.com/janus-idp/backstage-plugins/pull/1987/files) that was merged a little while ago. here is the [release](https://github.com/janus-idp/backstage-plugins/commit/e31c4a4b814c74da2dd336961135669e85949145) that happened shortly after.

I opened this [PR](https://github.com/janus-idp/backstage-plugins/pull/1990) in the hopes that it will solve that problem.

AndrienkoAleksandr:
Ok, closed on favour: https://github.com/janus-idp/backstage-plugins/pull/1987

PR #2007, opened 2024-08-05
feat(orchestrator): fix pagination issue and add filters to instances and overviews

gciavarrini:
Modify the definitions of `/v2/workflows/overview` and `/v2/workflows/instances` to properly handle pagination and to use filters.

Add unit tests for `queryBuilder` and `DataIndexService` to test the pagination and filter functionalities.


batzionb:
@gciavarrini 
PTAL one of the unit tests is failing

https://github.com/janus-idp/backstage-plugins/actions/runs/10253689160/job/28366890785?pr=2007#step:3:260

plugins/orchestrator-backend/src/service/api/v2.ts
```diff
@@ -75,19 +78,21 @@ export class V2 {
...
     const totalCount =
       await this.orchestratorService.fetchInstancesTotalCount();
 
     const result: ProcessInstanceListResultDTO = {
       items: instances?.map(mapToProcessInstanceDTO),
       paginationInfo: {
-        pageSize: pagination.limit,
-        page: pagination.offset,
+        pageSize: pagination?.limit,
+        page: pagination?.offset,
```
batzionb, line 95: this should be pagination?.offset / pageSize

plugins/orchestrator-backend/src/types/pagination.ts
```diff
@@ -8,23 +10,33 @@ export interface Pagination {
...
+  if (!req.body?.paginationInfo) {
+    return pagination;
   }
+  const { page, pageSize, orderBy, orderDirection } = req.body
+    .paginationInfo as PaginationInfoDTO;
 
-  if (!isNaN(Number(req.query.pageSize))) {
-    pagination.limit = Number(req.query.pageSize);
+  if (!isNaN(Number(page))) {
+    pagination.offset = Number(page);
```
batzionb, line 27: To fix an already existing bug,
This should be
pagination.offset = Number(page) * Number(pageSize)

plugins/orchestrator-backend/src/types/filters.ts
```diff
@@ -0,0 +1,40 @@
...
+      operator,
+      fieldValue: parseFilterValue(fieldValue),
+    };
+  }
+
+  return undefined;
+}
+
+function parseFilterValue(value: FilterValue): string | number | boolean {
+  if (typeof value === 'boolean') return Boolean(value);
```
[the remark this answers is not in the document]
gciavarrini: > if the type is boolean Boolean(value) equals to value`
correct. I'll fix it
>  if type is string ${value} equals to value
no, this method build the filter for graphQL query. I need to render the string properly (that is `"mystring"` with quotes)
>  as to the third condition - how can a value be also boolean and also number?
it's a typo. I'll replace `boolean` with `number`. thanks for catching this! :D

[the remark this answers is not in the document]
batzionb: no, this method build the filter for graphQL query. I need to render the string properly (that is "mystring" with quotes)
right
sorry, I missed the quotes

plugins/orchestrator-backend/src/types/pagination.ts
gciavarrini: what if `pageSize` is not defined?

plugins/orchestrator-backend/src/service/api/v2.ts
gciavarrini: ok, now i get it. Thanks for clarifying.
I'll change from `page` to `offset` to be, also,  "compliant" with graphql syntax.

plugins/orchestrator-backend/src/types/filters.ts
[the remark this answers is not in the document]
gciavarrini: addressed

plugins/orchestrator-backend/src/service/api/v2.ts
gciavarrini: done

batzionb (approved):

PR #2009, opened 2024-08-06
fix(acr): revert downgrade of shared-react and janus-cli

PatAKnight:
## Description

Reverts the downgrade for the `shared-react` and `janus-cli` dependencies in the ACR plugin.


gciavarrini (approved):
lgtm

PR #2015, opened 2024-08-06
fix(bulk-import): set initial value of annotations, labels and specs

ciiay:
Found this bug when testing #2013 .
Fixes: [RHIDP-3553](https://issues.redhat.com/browse/RHIDP-3533)
 Annotations, Labels, and Spec text field will have initial values as shown in Preview entities section.

Screen recording:

https://github.com/user-attachments/assets/f09c68dc-06bd-41ba-9db9-e7dae94a62af


debsmita1 (dismissed):
Thank you for the quick fix
Verified the changes. 
/lgtm

debsmita1 (approved):

PR #2031, opened 2024-08-08
feat(orchestrator): use v2 endpoint fot getWorkflowSource

gciavarrini:
Use v2 endpoints in orchestrator plugin UI for getting workflow source.

Warning: This DELETES corresponding v1 endpoints.

### DIFF
--- a/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
+++ b/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
@@ -217,10 +217,10 @@ const RefForwardingWorkflowEditor: ForwardRefRenderFunction<
             if (canceled.get()) {
               return;
             }
-            const definition = fromWorkflowSource(source);
+            const definition = fromWorkflowSource(source.data);
             setWorkflowDefinitionPromise({ data: definition });
 
-            const workflowFormat = extractWorkflowFormat(source);
+            const workflowFormat = extractWorkflowFormat(source.data);
 
             if (format && workflowId && format !== workflowFormat) {
               const link = viewWorkflowLink({
--- a/plugins/orchestrator-backend/src/service/api/v1.ts
+++ b/plugins/orchestrator-backend/src/service/api/v1.ts
@@ -26,19 +26,6 @@ export class V1 {
     return definition;
   }
 
-  public async getWorkflowSourceById(definitionId: string): Promise<string> {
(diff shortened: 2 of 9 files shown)


gciavarrini:
Need to wait for PR #2116 to fix CI

plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
```diff
@@ -217,10 +217,10 @@ const RefForwardingWorkflowEditor: ForwardRefRenderFunction<
             if (canceled.get()) {
               return;
             }
-            const definition = fromWorkflowSource(source);
+            const definition = fromWorkflowSource(source.data);
             setWorkflowDefinitionPromise({ data: definition });
 
-            const workflowFormat = extractWorkflowFormat(source);
+            const workflowFormat = extractWorkflowFormat(source.data);
```
batzionb, line 223: Something seems to not work here
With this PR when attempting to view a workflow with json format, such as Assessment Workflow from dev env, it navigates to /orchestrator/workflows/yaml/assessment instead of to  /orchestrator/workflows/json/assessment
This breaks the workflow viewer in this case

batzionb: The result of the API should be a string both for json and for yaml workflows
I noticed in the network tab that with main branch you see this in the response tab of the call to /source
image
But with this PR it's empty

plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
```diff
@@ -217,10 +217,10 @@ const RefForwardingWorkflowEditor: ForwardRefRenderFunction<
             if (canceled.get()) {
               return;
             }
-            const definition = fromWorkflowSource(source);
+            const definition = fromWorkflowSource(JSON.stringify(source.data));
```
batzionb, line 220: This is a workaround for the root cause, that the response should be a string
Better solve the root cause if possible

plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
gciavarrini: You right! 
Fixed here

See the results for: 
- Assessment
image

- Greeting yaml
image

batzionb (approved):

PR #2032, opened 2024-08-09
fix(acr): new credential settings in proxy backend

invincibleJai:
Fixes: https://issues.redhat.com/browse/RHIDP-3592 

This PR updated the credentials to be passed for the proxy and added more context around the new optional credential settings in the backstage 1.28 to explain different settings available to the users.

See Backstage 1.28 release notes about the BREAKING: Proxy backend plugin protected by default -
https://github.com/backstage/backstage/releases/tag/v1.28.0


karthikjeeyar (approved):
/lgtm

PR #2039, opened 2024-08-12
fix(rbac): clean up policies when policies files are detached

AndrienkoAleksandr:
# What does pull request do:
Clean up RBAC conditional policies, permission policies, roles from files after removing external files from configuration

# What does this pull request fix:

https://issues.redhat.com/browse/RHIDP-3601

### DIFF
--- a/plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
+++ b/plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
@@ -92,7 +92,6 @@ export class CSVFileWatcher extends AbstractFileWatcher<string[][]> {
    */
   async initialize(): Promise<void> {
     if (!this.filePath) {
-      this.logger.info('csv policies file feature was disabled');
       return;
     }
     let content: string[][] = [];
--- a/plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
+++ b/plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
@@ -475,7 +475,7 @@ export class CSVFileWatcher extends AbstractFileWatcher<string[][]> {
     this.csvFilePolicies.removedGroupPolicies = [];
   }
 
-  async cleanUpPolicies(): Promise<void> {
+  async cleanUpRolesAndPolicies(): Promise<void> {
     const roleMetadatas =
       await this.roleMetadataStorage.filterRoleMetadata('csv-file');
     const fileRoles = roleMetadatas.map(meta => meta.roleEntityRef);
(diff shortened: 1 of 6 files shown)


plugins/rbac-backend/src/service/permission-policy.ts
```diff
@@ -270,31 +270,38 @@ export class RBACPermissionPolicy implements PermissionPolicy {
...
+      auth,
+      pluginMetadataCollector,
+      roleMetadataStorage,
+      enforcerDelegate,
+    );
+    await conditionalFile.initialize();
+
+    if (!conditionalPoliciesFile) {
+      // clean up conditional policies corresponding to roles from csv file
+      logger.info('csv policies file feature was disabled');
```
PatAKnight, line 298:
```suggestion
      logger.info('conditional policies file feature was disabled');
```
Looks like this and line 303 are swapped

plugins/rbac-backend/src/service/permission-policy.ts
```diff
@@ -270,31 +270,38 @@ export class RBACPermissionPolicy implements PermissionPolicy {
...
+      await conditionalFile.cleanUpConditionalPolicies();
+    }
+    if (!policiesFile) {
+      // remove roles and policies from csv file
+      logger.info('conditional policies file feature was disabled');
```
PatAKnight, line 303:
```suggestion
      logger.info('csv policies file feature was disabled');
```

plugins/rbac-backend/src/file-permissions/csv-file-watcher.ts
```diff
@@ -474,4 +478,19 @@ export class CSVFileWatcher extends AbstractFileWatcher<string[][]> {
...
+  async cleanUpRolesAndPolicies(): Promise<void> {
+    const roleMetadatas =
+      await this.roleMetadataStorage.filterRoleMetadata('csv-file');
+    const fileRoles = roleMetadatas.map(meta => meta.roleEntityRef);
+
+    if (fileRoles.length > 0) {
+      this.csvFilePolicies.removedGroupPolicies =
+        await this.enforcer.getFilteredGroupingPolicy(1, ...fileRoles);
+      this.csvFilePolicies.removedPolicies =
+        await this.enforcer.getFilteredPolicy(0, ...fileRoles);
```
PatAKnight, line 491:
```suggestion
      for (const fileRole of fileRoles) {
        this.csvFilePolicies.removedGroupPolicies.push(
          ...(await this.enforcer.getFilteredGroupingPolicy(1, fileRole)),
        );
        this.csvFilePolicies.removedPolicies.push(
          ...(await this.enforcer.getFilteredPolicy(0, fileRole)),
        );
      }
```
I ran into a bug whenever I disabled my CSV file. This resulting in me not having any roles or permissions removed. I couldn't make what you had originally had work for me, so I poked around and came up with a potential solution. Don't know if you want to incorporate it or if you want to still try to fix what you originally had.

PatAKnight (commented):
Looks pretty good so far, just had a couple of suggestions. Also, something that I think we should add to our tech debt is the prevention of initializing the permission policies and roles that happen at start up whenever permissions are disabled. What do you think?

PatAKnight (approved):
Tested and it worked for me!

/lgtm

PR #2043, opened 2024-08-12
feat(rbac): add extensions to rbac backend plugin

PatAKnight:
## Description

Adds extension points to the RBAC backend plugin to allow for backend module support. This extension support will allow for the use of third party access management tools when it comes to loading in permission policies and roles. Included in this PR is an example backend module that can be paired with the RBAC backend plugin.

## Fixes
* Fixes: [RHIDP-2724](https://issues.redhat.com/browse/RHIDP-2724)

## Special Notes to the reviewer
Included is an example backend module that can be used to test the new feature. Within the example module is a CSV file in which you can manipulate the roles and permissions to test the scenarios of adding / removing permissions and roles. 

### Adding the test provider to test new functionality
Install the provider 
```
backend.add(
  import('@janus-idp/backstage-plugin-rbac-backend-module-test/alpha'),
);
```
Configure the test provider
```
permission:
  rbac:
    providers:
      test:
        baseUrl: http://localhost:8099
        accessToken: temp
        schedule:
          frequency: { minutes: 1 }
          timeout: { minutes: 1 }
          initialDelay: { seconds: 1 }
```
This will set the provider schedule to apply the roles and permissions from a csv file every minute.

To see the permissions and roles be updated, edit the csv file `plugins/rbac-backend-module-test/test-policy.csv`

Finally, to test the manual refresh capability use update the config to adjust the frequency of the schedule
```
permission:
  rbac:
    providers:
      test:
        baseUrl: http://localhost:8099
        accessToken: temp
        schedule:
          frequency: { minutes: 10 }
          timeout: { minutes: 1 }
          initialDelay: { seconds: 1 }

```
10 Minutes should give you enough time to edit the csv file and manually trigger refresh.

Edit the csv file with your desire changes.

Call the refresh endpoint
```
curl -X POST "http://localhost:7007/api/permission/refresh/testProvider" -H "Authorization: Bearer $token" -v
```
Should return a 200


plugins/rbac-backend/docs/providers.md
```diff
@@ -0,0 +1,328 @@
...
+
+import { TestProvider } from './TestProvider';
+
+/**
+ * The test backend module for the rbac plugin.
+ *
+ * @alpha
+ */
+export const rbacModuleTest = createBackendModule({
+  pluginId: 'rbac',
```
AndrienkoAleksandr, line 251: Can you fix it to permission too, please?

AndrienkoAleksandr:
Tested, works fine.

plugins/rbac-backend/docs/providers.md
```diff
@@ -0,0 +1,328 @@
...
+);
+```
+
+Configure the test provider in the `app-config`.
+
+```yaml
+permission:
+  rbac:
+    providers:
+      test:
```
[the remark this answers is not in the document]
PatAKnight: The doc is to create a new test provider. This isn't meant to mirror the test module that we currently have in the repo.

plugins/rbac-backend-module-test/src/provider/TestProvider.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.connection = connection;
+    this.scheduleFn();
+  }
+
+  async refresh(): Promise<void> {
+    try {
+      await this.run();
+    } catch (error: any) {
+      this.logger.error(`Error occurred, here is the error ${error}`);
+      console.log(error);
```
AndrienkoAleksandr, line 72: I think we can remove this console.log.

dzemanov:
Works by me as well.

<details open>
  <summary>Tested scenarios:
</summary>

- no provider

```
{"error":{"name":"NotFoundError","message":"The RBAC provider testProvider was not found","stack":"NotFoundError: The RBAC provider testProvider was not found\n    at <anonymous> (/home/dzemanov/Projects/janus2/backstage-plugins/plugins/rbac-backend/src/service/policies-rest-api.ts:1030:15)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"},"request":{"method":"POST","url":"/api/permission/refresh/testProvider"},"response":{"statusCode":404}}%
```

- Provider loaded - correctly loads data defined by provider

- Editing policies from provider by rbac ui correctly results in error
```
Unable to edit role: source does not match originating role role:default/test-provider, consider making changes to the 'TESTPROVIDER'
```

-  Changing csv file correctly updates - removing / adding works

-  Adding new role via ui with the same name as from provider correctly does not create


```
Error: Unable to create role. Unable to add role: source does not match originating role role:default/test-provider, consider making changes to the 'TESTPROVIDER'
```

-  Refresh endpoint works as expected
</details>

plugins/rbac-backend/src/providers/connect-providers.ts
```diff
@@ -0,0 +1,289 @@
...
+            source: this.id,
+          },
+          stage: HANDLE_RBAC_DATA_STAGE,
+          status: 'succeeded',
+        });
+      }
+    }
+  }
+
+  private async getProviderRoleMetadata(): Promise<string[]> {
```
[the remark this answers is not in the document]
PatAKnight: Done. Used `getProviderRoles` as we are getting the roles for the provider.

plugins/rbac-backend/src/providers/connect-providers.ts
```diff
@@ -276,14 +276,20 @@ export async function connectRBACProviders(
...
+        const connection = new Connection(
+          provider.getProviderName(),
+          enforcer,
+          roleMetadataStorage,
+          logger,
+          auditLogger,
+        );
+        return provider.connect(connection);
+      } catch (error) {
+        throw new Error(
```
AndrienkoAleksandr, line 289: Do we want cancel all providers, if one of them fail?

PatAKnight: I think it is fine for a first pass. I don't see a scenario where there would be a failure here as it should be captured [here](https://github.com/PatAKnight/backstage-plugins/blob/provide-rbac-extensions/plugins/rbac-backend-module-test/src/provider/TestProvider.ts#L90:L94). And that try catch is mainly for whenever a connection with the external provider fails. The actual connection to the provider module shouldn't fail as it is just assigning values.

Also, I haven't decided if we want the ability to have roles and permissions come from many providers. I think that it shouldn't but figured that is a discussion that we can have whenever work is started on the rbac-backend-keycloak-module.

AndrienkoAleksandr (approved):

PR #2050, opened 2024-08-13
feat(ocm): add openApi schema and OpenApiRouter

dzemanov:
## Description:
Introduces openapi schema for ocm-backend.
Introduces autogeneration of openapi documentation via `yarn build:api-docs`
Introduces `openapi-utils` dependency that adds:
 - A typed express router
 Some examples:
ocm-1
ocm-2
(When `openapi.yaml` specifies endpoint /example that requires name query param.)
- Validation and verification tooling
For example when calling `/example` endpoint without name:
```
 curl -X GET http://localhost:7007/api/ocm/hello                                                                                                                          ─╯
{"error":{"name":"InputError","message":"request/query must have required property 'name'","stack":"InputError: 
...
```
```
2024-08-13T14:48:35.814Z backstage error Request failed with status 400 request/query must have required property 'name' type=errorHandler cause=undefined name=InputError stack=InputError: request/query must have required property 'name'
    at /home/dzemanov/Projects/janus/backstage-plugins/node_modules/@backstage/backend-openapi-utils/dist/index.cjs.js:23:10
    at handleReturn (/home/dzemanov/Projects/janus/backstage-plugins/node_modules/express-promise-router/lib/express-promise-router.js:24:23)
...
```
- option to autogenerate client code like Apis via:
```yarn backstage-repo-tools package schema openapi generate client --client-package <directory>```
For ocm, this command could be run when inside ocm-backend folder:
```yarn backstage-repo-tools package schema openapi generate client --client-package plugins/ocm-common```
Example of how this autogenerated code looks like can be seen here:
https://github.com/backstage/backstage/tree/master/packages/catalog-client/src/generated

- this could be used to generate catalog entities with swagger openapi docs (haven't tried this one yet)
https://www.npmjs.com/package/@backstage/plugin-catalog-backend-module-backstage-openapi

The `openapi.generated.ts` needs to be regenerated when openapi schema changes via running:
`yarn build:api-server `. 
"@backstage/repo-tools"` must be installed in root workspace for it to run.

Reference: https://backstage.io/docs/openapi/01-getting-started/

## Fixes:
[RHIDP-3570](https://issues.redhat.com/browse/RHIDP-3570)

### CODE REVIEW

File: plugins/ocm-backend/scripts/openapi-doc.sh [hunk 1]
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+pwd
+set -ex
+
+# Generate doc
+npx --yes --package=@openapitools/openapi-generator-cli@2.13.4 -- \
+  openapi-generator-cli generate -i ./src/schema/openapi.yaml -g markdown -o ./api-docs/
```
PatAKnight, line 7: Fix new line

plugins/ocm-backend/src/schema/openapi.yaml
PatAKnight, line 1: What are your thoughts on creating another catalog entity that will point to this file? I think it would be nice to have as we can use be able to show the api specs from within Backstage and RHDH. It could be something like this 
```
# plugins/ocm-backend/catalog-info.yaml
---
apiVersion: backstage.io/v1alpha1
kind: API
metadata:
  name: janus-idp-ocm-backend-api
  title: '@janus-idp/backstage-plugin-ocm-backend'
  description: Open Cluster Management backend plugin OpenAPI specs for Backstage
  annotations:
    backstage.io/source-location: url:https://github.com/janus-idp/backstage-plugins/tree/main/plugins/ocm-backend
    backstage.io/view-url: https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm-backend/catalog-info.yaml
    backstage.io/edit-url: https://github.com/janus-idp/backstage-plugins/edit/main/plugins/ocm-backend/catalog-info.yaml
    github.com/project-slug: janus-idp/backstage-plugins
    #github.com/team-slug: janus-idp/maintainers
    sonarqube.org/project-key: janus-idp_backstage-plugins
  tags:
    - kubernetes
    - openshift
  links:
    - url: https://github.com/janus-idp/backstage-plugins/tree/main/plugins/ocm-backend
      title: GitHub Source
      icon: source
      type: source
spec:
  type: openapi
  lifecycle: production
  owner: rhdh-team
  system: rhdh
  subcomponentOf: janus-idp-ocm
  definition:
    $text: https://github.com/dzemanov/backstage-plugins/blob/ocm-openapi/plugins/ocm-backend/src/schema/openapi.yaml # This URL would need to change
```
And could then be registered in the app-config like this
```
    - type: file
      target: ../../plugins/ocm-backend/catalog-info.yaml
```
Then our specs would show up in the UI like [this](https://demo.backstage.io/catalog/default/api/petstore/definition).

We could possibly take this a step further and add catalog api entities to all of the plugins that current have OpenAPI specs. But that can be separate from this PR. WDYT?

PatAKnight (commented):
Tested and everything works!

One suggestion and one question for you.

plugins/ocm-backend/src/schema/openapi.yaml
dzemanov: Looks good. I tried swagger with this [Package](https://www.npmjs.com/package/@backstage/plugin-catalog-backend-module-backstage-openapi) and API showed under BackstageAPI, but it makes sense to also separate apis per component.

So adding something like `catalog-info2.yaml` and then reference it in `ocm-backend/catalog-info` via 
```
providesApis: [janus-idp-ocm-backend-api]
```

dzemanov:
One thing I would mention is that I didn't add `security` to `openapi.yaml` schema because that way openApiRouter checks that requests contain it and for dev development it was giving Error: Authorization header required.

plugins/ocm-backend/src/schema/openapi.yaml
dzemanov: Yes, that would be good

PatAKnight: > However, it calls 3000 endpoint as opposed to 7007, I am checking what can be done, since Backstage API definition correctly calls 7007.

Were you able to find a way around this?

dzemanov: https://github.com/user-attachments/assets/09f3a27f-053e-4c85-9aaf-85aac74bf29b

dzemanov: Wdyt @PatAKnight, would it be worth it to add the API entity with defined local server url?

PatAKnight: I was trying to test this out but what do you think about [this](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/bulk-import-backend/src/schema/openapi.yaml#L21:L32)? However, for some reason I can't seem to get it to work locally for me.

dzemanov: image
This works. I will wait for the ocm cluster to try it out with it.
What errors do you have?

PatAKnight: It involved getting the catalog to read in the file. But if you are able to get it to work, then that is all that matters!

dzemanov: Works with for now changed:
    $text: https://github.com/dzemanov/backstage-plugins/blob/ocm-openapi/plugins/ocm-backend/src/schema/openapi.yaml


image

plugins/ocm-backend/package.json
```diff
@@ -42,20 +42,23 @@
...
-    "tsc": "tsc"
+    "tsc": "tsc",
+    "build:api-docs": "./scripts/openapi-doc.sh",
+    "build:api-server": "yarn backstage-repo-tools package schema openapi generate --server"
   },
   "configSchema": "config.d.ts",
   "dependencies": {
     "@backstage/backend-common": "^0.23.3",
+    "@backstage/backend-dynamic-feature-service": "^0.2.15",
+    "@backstage/backend-openapi-utils": "^0.1.15",
```
dzemanov, line 53: I added this dependency. Do I need to add that to peerDependencies of dist-dynamic package.json?

PatAKnight, line 53: Sorry, I think I misunderstood you earlier. You were talking about these [peer dependencies](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm-backend/dist-dynamic/package.json#L67), right? Yes, I believe that they need to be added there as well.

PatAKnight (changes requested):

plugins/ocm-backend/package.json
dzemanov: Yes. I will update.

PatAKnight (commented):
Looks good, might need to pull in the latest changes to get the CI to pass

PatAKnight (approved):
/lgtm

PR #2056, opened 2024-08-14
fix(ocm): call OCM only for kubernetes-cluster entities

dzemanov:
## Description
This is a temporary fix to not call OCM on `entity.page.overview` mount for entities without `isType: kubernetes-cluster`.
Will need to sync with someone from dynamic plugins about context mounting, as `if` is supported only for `cards`.

## Fixes
https://issues.redhat.com/browse/RHIDP-2438

### DIFF
--- a/plugins/ocm/src/components/ClusterContext/ClusterContext.test.tsx
+++ b/plugins/ocm/src/components/ClusterContext/ClusterContext.test.tsx
@@ -1,27 +1,31 @@
 import React from 'react';
 
+import { useEntity } from '@backstage/plugin-catalog-react';
+
 import { render } from '@testing-library/react';
 
 import data from '../../../__fixtures__/cluster1.json';
 import { ClusterContextProvider } from './ClusterContext';
+
+const mockEntity = {
+  apiVersion: 'backstage.io/v1beta1',
+  kind: 'Resource',
+  spec: { owner: 'unknown', type: 'kubernetes-cluster' },
+  metadata: {
+    name: 'foo',
+    namespace: 'default',
+    annotations: {
+      'janus-idp.io/ocm-provider-id': 'hub',
+    },
+  },
+};
 
 jest.mock('../ClusterContext/', () => ({
   useCluster: jest.fn().mockReturnValue({}),
 }));
 
 jest.mock('@backstage/plugin-catalog-react', () => ({
-  useEntity: () => ({
-    apiVersion: 'backstage.io/v1beta1',
-    kind: 'Resource',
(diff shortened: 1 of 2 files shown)


durandom:
@kadel @gashcrumb can you answer @dzemanov question here? Maybe @dzemanov can give a bit more context

dzemanov:
Right now, if user specifies context mounting, it will be always mounted, if I understand correctly. However, it would be good to not add context when it is not needed.

For example, for ocm setup:
```
    janus-idp.backstage-plugin-ocm:
          appIcons:
            - name: ocmIcon
              importName: OcmIcon
          dynamicRoutes:
            - path: /ocm
              importName: OcmPage
              menuItem:
                icon: ocmIcon
                text: Clusters
          mountPoints:
            - mountPoint: entity.page.overview/context
              importName: ClusterContextProvider  # <-- provider always rendered for every `entity.page.overview`
            - mountPoint: entity.page.overview/cards
              importName: ClusterAvailableResourceCard
              config:
                layout:
                  gridColumnEnd:
                    lg: "span 4"
                    md: "span 6"
                    xs: "span 12"
                if:
                  allOf:
                  - isKind: resource
                  - isType: kubernetes-cluster
            - mountPoint: entity.page.overview/cards
              importName: ClusterInfoCard
              config:
                layout:
                  gridColumnEnd:
                    lg: "span 4"
                    md: "span 6"
                    xs: "span 12"
                if:
                  allOf:
                  - isKind: resource
                  - isType: kubernetes-cluster
```

`ClusterAvailableResourceCard` and `ClusterInfoCard` will be rendered only if `if` condition is met - for cluster resources. However, context is rendered always, no matter the entity. This means, that e.g when user clicks on user entity and overview page is rendered for it, we make requests to OCM, although they are not needed. Furthermore, we make invalid requests because context provider expects certain entity type.

If you try and add this to context provider and mount it, you can see it is rendered in every entity overview page:
```
  return (
    <ClusterContext.Provider value={value}>
      <div
        style={{ border: '1px solid red', padding: '10px', marginTop: '20px' }}
      >
        <h2>The OCM API has been invoked to retrieve cluster information.</h2>
      </div>
      {props.children}
    </ClusterContext.Provider>
  );
```

image


Maybe someone from frontend team can confirm, but I think it would be safe to add context only if condition is met in `DynamicEntityTab.tsx`.
Maybe some changes in code here: https://github.com/janus-idp/backstage-showcase/blob/main/packages/app/src/components/catalog/EntityPage/DynamicEntityTab.tsx#L54-L55

gashcrumb (approved):
:+1: it's worth doing this check and preventing the API call for this condition anyways.

PR #2059, opened 2024-08-16
fix(cli): support BackendFeatureCompat in export-dynamic cli

JessicaJHee:
This is as a result of a new [change](https://backstage.io/docs/reference/backend-plugin-api.createbackendmodule) where createBackendModule() now returns `BackendFeatureCompat`, which can have a call signature/ can be a function. This change allows both the old and new behaviour to be recognized as valid backend features. 

Fixes [RHIDP-2529](https://issues.redhat.com/browse/RHIDP-2529)
* More details on the error in the issue

Tested this change locally and can confirm that the `yarn export-dynamic` cli now succeeds with no errors


Zaperex (approved):
/lgtm

kim-tsao (approved):

PR #2065, opened 2024-08-19
fix(rbac): validate name and namespace of entities

dzemanov:
## Description
Creating role with name that contains ':' or '/' creates role that can't be used.

Invalidate names that does not conform with format:
- Strings of length at least 1, and at most 63
- Must consist of sequences of [a-z0-9A-Z] possibly separated by one of [-_.]

Invalidate namespaces that does not conform with format:
- must be sequences of [a-zA-Z0-9], possibly separated by -, at most 63 characters in total.  

Deletes function `getKindNamespaceName` and uses instead more robust `parseEntityRef`. This change was originally introduced to support `:`, can be reverted.

Reference:
https://backstage.io/docs/features/software-catalog/descriptor-format/#name-required

## Fixes
https://issues.redhat.com/browse/RHIDP-3580

### DIFF
--- a/plugins/rbac-backend/src/validation/policies-validation.test.ts
+++ b/plugins/rbac-backend/src/validation/policies-validation.test.ts
@@ -203,6 +203,10 @@ describe('rest data validation', () => {
           ref: 'admin:default/test',
           expectedError: `Unsupported kind admin. List supported values ["user", "group", "role"]`,
         },
+        {
+          ref: 'user:default/doe/developer',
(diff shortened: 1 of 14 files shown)


dzemanov:
Maybe we should consider as valid role/group/user names only:
[a-zA-Z0-9], [-_.]

AndrienkoAleksandr (commented):
Looks good to me. @PatAKnight ?

PatAKnight (approved):
Yep, makes sense to me. 
LGTM

PR #2067, opened 2024-08-20
chore: start frontend and backend with yarn dev

jerolimov:
One extra run script that makes it easier to switch between backstage/rhdh projects.

Adds `yarn dev` to start frontend and backend together.

`yarn start:backstage` still work, of course.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -455,7 +455,7 @@ Backstage's support for standalone plugin development is minimal (especially for
        packages/app/package.json \
        packages/app/src/components/catalog/EntityPage.tsx
    ```
-4. Run `yarn start:backstage`
+4. Run `yarn dev`
 
 ### Standalone standalone plugin development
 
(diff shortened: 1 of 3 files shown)


jerolimov:
Wdyt  @schultzp2020, I believe you setup most of that scripts?

schultzp2020:
I don't think we should switch the script name because we will soon be migrating away from here.

AndrienkoAleksandr:
For now `yarn start` command execute only backend in the backstage-showcase 

https://github.com/janus-idp/backstage-showcase/blob/main/package.json#L12

To start dev instance you have to use `yarn dev`:

https://github.com/janus-idp/backstage-showcase/blob/main/package.json#L14

jerolimov:
Thanks @AndrienkoAleksandr, you're right. I changed the PR from `yarn start` to `yarn dev`

/cc @AndrienkoAleksandr @debsmita1

debsmita1 (approved):

PR #2081, opened 2024-08-22
feat: update keycloak-admin-client to use v24.0.5

04kash:
- Used the [inclusion package](https://www.npmjs.com/package/inclusion) to dynamically import KeycloakAdminClient
- Added `processGroupsRecursively` function to `read.ts` to get subgroups correctly to resolve breaking change in Keycloak  23+
- Updated Unit tests to take into account the above changes
- Manually tested using RHBK v22 and v24, and also Keycloak v18

Fixes: https://issues.redhat.com/browse/RHIDP-3542


04kash:
Do we want to support older keycloak versions as well?

kim-tsao:
@04kash it looks like these changes will break integration with RHSSO 7.6 (keycloak v18)?  We still need to maintain backward compatibility.  

Also, let's align the supported Keycloak client with the productized RHBK version v24.  I don't know if v25 is compatible

04kash:
Yes, according to https://www.keycloak.org/docs/25.0.4/upgrading/#upgrading-the-keycloak-admin-client:
> earlier versions of Keycloak server might not work with later versions of the admin-client

I think it would be best to use v24 for the Keycloak client

kim-tsao:
> Do we want to support older keycloak versions as well?

No, just v18.  We know the client can tolerate Keycloak v22 so no need to update

JessicaJHee (commented):
Looks great! I tested on my end and the users/groups (and sub groups) are being ingested properly.

Just one thing, I was wondering why `plugins/keycloak-backend/dist-dynamic/package.json` and `plugins/keycloak-backend/dist-dynamic/yarn.lock` was deleted? When I ran `yarn export-dynamic` in `plugins/keycloak-backend` it was regenerated.

04kash:
Thanks for taking a look at this @JessicaJHee! 

> Just one thing, I was wondering why plugins/keycloak-backend/dist-dynamic/package.json and plugins/keycloak-backend/dist-dynamic/yarn.lock was deleted? When I ran yarn export-dynamic in plugins/keycloak-backend it was regenerated. 

I deleted these as the CI was failing when I added them.

plugins/keycloak-backend/__fixtures__/helpers.ts
```diff
@@ -2,14 +2,19 @@ import { getVoidLogger } from '@backstage/backend-common';
...
+  users,
+} from './data';
 
 export const BASIC_VALID_CONFIG = {
   catalog: {
     providers: {
       keycloakOrg: {
         default: {
-          baseUrl: 'http://localhost:8080/auth',
+          baseUrl: 'http://localhost:8080',
```
kim-tsao, line 17: I was wondering why `auth` was removed and saw that there's a comment in our [README](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/README.md#new-backend-configuration) stating that it's not needed for Keycloak v17+.  Can you update the readme examples (on both the legacy and new backends) to not have `auth` so it's correct by default? 

I also checked our [product docs](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.2/html/configuring_plugins_in_red_hat_developer_hub/rhdh-installing-dynamic-plugins#advanced-configuration) and we mention it there as well.  Let's make sure this gets updated for 1.4 as well

plugins/keycloak-backend/src/lib/read.ts
```diff
@@ -165,24 +190,61 @@ export const readKeycloakRealm = async (
...
   );
 
-  const rawKGroups = await getEntities(
+  const topLevelKGroups = (await getEntities(
     client.groups,
     config,
     options?.groupQuerySize,
+  )) as GroupRepresentationWithParent[];
+
+  const serverInfo = await client.serverInfo.find();
```
kim-tsao, line 199: This is a REST API call to retrieve the server info.  Can we have some error handling if the server gives a bad response?

kim-tsao:
@04kash in addition to the review comments:

1.  Can you also update the [metadata](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/dist-dynamic/package.json#L57-L58) to state what versions this plugin integrates with?

    "works-with:rhbk:^24",
    "works-with:keycloak:~24"

     We can just stick with the versions that were tested.  If someone wants to use this plugin with Keycloak 23/25, they can do so at their own risk

2.  Update the [minor version](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/package.json#L3) of this plugin to `1.14.0` (note, planned main branch version bumps may be doing this for us depending on when we get this PR merged,  so we should align with those versions).  Minor version updates allow us to adhere to [semver](https://semver.org/) since we are introducing functionality but doing so in a non-breaking manner

plugins/keycloak-backend/__fixtures__/helpers.ts
04kash: Updated! I'll also create a separate issue to update this in our products.

plugins/keycloak-backend/README.md
```diff
@@ -30,8 +30,8 @@ yarn workspace backend add @janus-idp/backstage-plugin-keycloak-backend
      providers:
        keycloakOrg:
          default:
-           # Remove the `/auth` if using keycloak 17+
-           baseUrl: https://<keycloak_host>/auth
+           # Add `/auth` to the baseUrl if using Keycloak versions older than 17
```
kim-tsao, line 33: Let's remove this comment altogether since it implies we could support Keycloak < 17.  We can safely assume all the older versions are EOL

plugins/keycloak-backend/README.md
```diff
@@ -220,8 +220,8 @@ yarn workspace backend add @janus-idp/backstage-plugin-keycloak-backend
...
```
kim-tsao, line 223: same comment as above

plugins/keycloak-backend/README.md
```diff
@@ -307,17 +307,17 @@ Communication between Backstage and Keycloak is enabled by using the Keycloak AP
...
-| `userQuerySize`  | Number of users to query at a time                                                                                             | `100`         | No                                                   |
-| `groupQuerySize` | Number of groups to query at a time                                                                                            | `100`         | No                                                   |
+| Name             | Description                                                                                                                             | Default Value | Required                                             |
+| ---------------- | --------------------------------------------------------------------------------------------------------------------------------------- | ------------- | ---------------------------------------------------- |
+| `baseUrl`        | Location of the Keycloak server, such as `https://localhost:8443`. Add `/auth` to the baseUrl if using Keycloak versions older than 17. | ""            | Yes                                                  |
```
kim-tsao, line 312: same comment as above

schultzp2020 (approved):
/lgtm
/approve

PR #2087, opened 2024-08-26
fix(rbac): role name update should redirect to correct overview page

divyanshiGupta:
Fixes: https://issues.redhat.com/browse/RHIDP-3665

https://github.com/user-attachments/assets/8a9dbeaa-2e3e-41e5-845f-e66009581ce2

### DIFF
--- a/plugins/rbac/src/components/CreateRole/RoleForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/RoleForm.tsx
@@ -69,12 +69,18 @@ export const RoleForm = ({
   const navigate = useNavigate();
   const rbacApi = useApi(rbacApiRef);
 
-  const navigateTo = (action?: string) => {
-    const stateProp = action
-      ? { state: { toastMessage: `Role ${action} successfully` } }
-      : { state: { toastMessage: '' } };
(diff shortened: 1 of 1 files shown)


plugins/rbac/src/components/CreateRole/RoleForm.tsx
```diff
@@ -72,12 +72,13 @@ export const RoleForm = ({
...
-    const stateProp = action
-      ? { state: { toastMessage: `Role ${action} successfully` } }
-      : { state: { toastMessage: '' } };
-    if (step && roleName) {
-      const { kind, namespace, name } = getKindNamespaceName(roleName);
+  const navigateTo = (rName?: string, action?: string) => {
+    const stateProp =
+      rName && action
+        ? { state: { toastMessage: `Role ${rName} ${action} successfully` } }
+        : { state: { toastMessage: '' } };
```
jerolimov, line 79: Couldn't the stateProp be undefined when there is no action instead of an empty (string) toastMessage?

jerolimov (approved):
Tested it locally and it works :+1:

/lgtm

PR #2089, opened 2024-08-26
fix(topology): fixes custom styles loading

invincibleJai:
Fixes:

https://issues.redhat.com/browse/RHIDP-3710, https://issues.redhat.com/browse/RHIDP-3466

**Description:** 

The sideEffects key in a package.json file is used to inform bundlers like Webpack about the presence of side effects in your code. This key plays a crucial role in optimizing the bundle size during the build process by enabling tree shaking, which is the process of eliminating unused code from the final bundle.

In the context of JavaScript, a side effect refers to any operation or code that affects something outside its scope, such as modifying a global variable, performing I/O operations, or altering the DOM. Pure functions, on the other hand, have no side effects—they return the same output given the same input and do not alter external states.

With sideEffects false, custom styles were being removed as part of optimization by webpack. 

custom css are supported loaders for backstage https://backstage.io/docs/tooling/cli/build-system/#loaders 

https://webpack.js.org/guides/tree-shaking/

**Before:**

Topology:

image

Tekton:

image

**After:**

Topology:

image

Tekton:

image

### DIFF
--- a/plugins/tekton/package.json
+++ b/plugins/tekton/package.json
@@ -19,7 +19,9 @@
       "@janus-idp/backstage-plugin-tekton-common"
     ]
   },
-  "sideEffects": false,
+  "sideEffects": [
+    "./**/*.css"
+  ],
   "scripts": {
     "build": "backstage-cli package build",
     "clean": "backstage-cli package clean",
--- a/plugins/topology/package.json
+++ b/plugins/topology/package.json
@@ -18,7 +18,9 @@
       "@janus-idp/backstage-plugin-topology-common"
     ]
   },
-  "sideEffects": false,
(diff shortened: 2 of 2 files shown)


karthikjeeyar (commented):
SideEffects field is boolean, check this out https://github.com/janus-idp/backstage-plugins/blob/main/packages/cli/src/commands/metadata/command.ts#L55

I think It should work if you simply set it to `true`


Edit: I see Webpack accepts glob patterns as well https://webpack.js.org/guides/tree-shaking/ so this changes looks good.

karthikjeeyar (approved):
Verified locally, it now loads the external css files.

/lgtm

PR #2090, opened 2024-08-26
chore(doc): add audit log event naming convention

kim-tsao:
fixes https://issues.redhat.com/browse/RHIDP-3717

### DIFF
--- a/plugins/audit-log-node/README.md
+++ b/plugins/audit-log-node/README.md
@@ -24,6 +24,10 @@ The `DefaultAuditLogger.createAuditLogDetails` will generate the `actorId` of th
 - The actor id generated from the `express.Request` object provided in the arguments
 - `null` if neither of the above fields were provided in the arguments
 
+#### Event Naming Convention
+
+It is recommended that you prefix the `eventName` value with the name of the component you are audit logging. This will help with searchability in the central log collector.
+
 ---
 
 **IMPORTANT**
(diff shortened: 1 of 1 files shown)


plugins/audit-log-node/README.md
```diff
@@ -24,6 +24,10 @@ The `DefaultAuditLogger.createAuditLogDetails` will generate the `actorId` of th
 - The actor id generated from the `express.Request` object provided in the arguments
 - `null` if neither of the above fields were provided in the arguments
 
+#### Event Naming Convention
+
+It is recommended that you prefix the `eventName` value with the name of the component you are audit logging. This will help with searchability in the central log collector.
```
Zaperex, line 29:
```suggestion
It is recommended that you prefix the `eventName` value with the name of the component you are audit logging. This will help with searchability in the central log collector.

For example, "ScaffolderTaskRead", "CatalogEntityFetch", etc.
```

Zaperex (changes requested):
I guess we should give an example of what it should look like?

plugins/audit-log-node/README.md
Zaperex: Feel free to reword it

kim-tsao: sure sounds good.  I actually added one of the Scaffolder events but thought it cluttered the paragraph.  What you have works!

rm3l (approved):

PR #2096, opened 2024-08-27
fix(rbac): remove tab from rbac page

ciiay:
For [RHIDP-3649
](https://issues.redhat.com/browse/RHIDP-3649)Need to get rid of the tab structure from RBAC page for the new navigation design.
This PR works along with updates in this [backstage-showcase PR](https://github.com/janus-idp/backstage-showcase/pull/1507).

Screen recording:

https://github.com/user-attachments/assets/323346d5-a526-4fea-a722-49f58cc2d0ce

Screenshot from backstage-showcase:
image

### DIFF
--- a/plugins/bulk-import-backend/package.json
+++ b/plugins/bulk-import-backend/package.json
@@ -57,6 +57,7 @@
     "@backstage/plugin-catalog-node": "^1.12.4",
     "@backstage/plugin-permission-common": "^0.8.0",
     "@backstage/plugin-permission-node": "^0.8.0",
+    "@janus-idp/backstage-plugin-audit-log-node": "1.4.0",
     "@janus-idp/backstage-plugin-bulk-import-common": "1.0.0",
     "@octokit/auth-app": "^6.0.3",
     "@octokit/core": "^5.1.0",
--- a/plugins/bulk-import-backend/package.json
+++ b/plugins/bulk-import-backend/package.json
@@ -57,7 +57,6 @@
     "@backstage/plugin-catalog-node": "^1.12.4",
     "@backstage/plugin-permission-common": "^0.8.0",
     "@backstage/plugin-permission-node": "^0.8.0",
-    "@janus-idp/backstage-plugin-audit-log-node": "1.4.0",
     "@janus-idp/backstage-plugin-bulk-import-common": "1.0.0",
     "@octokit/auth-app": "^6.0.3",
     "@octokit/core": "^5.1.0",
--- a/plugins/rbac/src/components/RbacPage.tsx
+++ b/plugins/rbac/src/components/RbacPage.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { Header, Page, TabbedLayout } from '@backstage/core-components';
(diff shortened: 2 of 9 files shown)


plugins/rbac/src/components/RbacPage.tsx
```diff
@@ -1,40 +1,25 @@
...
-            </TabbedLayout>
-          </DeleteDialogContextProvider>
-        </Page>
-      </RequirePermission>
-    ) : (
-      <RequirePermission
-        permission={policyEntityReadPermission}
-        resourceRef={policyEntityReadPermission.resourceType}
-      >
+export const RbacPage = () => (
```
invincibleJai, line 11: do we need the change here i.e if some user is not updating to the new config and using old way then for them tabbed view should as well work or no?

plugins/rbac/src/index.ts
```diff
@@ -1,3 +1,3 @@
 export { rbacPlugin, RbacPage, Administration } from './plugin';
 
-export { default as AdminPanelSettingsOutlinedIcon } from '@mui/icons-material/AdminPanelSettingsOutlined';
+export { default as RbacIcon } from '@mui/icons-material/VpnKeyOutlined';
```
invincibleJai, line 3: let's ensure that it's backword compatible and not remove any exports i.e keep old one add new one as well and maybe make a comment as deprecated for now WDYT?

invincibleJai:
@ciiay thanks 👍 

Let's ensure we test this changes so that it is backward compatible if not already done it i.e it works with/without new configurations.

plugins/rbac/src/components/RbacPage.tsx
invincibleJai, line 11: let keep the prop `useHeader` as this might be used by some customer but pls remove the tabbedLayout and update title to RBAC

plugins/rbac/src/components/Router.tsx
```diff
@@ -38,7 +38,7 @@ export const Router = ({ useHeader = true }: { useHeader?: boolean }) => {
   return (
     <ToastContextProvider>
       <Routes>
-        <Route path="/" element={<RbacPage useHeader={useHeader} />} />
+        <Route path="/" element={<RbacPage />} />
```
invincibleJai, line 41: based on above this change might not be needed, check once

invincibleJai (approved):
/lgtm

invincibleJai (dismissed):
/lgtm

invincibleJai (dismissed):
/lgtm

invincibleJai (approved):
/lgtm

plugins/bulk-import-backend/package.json
```diff
@@ -57,6 +57,7 @@
     "@backstage/plugin-catalog-node": "^1.12.4",
     "@backstage/plugin-permission-common": "^0.8.0",
     "@backstage/plugin-permission-node": "^0.8.0",
+    "@janus-idp/backstage-plugin-audit-log-node": "1.4.0",
```
invincibleJai, line 60: why is this change needed?

invincibleJai (approved):
/lgtm

PR #2098, opened 2024-08-28
fix(orchestrator): make "Retrigger" clear to the user

mareklibra:
[FLPATH-1600](https://issues.redhat.com//browse/FLPATH-1600)

### DIFF
--- a/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
+++ b/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
@@ -169,6 +169,13 @@ export const WorkflowInstancePage = ({
     navigate(urlToNavigate);
   }, [value, navigate, executeWorkflowLink]);
 
+  let retriggerTooltip =
+    'Resume the workflow from the error stage, continuing with the same instance.';
+  if (!permittedToExecute.allowed) {
+    retriggerTooltip = 'The user is not authorized to execute workflow';
+  }
+  const isRetriggerDisabled = !permittedToExecute.allowed;
+
   return (
     <BaseOrchestratorPage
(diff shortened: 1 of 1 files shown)


mareklibra:
Screenshot from 2024-08-28 11-38-01

---

Screenshot from 2024-08-28 11-38-07

plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -208,17 +215,14 @@ export const WorkflowInstancePage = ({
...
-                  <Tooltip
-                    title="user not authorized to execute workflow"
-                    disableHoverListener={permittedToExecute.allowed}
-                  >
+                  <Tooltip title={retriggerTooltip}>
                     <Button
                       variant="contained"
                       color="primary"
-                      disabled={!permittedToExecute.allowed}
+                      disabled={isRetriggerDisabled}
```
gciavarrini, line 222: does it make sense to show the button if the action can't be performed? Is it possible to hide the button?
What's your opinion? Not sure if this can improve the user experience.

mareklibra: It's an UX topic discussed again and again, there are 50% people who like hiding and 50% who don't :-)

I incline to the opinion that from the user's perspective is better to know that a feature is there but can not be executed right now because of reason XY.

Hiding the button can make the user confused since "I know it was somewhere here but can not find it now. Where was it? Different page? I am confused. Do I need to do something to see it again? What is that?"

gciavarrini: It makes sense, thanks :)

PR #2129, opened 2024-09-05
feat(bulk-import): introduce api catalog info

dzemanov:
## Description

Introduces OpenApi definition.

To see swagger endpoints, add to `app-config.local.yaml`:
```
    - type: file
      target: ../../plugins/bulk-import-backend/catalog-info.yaml
```
(Or use url once merged.)

bulk-import

## Fixes
[RHIDP-3797](https://issues.redhat.com/browse/RHIDP-3797)

### DIFF
--- a/plugins/bulk-import-backend/catalog-info.yaml
+++ b/plugins/bulk-import-backend/catalog-info.yaml
@@ -23,6 +23,9 @@ spec:
   owner: rhdh-team
   system: rhdh
   subcomponentOf: janus-idp-bulk-import
+  providesApis:
+    - janus-idp-bulk-import-backend-api
+
 ---
 # https://backstage.io/docs/features/software-catalog/descriptor-format#kind-api
 apiVersion: backstage.io/v1alpha1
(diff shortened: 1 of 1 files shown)


plugins/bulk-import-backend/catalog-info.yaml
```diff
@@ -23,3 +23,31 @@ spec:
   owner: rhdh-team
   system: rhdh
   subcomponentOf: janus-idp-bulk-import
+---
```
rm3l, line 26:
```suggestion
  providesApis:
    - janus-idp-bulk-import-backend-api

---
```

Should we add a `providesApi` relation between the Component and API entities, just like https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm-backend/catalog-info.yaml#L29C3-L30 ?

rm3l (commented):
Thanks for this, @dzemanov !

plugins/bulk-import-backend/catalog-info.yaml
dzemanov: Oh pardon, forgot this. Fixed.

rm3l (approved):

PR #2130, opened 2024-09-05
fix(orchestrator): typos in auditLog messages

gciavarrini:


batzionb (approved):

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -876,7 +876,7 @@ function setupInternalRoutes(
     'abortWorkflow',
     async (c, _req, res, next) => {
       const instanceId = c.request.params.instanceId as string;
-      const endpointName = 'getInstanceById';
+      const endpointName = 'abortWorkflow';
       const endpoint = `/v2/instances/${instanceId}/abort`;
```
rgolangh, line 880: /v2/instances here also looks wrong @batzionb

rgolangh: https://github.com/janus-idp/backstage-plugins/pull/2141

batzionb: @gciavarrini

gciavarrini: @rgolangh i wouldn't say "wrong" but inconsistent. Thanks for the fix :)

PR #2133, opened 2024-09-05
chore(deps): move audit log dep to peerDependencies

PatAKnight:
## Description

For some reason the release started to fail for `@janus-idp/backstage-plugin-bulk-import-backend-dynamic` after the audit log [PR](https://github.com/janus-idp/backstage-plugins/pull/1966) was merged. Failed release can be found [here](https://github.com/janus-idp/backstage-plugins/actions/runs/10581736766/job/29319768586).

To hopefully fix this, I moved `@janus-idp/backstage-plugin-audit-log-node` from a dependency to a peerDependency. I got the inspiration from the [Orchestrator backend plugin](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator-backend/package.json#L112). 

@rm3l I am not 100% on if this is the actual problem or not but figure that I could open this PR and we could discuss it here.

Also, included in the PR is a fix for some of the other plugins that were affected by the failed release process.

### DIFF
--- a/plugins/bulk-import-backend/dist-dynamic/package.json
+++ b/plugins/bulk-import-backend/dist-dynamic/package.json
@@ -28,7 +28,6 @@
   },
   "scripts": {},
   "dependencies": {
-    "@janus-idp/backstage-plugin-audit-log-node": "^1.4.0",
     "@octokit/auth-app": "^6.0.3",
     "@octokit/core": "^5.1.0",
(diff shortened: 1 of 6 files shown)


rm3l (approved):

rm3l:
Thanks @PatAKnight for looking into this. Let's see how it goes🤞🏿

PR #2135, opened 2024-09-05
fix(deps): remove unncessary dependency of older version of orchestrator-form-react package

batzionb:


batzionb:
small fix following https://github.com/janus-idp/backstage-plugins/pull/2133

PR #2152, opened 2024-09-11
chore(release): revert downgrade of a few plugins

rm3l:
This reverts commit 41176498b3fbdcc3a6d7fa32808c33f4b0f7664c.

Looks like another MSR issue again after https://github.com/janus-idp/backstage-plugins/pull/2150 was merged - logs in [1].

[1] https://github.com/janus-idp/backstage-plugins/actions/runs/10810499470/job/29990246035

/cc @PatAKnight @AndrienkoAleksandr

PR #2153, opened 2024-09-11
chore(release): revert downgrade of bulk import backend

PatAKnight:
## Description

Fix the MSR downgrade of bulk import backend that triggered a yarn lock PR.

/cc @rm3l


PatAKnight:
close in favor of #2152

PR #2162, opened 2024-09-12
chore(deps): fix CVE-2024-45590, CVE-2024-43799, CVE-2024-45296

kim-tsao:
Mostly a bump to update express to v4.21.0 which updates a number of transitive dependencies that fix the following CVEs:
 * CVE-2024-45590
 * CVE-2024-43799
 * CVE-2024-45296
 
 
 Fixes: 
 https://issues.redhat.com/browse/RHIDP-3917
 https://issues.redhat.com/browse/RHIDP-3945


kim-tsao:
> Didn't worry about testing the rbac-backend-module-test as it is just a test module for the rbac plugin and isn't released
> 
> Checked on ocm backend plugin and it looks good
> 
> Looks good to me, going to hold off on approving in case any other teams want to check on their plugins.

yes, it's test only but I updated it just to keep things current and consistent.  Thanks for taking a look!

aljesusg (approved):
LGFM

PR #2169, opened 2024-09-13
fix(orchestrator): improve error handling in SonataFlowService fetch

gciavarrini:
# Current issue

In some cases, workflows fail without providing any information about the root cause, making debugging difficult.

For example: 

image

The backstage log contains:
```
backend:start: {"level":"error","message":"Error when fetching process instances: [GraphQL] Variable 'instanceId' has an invalid value: Variable 'instanceId' has coerced Null value for NonNull type 'String!'","plugin":"orchestrator","service":"backstage","timestamp":"2024-09-19 17:33:03"}
backend:start: {"actor":{"actorId":"user:development/guest","hostname":"localhost","ip":"::ffff:127.0.0.1","userAgent":"Mozilla/5.0 (X11; Linux x86_64; rv:130.0) Gecko/20100101 Firefox/130.0"},"errors":[{"message":"[GraphQL] Variable 'instanceId' has an invalid value: Variable 'instanceId' has coerced Null value for NonNull type 'String!'","name":"CombinedError","stack":"CombinedError: [GraphQL] Variable 'instanceId' has an invalid value: Variable 'instanceId' has coerced Null value for NonNull type 'String!'\n    at makeResult (/home/gloria/repos/backstage-plugins-devel/node_modules/@urql/core/src/utils/result.ts:43:9)\n    at fetchOperation (/home/gloria/repos/backstage-plugins-devel/node_modules/@urql/core/src/internal/fetchSource.ts:204:11)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async /home/gloria/repos/backstage-plugins-devel/node_modules/wonka/dist/wonka.js:346:20"}],"eventName":"executeWorkflowEndpointHit","isAuditLog":true,"level":"error","message":"Error occured while requesting the '/v2/workflows/m2k/execute' endpoint","meta":{},"plugin":"orchestrator","request":{"body":{"inputData":{"projectId":"p1","recipients":["user:default/jsmith"],"repositoryURL":"https://bitbucket.org/gfarache31/m2k-test","sourceBranch":"master","targetBranch":"g1","workspaceId":"w1"}},"method":"POST","params":{},"query":{},"url":"/api/orchestrator/v2/workflows/m2k/execute"},"response":{"body":{"errors":[{"message":"[GraphQL] Variable 'instanceId' has an invalid value: Variable 'instanceId' has coerced Null value for NonNull type 'String!'","name":"CombinedError"}]},"status":500},"service":"backstage","stage":"completion","status":"failed","timestamp":"2024-09-19 17:33:03"}
```

## Cause

The `fetch` function does not throw an exception for HTTP response codes like `4xx`/`5xx`; it only does so for network-related issues. 
The current implementation does not properly handle these non-successful responses.

# Proposed solution
Check `response.ok` to identify failed responses and handle errors accordingly.

In this way, backstage log contains:

```
backend:start: {"level":"error","message":"Error when executing workflow: Error: Response status was NOT successful when executeWorkflow(POST) fetch(http://m2k-sonataflow-infra.apps.cluster-lw48v.lw48v.sandbox2144.opentlc.com/m2k) - Details: Error id d9d5f1c0-c04e-4c03-91a2-037bc1e099be-2, Stack: No stack trace available","plugin":"orchestrator","service":"backstage","timestamp":"2024-09-20 11:18:24"}
```

# Jira issue
Fix [FLPATH-1032](https://issues.redhat.com/browse/FLPATH-1032)

### DIFF
--- a/plugins/orchestrator-backend/src/service/SonataFlowService.ts
+++ b/plugins/orchestrator-backend/src/service/SonataFlowService.ts
@@ -38,9 +38,11 @@ export class SonataFlowService {
         );
         return json;
       }
-      const responseStr = JSON.stringify(response);
-      this.logger.error(
-        `Response was NOT okay when fetch(${urlToFetch}). Received response: ${responseStr}`,
+      throw new Error(
+        this.createPrefixFetchErrorMessage(
+          urlToFetch,
+          'fetchWorkflowInfoOnService',
+        ),
       );
     } catch (error) {
       this.logger.error(`Error when fetching workflow info: ${error}`);
@@ -104,15 +106,21 @@ export class SonataFlowService {
         ? `${args.serviceUrl}/${args.definitionId}?businessKey=${args.businessKey}`
         : `${args.serviceUrl}/${args.definitionId}`;
 
-      const result = await fetch(urlToFetch, {
+      const response = await fetch(urlToFetch, {
         method: 'POST',
         body: JSON.stringify(args.inputData),
         headers: { 'content-type': 'application/json' },
       });
 
-      const json = await result.json();
-      this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
-      return json;
+      if (response.ok) {
+        const json = await response.json();
+        this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
+        return json;
+      }
+      const res = await response.json();
+      throw new Error(
+        `${this.createPrefixFetchErrorMessage(urlToFetch, 'executeWorkflow', 'POST')} - Details: ${res?.details}, Stack: ${res?.stack}`,
+      );
     } catch (error) {
       this.logger.error(`Error when executing workflow: ${error}`);
     }
@@ -218,5 +226,13 @@ export class SonataFlowService {
       this.logger.error(`Error when updating instance input data: ${error}`);
     }
     return false;
+  }
+
+  public createPrefixFetchErrorMessage(
+    urlToFetch: string,
+    functionName: string,
+    httpMethod = 'GET',
+  ): string {
+    return `Response status was NOT successful when ${functionName}(${httpMethod}) fetch(${urlToFetch})`;
   }
 }
--- a/plugins/orchestrator-backend/src/service/SonataFlowService.ts
+++ b/plugins/orchestrator-backend/src/service/SonataFlowService.ts
@@ -39,10 +39,7 @@ export class SonataFlowService {
         return json;
       }
       throw new Error(
-        this.createPrefixFetchErrorMessage(
-          urlToFetch,
-          'fetchWorkflowInfoOnService',
-        ),
+        await this.createPrefixFetchErrorMessage(urlToFetch, response),
       );
     } catch (error) {
       this.logger.error(`Error when fetching workflow info: ${error}`);
@@ -117,9 +114,8 @@ export class SonataFlowService {
         this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
         return json;
       }
-      const res = await response.json();
       throw new Error(
-        `${this.createPrefixFetchErrorMessage(urlToFetch, 'executeWorkflow', 'POST')} - Details: ${res?.details}, Stack: ${res?.stack}`,
+        `${await this.createPrefixFetchErrorMessage(urlToFetch, response, 'POST')}`,
       );
     } catch (error) {
       this.logger.error(`Error when executing workflow: ${error}`);
@@ -228,11 +224,30 @@ export class SonataFlowService {
     return false;
   }
 
-  public createPrefixFetchErrorMessage(
+  public async createPrefixFetchErrorMessage(
     urlToFetch: string,
-    functionName: string,
+    response: Response,
     httpMethod = 'GET',
-  ): string {
-    return `Response status was NOT successful when ${functionName}(${httpMethod}) fetch(${urlToFetch})`;
+  ): Promise<string> {
+    const res = await response.json();
+    const errorInfo = [];
+    let errorMsg = `Request ${httpMethod} ${urlToFetch} failed with: StatusCode: ${response.status}`;
+
+    if (response.statusText) {
+      errorInfo.push(`StatusText: ${response.statusText}`);
+    }
+    if (res?.details) {
+      errorInfo.push(`Details: ${res?.details}`);
+    }
+    if (res?.stack) {
+      errorInfo.push(`Stack: ${res?.stack}`);
+    }
+    if (errorInfo.length > 0) {
+      errorMsg += ` ${errorInfo.join(', ')}`;
+    } else {
+      errorMsg += ' Unexpected error';
+    }
+
+    return errorMsg;
   }
 }
--- a/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
+++ b/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
@@ -1,5 +1,7 @@
 import { LoggerService } from '@backstage/backend-plugin-api';
 
+import { WorkflowExecutionResponse } from '@janus-idp/backstage-plugin-orchestrator-common';
+
 import { DataIndexService } from './DataIndexService';
 import { SonataFlowService } from './SonataFlowService';
 
@@ -116,20 +118,39 @@ describe('SonataFlowService', () => {
     const definitionId = 'workflow-123';
     const urlToFetch = `${serviceUrl}/${definitionId}`;
     const functionName = 'executeWorkflow';
+    const inputData = { var1: 'value1' };
+
+    const setupTest = (responseConfig: {
+      ok: boolean;
+      status?: number;
+      json: any;
+    }) => {
+      const mockResponse = {
+        ok: responseConfig.ok,
+        status: responseConfig.status || (responseConfig.ok ? 200 : 500),
+        statusText: responseConfig.ok ? 'OK' : 'Internal Server Error',
+        json: jest.fn().mockResolvedValue(responseConfig.json),
+      };
+      global.fetch = jest.fn().mockResolvedValue(mockResponse as any);
+    };
+
+    const runErrorTest = async (
+      jsonResponse: any,
+    ): Promise<WorkflowExecutionResponse | undefined> => {
+      setupTest({ ok: false, json: jsonResponse });
+      return await sonataFlowService.executeWorkflow({
+        definitionId,
+        serviceUrl,
+        inputData,
+      });
+    };
 
     beforeEach(() => {
       jest.clearAllMocks();
     });
     it('should return workflow execution response when the request is successful', async () => {
       // Given
-      const mockResponse = {
-        ok: true,
-        status: 200,
-        json: jest
-          .fn()
-          .mockResolvedValue({ id: definitionId, status: 'completed' }),
-      };
-      global.fetch = jest.fn().mockResolvedValue(mockResponse as any);
+      setupTest({ ok: true, json: { id: definitionId, status: 'completed' } });
 
       // When
       const result = await sonataFlowService.executeWorkflow({
@@ -159,14 +180,7 @@ describe('SonataFlowService', () => {
     it('should include businessKey in the URL if provided', async () => {
       // Given
       const businessKey = 'key-123';
-      const inputData = { var1: 'value1' };
-      const mockResponse = {
-        ok: true,
-        json: jest
-          .fn()
-          .mockResolvedValue({ id: definitionId, status: 'completed' }),
-      };
-      global.fetch = jest.fn().mockResolvedValue(mockResponse as any);
+      setupTest({ ok: true, json: { id: definitionId, status: 'completed' } });
 
       // When
       const result = await sonataFlowService.executeWorkflow({
@@ -189,22 +203,8 @@ describe('SonataFlowService', () => {
     });
 
     it('should log an error and return undefined when the fetch response is not ok without extra info', async () => {
-      // Given
-      const mockResponse = {
-        ok: false,
-        status: 500,
-        statusText: 'Internal Server Error',
-        json: jest.fn().mockResolvedValueOnce({}),
-      };
-      global.fetch = jest.fn().mockResolvedValue(mockResponse as any);
-      const inputData = { var1: 'value1' };
-
       // When
-      const result = await sonataFlowService.executeWorkflow({
-        definitionId,
-        serviceUrl,
-        inputData,
-      });
+      const result = await runErrorTest({});
 
       // Then
       expect(fetch).toHaveBeenCalledWith(urlToFetch, {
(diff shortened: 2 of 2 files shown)


plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -237,4 +241,8 @@ export class SonataFlowService {
     }
     return false;
   }
+
+  public getFetchErrorMessage(urlToFetch: string, responseStr: string): string {
+    return `Response was NOT okay when fetch(${urlToFetch}). Received response ${responseStr}`;
```
gabriel-farache, line 246: Response was NOT okay => Response status was NOT successful

gciavarrini: Addressed

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -237,4 +241,8 @@ export class SonataFlowService {
     }
     return false;
   }
+
+  public getFetchErrorMessage(urlToFetch: string, responseStr: string): string {
+    return `Response status was NOT successful when fetch(${urlToFetch}). Received response ${responseStr}`;
```
gabriel-farache, line 246: Still wondering about `fetch(${urlToFetch})`; what if the same URL is used but with different method, this can get confusing.
Similarly, do we want to output the body of a POST that was used when the request fails? It can be dangerous is sensitive information are in the body but it may come in handy to investigate when something goes wrong.
Maybe with DEBUG log level the `fetch` gives more info so let's not include the body, but the method can be useful, i.e: `Response status was NOT successful when $ {method} fetch(${urlToFetch})`

gciavarrini: Agree. What about it now?

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -104,15 +103,18 @@ export class SonataFlowService {
...
 
-      const json = await result.json();
-      this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
-      return json;
+      if (response.ok) {
+        const json = await response.json();
+        this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
+        return json;
+      }
+      throw new Error(this.getFetchErrorMessage(urlToFetch, 'executeWorkflow'));
```
gabriel-farache, line 117: I would have used `POST ` instead of `executeWorkflow` as the REST method is POST

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -38,9 +38,8 @@ export class SonataFlowService {
         );
         return json;
       }
-      const responseStr = JSON.stringify(response);
-      this.logger.error(
-        `Response was NOT okay when fetch(${urlToFetch}). Received response: ${responseStr}`,
+      throw new Error(
+        this.getFetchErrorMessage(urlToFetch, 'fetchWorkflowInfoOnService'),
```
gabriel-farache, line 42: I would have used `GET ` instead of `fetchWorkflowInfoOnService`

gabriel-farache (commented):
Almost there

batzionb:
From my experience, we should not wrap the fetch command with try and catch, but just run it and throw the error and pass the original error from sonata to the router.
Otherwise we are loosing information and it's much harder to debug.

plugins/orchestrator-backend/src/service/SonataFlowService.ts
gciavarrini: I'm ok with that, but not sure if the context of this PR is the right one. I think it's more a refectory activity and it should be "extended" to other functions in this file. WDYT?

plugins/orchestrator-backend/src/service/SonataFlowService.ts
gciavarrini: I'm ok with that, but not sure if the context of this PR is the right one. I think it's more a refectory activity and it should be "extended" to other functions in this file. WDYT?

gabriel-farache: ok, let's go with that, please create a follow up issue :)

gabriel-farache:
> From my experience, we should not wrap the fetch command with try and catch, but just run it and throw the error and pass the original error from sonata to the router.
Otherwise we are loosing information and it's much harder to debug.

@batzionb yeah, would be nice if the error bubbled up to the router, this would avoid useless code (try - catch) but this would require a refactor of the whole file/plugin as it appears it's done like that in many places

Here, the goal is to actually check the status code of the request and throw an error when the status is not OK: when the execute fails (status not ok), just after the execution, the code tries to fetch the workflow instance by using the id of the response but there is no `id` in the response as the execution fails and we see a weird error. Actually, this PR is solving https://issues.redhat.com/browse/FLPATH-1595 too

@gciavarrini maybe https://issues.redhat.com/browse/FLPATH-1697 could be a refactoring of the whole file or even plugin to remove the try - catch and let the error bubble up to the route?

gciavarrini:
> @gciavarrini maybe https://issues.redhat.com/browse/FLPATH-1697 could be a refactoring of the whole file or even plugin to remove the try - catch and let the error bubble up to the route?

@gabriel-farache for that, i think we need to create a separate epic and prioritize it: it's a time-consuming task so a story is not enough. It also requires writing unit tests.

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: I actually think executeWorkflow is a good name for the function. It explains exactly what this function does. The POST is an implementation detail

plugins/orchestrator-backend/src/service/SonataFlowService.ts
ydayagi: I am in favor of putting the name of the operation and not the HTTP method. But...it is easy to put both: 'fetchWorkflowInfoOnService (GET)'. The more info the better

plugins/orchestrator-backend/src/service/SonataFlowService.ts
ydayagi: I am also in favor of putting the name of the operation and not the HTTP method. But...it is easy to put both: 'executeWorkflow (POST)'. The more info the better

gciavarrini:
@gabriel-farache @batzionb I updated the PR description to clarify the PR. 
PTAL

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -104,15 +106,21 @@ export class SonataFlowService {
...
-      this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
-      return json;
+      if (response.ok) {
+        const json = await response.json();
+        this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
+        return json;
+      }
+      const res = await response.json();
+      throw new Error(
+        `${this.createPrefixFetchErrorMessage(urlToFetch, 'executeWorkflow', 'POST')} - Details: ${res?.details || 'No details provided'}, Stack: ${res?.stack || 'No stack trace available'}`,
```
gabriel-farache, line 122: you don't need to do `Stack: ${res?.stack || 'No stack trace available'}`,` leaving the empty string or object is fine too, if there is nothing in it, the user will understand there is no stack trace
Or simply do not add the `Stack` in the error message if there is none

gciavarrini: I added for more clarity, I think `No stack trace/details available` is clearer than empty.
But I can remove it

gabriel-farache: Why saying something is not there? If it's not there, it's not there, no need to add "noise" (and use some bytes)
From an UI perspective, I get it why you would print something like that, but this error will be catch and used internally before being displayed to the end user I think :)

gciavarrini: Can you please check if it looks ok now?

gabriel-farache (approved):
noice!

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -104,15 +106,21 @@ export class SonataFlowService {
...
+        const json = await response.json();
+        this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
+        return json;
+      }
+      const res = await response.json();
+      throw new Error(
+        `${this.createPrefixFetchErrorMessage(urlToFetch, 'executeWorkflow', 'POST')} - Details: ${res?.details}, Stack: ${res?.stack}`,
+      );
     } catch (error) {
       this.logger.error(`Error when executing workflow: ${error}`);
```
batzionb, line 125: if we are already making this change let's also remove the try and catch
in case of network error it goes to the catch

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -104,15 +106,21 @@ export class SonataFlowService {
...
-      const json = await result.json();
-      this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
-      return json;
+      if (response.ok) {
+        const json = await response.json();
+        this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
+        return json;
+      }
+      const res = await response.json();
+      throw new Error(
```
batzionb, line 121: by chance it reproduced for me today during development with this response:

Response {
backend:start:   status: 400,
backend:start:   statusText: 'Bad Request',
backend:start:   headers: Headers { 'content-type': 'application/json', 'content-length': '271' },
backend:start:   body: ReadableStream { locked: false, state: 'readable', supportsBYOB: true },
backend:start:   bodyUsed: false,
backend:start:   ok: false,
backend:start:   redirected: false,
backend:start:   type: 'basic',
backend:start:   url: 'http://localhost:8899/spring-boot-backend'
backend:start: }

so sometimes the data is in statusText

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -219,4 +227,12 @@ export class SonataFlowService {
     }
     return false;
   }
+
+  public createPrefixFetchErrorMessage(
+    urlToFetch: string,
+    functionName: string,
+    httpMethod = 'GET',
+  ): string {
+    return `Response status was NOT successful when ${functionName}(${httpMethod}) fetch(${urlToFetch})`;
```
batzionb, line 236: I think it's better to say
Request ${httpMetdho} ${urlToFetch} failed with:

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -104,15 +106,21 @@ export class SonataFlowService {
...
-      this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
-      return json;
+      if (response.ok) {
+        const json = await response.json();
+        this.logger.debug(`Execute workflow result: ${JSON.stringify(json)}`);
+        return json;
+      }
+      const res = await response.json();
+      throw new Error(
+        `${this.createPrefixFetchErrorMessage(urlToFetch, 'executeWorkflow', 'POST')} - Details: ${res?.details}, Stack: ${res?.stack}`,
```
batzionb, line 122: I think many times you won't have the stack and the details and the error message will not be user friendly

batzionb: Also better separate the whole generation of error message, we will need it for other requests as well

batzionb: Also I'm not sure you can access details directly. You might need to do res.json().details
Is there a reproduction?

plugins/orchestrator-backend/src/service/SonataFlowService.ts
gciavarrini: @gabriel-farache do you agree?

plugins/orchestrator-backend/src/service/SonataFlowService.ts
gciavarrini: @batzionb

> I think many times you won't have the stack and the details and the error message will not be user friendly

please see this discussion https://github.com/janus-idp/backstage-plugins/pull/2169#discussion_r1768742981

> Also better separate the whole generation of error message, we will need it for other requests as well

I'll adress this

> Also I'm not sure you can access details directly. You might need to do res.json().details
Is there a reproduction?

Can you clarify? 
First I've this `const res = await response.json();` and then ` Details: ${res?.details}`, what's wrong?
I verified it both with unit test and on the cluster, I didn't see any issue.

plugins/orchestrator-backend/src/service/SonataFlowService.ts
gabriel-farache: I would prefer to keep the consistency across at least the file with those try-catch
Could this be in a follow-up PR/jira issue?

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: it works on the cluster with res?.details

batzionb: if that's the case then no need for 
const res = await response.json();

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: anyhow it's not consistent now
I think better we have one example of proper error handling that we can refer to in future PRs

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: sorry didn't notice you had
const res = await response.json();
so comment irrelevant

batzionb:
@gabriel-farache @gciavarrini 
There are too many discussions on the format of the error message
I want to get this right because we will reuse this
I suggest:

prefix:
Request {httpMethod} {url} failed with: 
If there's status text add the statusText
If there's details add the details
If there's stack trace add it as well
If there's none of the above put Unexpected error

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: this is easy to reproduce
You call the API with some missing required properties

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: Sorry I missed that you already have
const res = await response.json();
You can ignore this comment

batzionb (approved):

Issue #2025, opened 2024-08-07
Support wildcards in RBAC plugin casbin rule evaluation


rgolangh:
### What do you want to improve?
The definition of RBAC rules 

### What is the current behavior?
Currently there is no way for plugins to create a rule that can automatically address objects by wildcars.
The orchestrator plugin is using this role definition and it allow or deny a specific action for all the workflows object:
```
p, role:default/workflowAdmin, orchestrator.workflow.execute, use, allow
```

### What will the new behavior be?

We would like to change the rule to deny a specific workflow ID but allowing all
```
p, role:default/workflowAdmin, orchestrator.workflow.execute, use, allow
p, role:default/workflowAdmin, orchestrator.workflow.ABC, use, deny
```

And for that to work we need to implement the wildcard behaviour in our plugin because those are 2 different objects, while
with wildcards it can be: 
```
p, role:default/workflowAdmin, orchestrator.workflow.*, use, allow
p, role:default/workflowAdmin, orchestrator.workflow.ABC, use, deny
```
and the permission check is with a single object `orchestrator.workflow.XYZ`  (XYZ is a different workflow ID for that matter and should result an allow)

There is another scenario where the wildcards can be added to the `subject` section and with that capture 
a role in a way which is not only by using groups, but this is not my main interest, but was posted on #1994 

I think this is doable with casbin https://casbin.org/docs/rbac-with-pattern#use-pattern-matching-in-rbac

pkliczewski:
fyi @AndrienkoAleksandr @PatAKnight

rgolangh:
What I am looking here is patterns on the `policy` and not the `role` definition, and I am not sure this is supported. 
Perhaps RBAC with patterns will not help here.

rgolangh:
I took the model and worked with this [casbin online editor](https://editor.casbin.org/#7ZYJ5GMTC
) and used the `regexMatch` function to match object on a request and I was able to get the behaviour I wanted.

So given this policy:
```
p, role:default/workflowAdmin, orchestrator.workflows, read, allow

p, role:default/workflowViewer, orchestrator.workflows, read, deny
p, role:default/workflowViewer, orchestrator.workflows.infra, read, deny

g, user:default/guest, role:default/workflowViewer
g, user:development/admin, role:default/workflowAdmin
```

This requests for the admin user are allowed:
```
user:development/admin, orchestrator.workflows.foo, read
user:development/admin, orchestrator.workflows.infra, read
```

[...]

PR #2171, opened 2024-09-13
feat(rbac): patterns on rbac policy objects

rgolangh:
Motivation:
There is no way to capture multiple objects in a single policy today and
match against it. The outcome is that policies for basic permission,
where the objects are not catalog entities, are not fine grained. Either
you have permission on an action on all the category, or not at all.
For example the workflows that the orchestrator plugin exposes are not
a catalog entities. They only have a workflow ID the plugin exposes and
there is no way today to deny or allow some of the workflows usage.

I.e, this is the permission a workflow have today for read:

    orchestrator.workflow.read, read, allow

Modification:
The only modification for the rbac plugin is to use a regexp match:

    [matchers]
    m = g(r.sub, p.sub) && regexMatch(r.obj, p.obj) && r.act == p.act

Result:
With this change we have the permission be matched by a regexp:

    orchestrator.workflows ,       read, allow  // catches all workflow
    orchestrator.workflows.banned ,read, deny   // catches banned workflow

Resolves: https://github.com/janus-idp/backstage-plugins/issues/2025


rgolangh:
need to figure out why it does not work with the tempEnforcer

batzionb:
@PatAKnight 
@rgolangh implemented orchestrator workflow authorization by creating a role for each workflow instead of using conditions.
This requires a small change in RBAC plugin.
WDYT?

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -352,7 +353,10 @@ function setupInternalRoutes(
 
       const decision = await authorize(
         _req,
-        orchestratorWorkflowReadPermission,
+        createPermission({
```
batzionb, line 356: how would it work if the user put the global workflow viewing permission but not fine grained ones?

rgolangh: what exactly is the global one?

rgolangh: a 'global' can be set by the policy author, because now the permission object is matched by a regexp match
so if the policy says 'orchestrator.workflow' then every string that starts with 'orchestrator.workflow' is matched, unless there is a deny rule that matches.

plugins/orchestrator/docs/rbac-policy.csv
```diff
@@ -1,11 +1,15 @@
 p, role:default/workflowViewer, orchestrator.workflowInstances.read, read, allow
```
batzionb, line 1: I'm trying to understand how the policies will be used:
I think the most simple use case is that admins can view and run all workflows
And a group, let's say "developers-a" the admins will want to restrict to use workflow a
And another group, let's say "developers-b" the admins will want to restrict to use workflow b
How will this be done?

plugins/orchestrator-backend/src/service/router.ts
rgolangh: we would need to remove the current common permissions, they will not be in use and only leave the ones who are not workflow specific

batzionb: > we would need to remove the current common permissions, they will not be in use and only leave the ones who are not workflow specific

WDYM?

rgolangh:
> @PatAKnight @rgolangh implemented orchestrator workflow authorization by creating a role for each workflow instead of using conditions. This requires a small change in RBAC plugin. WDYT?

just to make sure there is no confusion, it is not a role per workflow. it is a policy that can match against multiple objects. The object identifier in the orchestrator case is the a string of the form `orchestartor.workflow.$WORKFLOW_ID`, for the route endpoint /workflows/$WORKFLOW_ID.

plugins/rbac-backend/src/service/enforcer-delegate.ts
```diff
@@ -434,7 +434,11 @@ export class EnforcerDelegate implements RoleEventEmitter<RoleEvents> {
     const filter = [];
     if (roles.length > 0) {
       roles.forEach(role => {
-        filter.push({ ptype: 'p', v0: role, v1: resourceType, v2: action });
+        //filter.push({ ptype: 'p', v0: role, v1: resourceType, v2: action });
+        // this commented filter is not designed to work with resourceType which is dynamically constructed
+        // using the resouece id. The switch to mathing policies on regexMatch on the object (here is is resourceType under v1 key) means we can't load policies by resourceType because the query to the repository will not find it. e.g query where orchestrator.workflows.abc where the policy has the object orchestrator.workflows will return empty
+        // We can expand the enforce method to conditionally use the former filter in case a policy for a catalog entity resource is used where the object name represents the resouce type
+        filter.push({ ptype: 'p', v0: role, v2: action });
```
rgolangh, line 441: @PatAKnight this is the change in the filter I had to make in order for the policies to be loaded.

app-config.yaml
```diff
@@ -18,7 +18,10 @@ backend:
     # Uncomment the following host directive to bind to specific interfaces
     # host: 127.0.0.1
   csp:
```
batzionb, line 20: please remove the changes in app-config and packages/*

rgolangh:
@AndrienkoAleksandr @PatAKnight please review

plugins/rbac-backend/src/service/enforcer-delegate.ts
```diff
@@ -434,7 +434,11 @@ export class EnforcerDelegate implements RoleEventEmitter<RoleEvents> {
     const filter = [];
     if (roles.length > 0) {
       roles.forEach(role => {
-        filter.push({ ptype: 'p', v0: role, v1: resourceType, v2: action });
+        // filter.push({ ptype: 'p', v0: role, v1: resourceType, v2: action });
+        // this commented filter is not designed to work with resourceType which is dynamically constructed
+        // using the resouece id. The switch to mathing policies on regexMatch on the object (here is is resourceType under v1 key) means we can't load policies by resourceType because the query to the repository will not find it. e.g query where orchestrator.workflows.abc where the policy has the object orchestrator.workflows will return empty
+        // We can expand the enforce method to conditionally use the former filter in case a policy for a catalog entity resource is used where the object name represents the resouce type
+        filter.push({ ptype: 'p', v0: role, v2: action });
```
PatAKnight, line 441: This change will lead to a negative impact on the performance of the RBAC backend plugin. It is hard to say how much of an impact, but reducing one of the filtering options will result in more matches that will have to be evaluated.

plugins/orchestrator-common/src/permissions.ts
```diff
@@ -22,12 +22,7 @@ export const orchestratorWorkflowReadPermission = createPermission({
 });
 
 export const orchestratorWorkflowExecutePermission = createPermission({
-  name: 'orchestrator.workflow.execute',
-  attributes: {},
-});
-
-export const orchestratorWorkflowInstanceAbortPermission = createPermission({
-  name: 'orchestrator.workflowInstance.abort',
+  name: 'orchestrator.workflows',
```
PatAKnight, line 25: As mentioned by @batzionb, couldn't conditional rules work instead? I don't have much experience defining them. But they could work for what you are trying to achieve. 

From my basic understanding, you wound need to define a resource type, in this case something like `orchestrator-workflows` .Next, you would need to define your conditional rules, something like `hasWorkflowID`. And finally you would want to update your router to use the conditional rules.

I have not tried them out personally but I can supply some resources and a few plugins that currently use them at the moment. 

[Adding a resource permission check](https://backstage.io/docs/permissions/plugin-authors/03-adding-a-resource-permission-check), this includes information on support for conditional decisions

[Playlist plugin](https://github.com/backstage/community-plugins/blob/main/workspaces/playlist/plugins/playlist-backend/src/permissions/conditions.ts)

[Catalog plugin](https://github.com/backstage/backstage/tree/master/plugins/catalog-backend/src/permissions)

[Scaffolder plugin](https://github.com/backstage/backstage/blob/f899a829e09d88a2e37bcd93c9fdca919878a6b8/plugins/scaffolder-backend/src/service/rules.ts)

plugins/orchestrator/docs/rbac-policy.csv
rgolangh: in this case the policy should have the exact name of the workflow, so there will be no match for any other(*):

```
p, role:default/groupA, orchestrator.workflows.workflow-a, read, allow

p, role:default/groupB, orchestrator.workflows.workflow-b, read, allow
```

The regexp match will not match other workflows names againt this policy, hence it will be denied 

 `* One thing to look out for is that since we're using a regexp match on the name of the workflow then a workflow with the name 'workflow-abc' will also be allowed.`

rgolangh:
> Sorry I am just now getting to this.
> 
> I left a couple of comments in the PR but also wanted to include some thoughts as well.
> 
> I realize now that I may have misunderstood what you were trying to accomplish with using pattern matching on permission policies. I was under the impression that you were actually trying to reduce the need to define many permission policies that all have a common resource type. An example in my head is something like the following:
> 
> ```
> # Assume role:default/example is already defined
> 
> # catalog-entity resource type has three different permissions associated with 
> # it so defining permission policies would look something like this
> p, role:default/example, catalog-entity, read, allow
> p, role:default/example, catalog-entity, delete, allow
> p, role:default/example, catalog-entity, update, allow
> 
> # With the policy matching that I was thinking about it would instead match based 
> # on the resource type with the wildcard on the action itself

wildcarding the action is useful on its own for composing roles. But yes, this is not what I was using it for.

> # thus granting read, update, and delete
> p, role:default/example, catalog-entity, *, allow
> ```
> 
> I am not certain at the moment if this is possible though.

The main difference in the orchestrator permission usage is that we don't have a resourceType, but instead we use basic or named permission since we don't have our object in the catalog.

Correct me if I'm wrong, conditions could not be declared on the policy file itself right? you have to use the rbac ui to add those?

AndrienkoAleksandr:
> Correct me if I'm wrong, conditions could not be declared on the policy file itself right? you have to use the rbac ui to add those?

we introduced separated external file for conditional policies for RHDH 1.3: https://github.com/janus-idp/backstage-plugins/blob/ef9ee685be72508eb6c5b174da2565076515cb56/plugins/rbac-backend/README.md#configuring-conditional-policies-via-file

rgolangh:
given the feedback I'm closing this PR for now. The future work might use conditions instead or just reopen and continue with this PR. I leave this to @batzionb

PR #2178, opened 2024-09-16
fix(bulk-import): handle Location entities created from other sources like the GH discovery as Import candidates [RHIDP-4021]

rm3l:
**What does this PR do / why we need it:**

This PR makes sure to also include existing Location entities (like the ones that might have been added through the GH auto-discovery plugin) as potential Import jobs.
And, to make sure the bulk import backend does not return any kind of Locations, it returns only the repos that are accessible from the configured GH integrations.

**Which issue(s) this PR fixes:**
https://issues.redhat.com/browse/RHIDP-4021

This fixes the following issues:
- adding any public repos in the app-config.yaml would make it be shown as an Import job, even if that repo is not accessible from any of the configured GH integrations
- repos auto-added by the GH discovery plugin would not be shown as Import jobs, leading users to try to bulk-import them (they are shown in the Bbulk Import pages because they are accessible from the configured GH integrations).

**PR acceptance criteria:**

- [x] Unit tests

- [ ] Integration tests

- [ ] Documentation 

**How to test changes / Special notes to the reviewer:**

- In your app-config.yaml, add a mix of Locations accessible from your GH integrations, and some others that are public, e.g.:
```yaml
integrations:
  # TODO: configure your GH integrations
  # https://backstage.io/docs/integrations/github/locations/#configuration
  github: []

catalog:
  locations:
    - type: url
      # Valid Import location pattern,
      # but should be ignored because 'janus-qe/acr-catalog' is not accessible from the configured GH integrations
      target: https://github.com/janus-qe/acr-catalog/blob/main/catalog-info.yaml

    - type: url
      # Valid Import location pattern. Repo is accessible from the configured GH integrations
      # => should be returned by the `GET /imports` endpoint along with the right status
      target: https://github.com/lemra-org-ent-2/expert-octo-guide/blob/main/catalog-info.yaml
```

- If you have the GH Discovery plugin enabled:

<details>
<summary>enabling the GH Discovery dynamic plugin</summary>

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: dynamic-plugins-rhdh
data:
  dynamic-plugins.yaml: |
    includes:
      - dynamic-plugins.default.yaml
    plugins:
	  # TRUNCATED
      - package: './dynamic-plugins/dist/backstage-plugin-catalog-backend-module-github-dynamic'
        disabled: false
        pluginConfig:
          catalog:
            providers:
              github:
                myorg:
                  organization: '${GH_ORG}'
                  schedule:
                    # supports cron, ISO duration, "human duration" (used below)
                    frequency: { minutes: 5}
                    # supports ISO duration, "human duration (used below)
                    timeout: { minutes: 3}
                    initialDelay: { seconds: 15}
```

</details>

With the examples above, this is how it looks like:

**Before**

image

- `janus-qe/acr-catalog` would be returned in the list of Imports, even if it is not accessible from any of my configured GH integrations. This makes it confusing since this repo will never be displayed in the list of organizations/repositories from the Bulk Import pages.
- none of the repos monitored by the GitHub auto-discovery plugin would show up in this list and their Import status won't be accurate, even though they are accessible from my configured GH integrations. The org watched is `lemra-org`.

image

**With the changes here**

Screenshot from 2024-09-16 10-11-29

- `janus-qe/acr-catalog` is no longer being returned in the list of Imports, since it is not accessible from any of my configured GH integrations
- all the repos monitored by the GitHub auto-discovery plugin now show up in this list and have the right Import status, since they are accessible from my configured GH integrations. The org watched is `lemra-org`.

Screenshot from 2024-09-16 10-12-40

/cc @subhashkhileri

### DIFF
--- a/plugins/bulk-import-backend/src/helpers/catalogInfoGenerator.ts
+++ b/plugins/bulk-import-backend/src/helpers/catalogInfoGenerator.ts
@@ -16,6 +16,7 @@
 import { PluginEndpointDiscovery } from '@backstage/backend-common';
 import { AuthService, DiscoveryService } from '@backstage/backend-plugin-api';
 import { CatalogApi } from '@backstage/catalog-client';
+import { LocationEntity } from '@backstage/catalog-model';
 import type { Config } from '@backstage/config';
 
 import gitUrlParse from 'git-url-parse';
@@ -134,12 +135,13 @@ ${jsYaml.dump(generatedEntity.entity)}`,
   async listCatalogUrlLocationsById(
     config: Config,
     search?: string,
-    _pageNumber: number = DefaultPageNumber,
-    _pageSize: number = DefaultPageSize,
+    pageNumber: number = DefaultPageNumber,
+    pageSize: number = DefaultPageSize,
   ): Promise<{ id?: string; target: string }[]> {
     const result = await Promise.all([
       this.listCatalogUrlLocationsFromConfig(config, search),
       this.listCatalogUrlLocationsByIdFromLocationsEndpoint(search),
+      this.listCatalogUrlLocationEntitiesById(search, pageNumber, pageSize),
     ]);
     return result.flat();
   }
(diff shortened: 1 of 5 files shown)


PatAKnight (approved):
Tested and it worked for me

/lgtm

PR #2183, opened 2024-09-16
fix(cli): add additional shared frontend packages

gashcrumb:
This change adds a number of additional shared frontend packages to the webpack federated module configuration to ensure that MUI v5 styles are correctly applied when run as a dynamic frontend plugin.  This change fixes [RHIDP-3471](https://issues.redhat.com/browse/RHIDP-3471).

As the backstage-showcase frontend has been migrated from the deprecated MUI styles API and is using tss-react it appears the v5 class name prepending is less neccessary.  However should a dynamic plugin need some CSS that requires the presence of separate class names for v5 components the following code can be added to the dynamic plugins `index.ts` or `plugin.ts`:

```typescript
import { unstable_ClassNameGenerator as ClassNameGenerator } from '@mui/material/className';

ClassNameGenerator.configure(componentName => {
  return componentName.startsWith('v5-')
    ? componentName
    : `v5-${componentName}`;
})
```

Testing this requires using the updated CLI to build the backstage-showcase frontend, as well as a dynamic plugin that has also been built with this updated CLI.  With this fix the dynamic plugin UI MUI v5 components should have `jss-*` named classes applied to them containing the theme style settings.


jesuino (approved):
I tested the changes locally and it works correctly. Here the steps I followed:

* First I reproduced the issue with the plugin mentioned in JIRA and I could indeed observe that it does not respect the dark mode:
REPRODUCED
* After this I built Stan's CLI and made sure it was correctly being used for the plugin and also for show case
* Starting show case in dev mode I can see that it now respect the dark mode:
FIXED

kadel (approved):

PR #2184, opened 2024-09-16
fix(rbac): improve members dropdown to handle large no. of users/groups

divyanshiGupta:
Fixes: https://issues.redhat.com/browse/RHDHBUGS-73

This PR fixes this issue by limiting the initial users/groups shown to 100 and once the user starts to search the search happens accross the whole data-set (tested with 45k users) and only top 100 results are rendered, for more options users can refine their search to get to the required user/group.

https://github.com/user-attachments/assets/2ed9bc2d-3d8f-4cd7-ba48-d87fa2aae637

### DIFF
--- a/plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
+++ b/plugins/rbac/src/components/CreateRole/AddMembersForm.tsx
@@ -51,23 +51,35 @@ export const AddMembersForm = ({
       : undefined;
   };
 
-  const membersOptions: SelectedMember[] = membersData.members
-    ? membersData.members.map((member: MemberEntity, index: number) => {
-        const tag =
-          member.metadata.etag ??
-          `${member.metadata.name}-${member.kind}-${index}`;
-        return {
-          id: tag,
-          label: member.spec?.profile?.displayName ?? member.metadata.name,
-          description: getDescription(member),
-          etag: tag,
-          type: member.kind,
-          namespace: member.metadata.namespace,
-          members: getMembersCount(member),
-          ref: stringifyEntityRef(member),
-        };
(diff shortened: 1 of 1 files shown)


its-mitesh-kumar:
@divyanshiGupta Please fix the Test build fail .
otherwise I have tested it for 5k users and its working perfectly fine . 

https://github.com/user-attachments/assets/f0daf504-25cd-432a-aedc-ce348e1c7509

cc : @invincibleJai

its-mitesh-kumar:
\lgtm
Cc: @invincibleJai

karthikjeeyar:
Just a suggestion, In addition to this change adding a debounce to input change (calls setSearch method) might also help a bit to avoid filtering the list on every keystroke.

divyanshiGupta:
> Just a suggestion, In addition to this change adding a debounce to input change (calls setSearch method) might also help a bit to avoid filtering the list on every keystroke.

@karthikjeeyar I don't think debounce is really suitable in this use case because of how autocomplete sends the events back to us, it creates unintented issues like missing the character inputs randomly for the amount of debounce wait time. Check this video out with debounce applied to input handler. It just misses the inputs in between the 500ms of debounce and only sets the last character. Reducing it to 100ms improves the input but the problem still remains.


https://github.com/user-attachments/assets/d66ca481-dffc-4016-b2cb-f774885129e8

Using throttle here works better since it doesn't skip the event callbacks in between but I don't think its necesarry to use it here either. We are only filtering out of already loaded data set and not making fetch calls so eventually the amount of gain by using it is negligible.

invincibleJai (approved):
/lgtm

PR #2189, opened 2024-09-17
feat(aap)!: migrate to the new backend and remove deprecations

schultzp2020:
fixes https://issues.redhat.com/browse/RHIDP-4054

- removes legacy backend system
- removes deprecations


plugins/aap-backend/src/providers/AapResourceEntityProvider.ts
```diff
@@ -31,75 +35,79 @@ export class AapResourceEntityProvider implements EntityProvider {
...
         taskRunner = options.scheduler.createScheduledTaskRunner(
           providerConfig.schedule,
         );
-      } else if (options.schedule) {
+      } else if ('schedule' in options) {
         taskRunner = options.schedule;
       } else {
-        throw new Error(
+        throw new InputError(
           `No schedule provided neither via code nor config for AapResourceEntityProvider:${providerConfig.id}.`,
```
04kash, line 61: Since in the new backend system schedule is provided only through config and not code, would it be better to update the Error message to something like this?
```suggestion
          `No schedule provided via config for AapResourceEntityProvider:${providerConfig.id}.`,
```

schultzp2020: Good point. We could likely remove the option to add it through code entirely, since a default schedule is applied if none is provided.

AndrienkoAleksandr:
Can you also remove `/alpha` from README.md, please?

```ts
backend.add(import('@janus-idp/backstage-plugin-aap-backend/alpha'));
```

PatAKnight (approved):
/lgtm

PR #2205, opened 2024-09-18
chore(release): catch audit log 1.3.x up with showcase 1.3.x

PatAKnight:
## Description
Copies the current version of the audit log over to the 1.3.x to ensure that we get caught back up.

## Additional Notes
Audit log version in the showcase 1.3.x branch: [1.4.0](https://github.com/janus-idp/backstage-showcase/blob/1.3.x/packages/backend/package.json#L45)
Audit log version in the PR is 1.4.1, PR needs to be opened up in the showcase repo to bump this plugin one more time.

### DIFF
--- a/plugins/audit-log-node/README.md
+++ b/plugins/audit-log-node/README.md
@@ -61,6 +61,7 @@ import {
   HttpAuthService,
   LoggerService,
 } from '@backstage/backend-plugin-api';
+import { Config } from '@backstage/config';
 
 /* highlight-add-start */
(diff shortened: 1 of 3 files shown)


PatAKnight:
Closing, moved this update to the [rbac pr](https://github.com/janus-idp/backstage-plugins/pull/2201) as the test failures are due to a mismatch in rbac versions. Newer version tests are running instead. Since it is just a doc update for audit log node, combining them shouldn't be an issue.

PR #2207, opened 2024-09-18
fix(orchestrator): v2 inputSchema API gets stuck for specific workflow

ydayagi:
[FLPATH-1703](https://issues.redhat.com//browse/FLPATH-1703)
https://issues.redhat.com/browse/FLPATH-1703

### DIFF
--- a/plugins/orchestrator-backend/src/service/router.ts
+++ b/plugins/orchestrator-backend/src/service/router.ts
@@ -779,7 +779,7 @@ function setupInternalRoutes(
       }
 
       if (!definition.dataInputSchema) {
-        res.status(200).json(res);
+        res.status(200).json({});
         return;
       }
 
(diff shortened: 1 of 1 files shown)


plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -779,7 +779,7 @@ function setupInternalRoutes(
       }
 
       if (!definition.dataInputSchema) {
-        res.status(200).json(res);
+        res.status(200).json({});
```
gciavarrini, line 782: In both cases, is `200` ok? I mean, it is a successful response?

batzionb: yes it's successful
you can have a workflow without an input schema

PR #2219, opened 2024-09-20
chore: update incubator-kie-sonataflow-devmode image to latest

mareklibra:
Former image on quay is 5 months old.

### DIFF
--- a/plugins/orchestrator-common/src/constants.ts
+++ b/plugins/orchestrator-common/src/constants.ts
@@ -1,6 +1,6 @@
 // Default values for the orchestrator plugin configuration
 export const DEFAULT_SONATAFLOW_CONTAINER_IMAGE =
-  'quay.io/kiegroup/kogito-swf-devmode-nightly:main-2024-04-17';
+  'docker.io/apache/incubator-kie-sonataflow-devmode:latest';
 export const DEFAULT_SONATAFLOW_PERSISTENCE_PATH = '/home/kogito/persistence';
 export const DEFAULT_SONATAFLOW_BASE_URL = 'http://localhost';
 


plugins/orchestrator-common/src/constants.ts
```diff
@@ -1,6 +1,6 @@
 // Default values for the orchestrator plugin configuration
 export const DEFAULT_SONATAFLOW_CONTAINER_IMAGE =
-  'quay.io/kiegroup/kogito-swf-devmode-nightly:main-2024-04-17';
+  'docker.io/apache/incubator-kie-sonataflow-devmode:latest';
```
gabriel-farache, line 3: why `latest` and not `main`? I am not sure of the stability of the `latest` in comparison of the `main` but the `main`is built daily/nightly

mareklibra: I am not sure about `latest` either. But do we want to deal with "nightly instability" of a dependency we have limited control of?

mareklibra: The latest is 10 days old. Without looking at their processes, this sounds like more promising. So far we were ok with 5 months old version. But not strong opinion here, I have limited experience with this project.

gabriel-farache: > But do we want to deal with "nightly instability" of a dependency we have limited control of?

We were already using nightly but I agree that the versioning was better, right now, with the new image, everything can break over a night and we may never know the last date when all was OK as the tag is the same

It may be worthy to start a conversation with sonataflow team for them to improve this or at least give details of what is latest

batzionb: It was recommended to us to use this one
https://catalog.redhat.com/software/containers/openshift-serverless-1/logic-swf-devmode-rhel8/6614edd9d4a41fb7bf4d40dc

batzionb (approved):

PR #2223, opened 2024-09-20
fix(bulk-import): fix missing `author` field in `package.json` files of `bulk-import-{backend,common}` [RHIDP-3818]

rm3l:
Otherwise, this causes the Backend plugin to be documented as 'Community Support' instead of 'Red Hat Tech Preview'  in the dynamic plugins CSV table [1]

image

ref https://issues.redhat.com/browse/RHIDP-3818

[1] https://github.com/redhat-developer/red-hat-developers-documentation-rhdh/blob/dynamic-plugins-table13x/modules/dynamic-plugins/rhdh-supported-plugins.csv


AndrienkoAleksandr (approved):
Looks good.

PR #2229, opened 2024-09-23
fix(orchestrator): remove field 'variables' from ProcessInstanceDTO

ydayagi:
[FLPATH-1672](https://issues.redhat.com//browse/FLPATH-1672)
https://issues.redhat.com/browse/FLPATH-1672


ydayagi:
@gabriel-farache @pkliczewski @gciavarrini

gciavarrini (approved):
/lgtm

batzionb:
@ydayagi 
Please revert this PR
Two reasons:
1. the UI uses the variables field, to remove it you need to adjust the UI
2. it's removed from the API definition, but still returned by the backend, so it isn't consistent. But at least this prevents the UI from being broken

batzionb:
I see now this breaks the build of the plugin
I'll create the revert now

PR #2235, opened 2024-09-24
fix(orchestrator): remove field 'variables' from ProcessInstanceDTO

ydayagi:
[FLPATH-1672](https://issues.redhat.com//browse/FLPATH-1672)
https://issues.redhat.com/browse/FLPATH-1672


batzionb:
@mareklibra can you PTAL?

mareklibra:
Let's wait for https://github.com/janus-idp/backstage-plugins/pull/2226. We will be able to remove the `nextWorkflowSuggestions` then

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -30,7 +30,7 @@ export const mapProcessInstanceToDetails = (
   }
 
   const started = start?.toDate().toLocaleString() ?? VALUE_UNAVAILABLE;
-  const variables = parseWorkflowVariables(instance?.variables);
+  const variables = parseWorkflowVariables(instance?.workflowdata);
   const nextWorkflowSuggestions: WorkflowRunDetail['nextWorkflowSuggestions'] =
```
mareklibra, line 34: I think we can remove both lines 33 and 34. 
Can you please double check they are not used anywhere else? Git-grep suggests that they are not.

batzionb:
@ydayagi 
please notice that the API still returns the variables and the workflow data because the implementation wasn't changed only the definition
you also need to change this function, which is the source of all these duplications
https://github.com/batzionb/backstage-plugins/blob/uischemamain/plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts#L115

gciavarrini:
> @ydayagi please notice that the API still returns the variables and the workflow data because the implementation wasn't changed only the definition you also need to change this function, which is the source of all these duplications https://github.com/batzionb/backstage-plugins/blob/uischemamain/plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts#L115

@batzionb FYI there is a dedicated [FLPATH-1675](https://issues.redhat.com/browse/FLPATH-1675) for this, but I'm totally ok to fix it in this PR.

@ydayagi we have the same issue in all the functions in that file: can you update the [FLPATH-1675](https://issues.redhat.com/browse/FLPATH-1675) accordingly?

batzionb:
> > @ydayagi please notice that the API still returns the variables and the workflow data because the implementation wasn't changed only the definition you also need to change this function, which is the source of all these duplications https://github.com/batzionb/backstage-plugins/blob/uischemamain/plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.ts#L115
> 
> @batzionb FYI there is a dedicated [FLPATH-1675](https://issues.redhat.com/browse/FLPATH-1675) for this, but I'm totally ok to fix it in this PR.
> 
> @ydayagi we have the same issue in all the functions in that file: can you update the [FLPATH-1675](https://issues.redhat.com/browse/FLPATH-1675) accordingly?

If it'll be more easy to fix that whole issue separately it can be done in a separate PR

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
ydayagi: the var 'variables' is used by line 34 and that is used in the return value

mareklibra: Right. But I do not see the returned value to be used any further.

ydayagi:
> Let's wait for #2226. We will be able to remove the `nextWorkflowSuggestions` then

@mareklibra what do u mean? what should be removed?

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
ydayagi: in line 36 we use variables for setting nextWorkflowSuggestions. then we use nextWorkflowSuggestions in returned value

batzionb: and then nextWorkflowSuggestions isn't used
It needs to be removed from the return value
and from the type definition WorkflowRunDetail

ydayagi: I think I fixed it. please check

mareklibra:
It needs rebase but lgtm

batzionb (approved):

PR #2238, opened 2024-09-24
feat(lightspeed): update /v1/query to use streaming and return chunks in http response

yangcao77:
For issue: https://issues.redhat.com/browse/DEVAI-156

This PR updates the `v1/query` API to set `streaming=true` when query from llm server, and redirect the chunks into http response

for example, a post request:
```
{"conversation_id": "user1+1q2w3e4r-abcd1234",
"model": "ibm-granite-8b-code-instruct",
"serverURL": "http://localhost:7007/api/proxy/lightspeed/api",
"query": "Hello"}
```
gets response. `response.kwargs.content` will be the chunk content, `response.kwargs.id` represents a particular chat query, this will be very useful upon concurrent queries. `response.kwargs.response_metadata.finish_reason` is the end of message signal. 
Screenshot 2024-09-24 at 12 12 25 PM

frontend can use the following to get the streaming chunks:
```
      const data = {
        model: 'ibm-granite-8b-code-instruct',
        conversation_id: "user1+1q2w3e4r-abcd123",
        serverURL: serverURL,
        query: message,
      };

      for await ( const chunk of streamingFetch( () =>  this.fetchApi.fetch(`${baseUrl}/v1/query`, {
        method: 'POST',
        body: JSON.stringify(data),
        headers: {
          'Content-Type': 'application/json',
        },
      }) ) ) {
        const value = JSON.parse(JSON.stringify(chunk))
        console.log( chunk )
      }

async function* streamingFetch( fetchcall: any ) {

  const response = await fetchcall();
  // Attach Reader
  const reader = response.body.getReader();
  while (true) {
      // wait for next encoded chunk
      const { done, value } = await reader.read();
       // check if stream is done
      if (done) break;
      // Decodes data chunk and yields it
      yield (new TextDecoder().decode(value));
  }
}
```


karthikjeeyar:
Tested it, I am able to receive the data streams in the fetch call. Also I noticed that respone preview is not showing any data in the browser, probably because of missing content-type headers from server. 

Note: In Post-man client I see the response maybe it automatically handles if server isn't setting one explicitly.

image

karthikjeeyar (approved):
Response not showing in the browser,  but that is not blocking this PR to be merged. we can revisit it later.

/lgtm

PR #2247, opened 2024-09-25
fix(orchestrator): fix ProcessInstanceDTO mapping

ydayagi:
[FLPATH-1675](https://issues.redhat.com//browse/FLPATH-1675)
https://issues.redhat.com/browse/FLPATH-1675

### DIFF
--- a/plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.test.ts
+++ b/plugins/orchestrator-backend/src/service/api/mapping/V2Mappings.test.ts
@@ -61,7 +61,6 @@ describe('scenarios to verify mapToWorkflowOverviewDTO', () => {
     expect(result.category).toBe('assessment');
     expect(result.avgDurationMs).toBe(overview.avgDurationMs);
     expect(result.description).toBe(overview.description);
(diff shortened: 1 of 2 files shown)


batzionb:
This PR has a bug
The name is missing in workflows table
Actually I think the spread operator (...) is suitable for the /overview API because it maps almost one to one the WorkflowOverview to WorklowOverviewDTO
I suggest to simply revert the change in this API
image

ydayagi:
> This PR has a bug The name is missing in workflows table Actually I think the spread operator (...) is suitable for the /overview API because it maps almost one to one the WorkflowOverview to WorklowOverviewDTO I suggest to simply revert the change in this API
> 
> And nice that the unit test caught the bug :)
> 
> image

ydayagi:
> > This PR has a bug The name is missing in workflows table Actually I think the spread operator (...) is suitable for the /overview API because it maps almost one to one the WorkflowOverview to WorklowOverviewDTO I suggest to simply revert the change in this API
> > And nice that the unit test caught the bug :)
> > image

already fixed

batzionb (approved):

PR #2216, opened 2024-09-19
fix(argocd): remove usage of pf react icons

debsmita1:
**_Resolves:_**
https://issues.redhat.com/browse/RHIDP-3796

**Screenshots:**

**Deployment lifecycle**
Before
Screenshot 2024-09-19 at 1 36 07 PM

After
Screenshot 2024-09-19 at 9 38 11 PM

**Deployment Summary**
Before
Screenshot 2024-09-19 at 1 36 15 PM

After
Screenshot 2024-09-19 at 9 37 45 PM

**_Note_**
Depends on https://github.com/janus-idp/backstage-plugins/pull/1803

PR #2249, opened 2024-09-25
chore: add deprecation note to acr plugin

04kash:
[RHIDP-3546](https://issues.redhat.com/browse/RHIDP-3546): Add deprecation note to acr plugin


nickboldt:
> @nickboldt can we remove the source from main now too? I think it reduces risk of things getting out-of-sync (refs: [#2216 (comment)](https://github.com/janus-idp/backstage-plugins/pull/2216#issuecomment-2361453784)), and reduce uncessary noise from dependency updates.

yes, you can delete everything and just leave the readme. https://github.com/janus-idp/backstage-plugins/tree/main/plugins/feedback for example

PR #2263, opened 2024-09-27
fix(topology): avoid css collision on plugins

ciiay:
For [RHIDP-2388](https://issues.redhat.com/browse/RHIDP-2388)

Tested with [RHIDP-4237](https://issues.redhat.com/browse/RHIDP-4237) test data and verified that CSS collision doesn't happen with this solution.

Screen recording

https://github.com/user-attachments/assets/4b3663e8-65fd-4a3c-ac60-fbed29af0aca

### DIFF
--- a/plugins/topology/src/components/Topology/TopologyComponent.tsx
+++ b/plugins/topology/src/components/Topology/TopologyComponent.tsx
@@ -10,23 +10,60 @@ import { ModelsPlural } from '../../models';
 import { ModelsPlural as TektonModels } from '../../pipeline-models';
 import { TopologyWorkloadView } from './TopologyWorkloadView';
 
+import '@patternfly/react-core/dist/styles/base.css';
+import '@patternfly/patternfly/patternfly-theme-dark.css';
+import '@patternfly/patternfly/patternfly-charts-theme-dark.css';
+import '@patternfly/patternfly/utilities/Accessibility/accessibility.css';
 import './TopologyComponent.css';
 
 const THEME_DARK = 'dark';
 const THEME_DARK_CLASS = 'pf-v5-theme-dark';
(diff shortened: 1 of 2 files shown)


invincibleJai:
Thanks @ciiay, this looks good!

/cc @karthikjeeyar does similar fix needed in tekton as well?

karthikjeeyar:
@invincibleJai Yes, the tekton plugin css styles were affecting other plugin in the CI tab like Jenkins logo was bigger due to PF css.

plugins/topology/src/components/Topology/TopologyComponent.tsx
```diff
@@ -10,23 +10,60 @@ import { ModelsPlural } from '../../models';
...
+    const scalprumStyles = Array.from(
+      document.querySelectorAll('link[rel="stylesheet"]'),
+    ).filter(link =>
+      link.attributes
+        .getNamedItem('href')
+        ?.value?.includes('backstage-plugin-topology'),
+    );
+
+    scalprumStyles.forEach(link =>
+      savedStylesheets.add(link as HTMLLinkElement),
```
karthikjeeyar, line 41: Tested the changes, it works fine. However I have a small suggestion to replace `useEffect` with `useLayoutEffect` to add the style sheets before the browser paint, so we could make sure style always loads before the element renders in the DOM to avoid any potential flickering due to styles in future. Let me know what you think about it?

ciiay: I have updated it to use `useLayoutEffect` and tested working good. Thanks for the feedback 👍

karthikjeeyar (approved):
/lgtm

PR #1756, opened 2024-05-31
chore: replaced Logger with backstage implementation LoggerService

04kash:
- Replaced ```Logger``` with backstage implementation ```LoggerService```
-  Renamed `plugins/feedback-backend/dynamic/inedx.ts` to `plugins/feedback-backend/dynamic/index.ts`

Closes issue #1623

PR #2277, opened 2024-10-01
feat(orchestrator)!: migrate to New Backend and Fix Deprecations

mareklibra:
Fixes: [RHIDP-4230](https://issues.redhat.com//browse/RHIDP-4230)

With this patch, the `orchestrator-backend` still loads well in the backstage of this `backstage-plugins` project on the `main` branch.

When the  `orchestrator-backend-dynamic` is exported and loaded to the `backstage-showcase` of the `main` branch, the plugin loads without any error and serves the requests as expected.


plugins/orchestrator-backend/src/service/ScaffolderService.ts
```diff
@@ -1,7 +1,10 @@
-import { loggerToWinstonLogger, UrlReader } from '@backstage/backend-common';
-import { LoggerService } from '@backstage/backend-plugin-api';
-import { CatalogApi } from '@backstage/catalog-client';
-import { Config } from '@backstage/config';
+import { loggerToWinstonLogger } from '@backstage/backend-common';
```
mareklibra, line 1: This is the only reference `@backstage/backend-common`. By removing it, we can get rid of the dependency in the `package.json`. 
Still looking for the right way how to do it.

mareklibra:
Cc @schultzp2020

batzionb:
Seems like the workflow viewer is broken with this PR
It requires authentication, previously the authentication requirement was skipped
image

ydayagi (approved):

plugins/orchestrator-backend/src/service/ScaffolderService.ts
AndrienkoAleksandr: I thought We were moving away from Winston logger in our plugins. For example: https://github.com/janus-idp/backstage-plugins/pull/1756/files

mareklibra:
> It requires authentication, previously the authentication requirement was skipped

Thanks for spotting it. Newly skipping authentication in the `plugin.ts`.

plugins/orchestrator-backend/src/plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+      async init(props) {
+        const { http } = props;
+        const router = await createRouter(props);
+        http.use(router);
+        http.addAuthPolicy({
+          path: '/ping',
+          allow: 'unauthenticated',
+        });
+        http.addAuthPolicy({
```
batzionb, line 33: please add skipping /docs as well

plugins/orchestrator-backend/src/plugin.ts
```diff
@@ -0,0 +1,44 @@
...
+        scheduler: coreServices.scheduler,
+        auth: coreServices.auth,
+        httpAuth: coreServices.httpAuth,
+        http: coreServices.httpRouter,
+      },
+      async init(props) {
+        const { http } = props;
+        const router = await createRouter(props);
+        http.use(router);
+        http.addAuthPolicy({
```
batzionb, line 29: it used to be called /health
is there a reason it changed to /ping?

mareklibra: nope, mistake

mareklibra: fixed

plugins/orchestrator-backend/src/service/ScaffolderService.ts
schultzp2020: Let's stick with your approach for now until https://github.com/backstage/backstage/issues/26933 is resolved.

schultzp2020 (approved):
/approve
/lgtm

PR #2279, opened 2024-10-01
fix(keycloak): remove --no-embed-as-dependencies flag from export-dynamic command

04kash:
As per https://github.com/janus-idp/backstage-plugins/pull/2258, this PR removes the `--no-embed-as-dependencies` flag from the `export-dynamic-plugin` command for the keycloak-backend plugin. 

Fixes: [RHIDP-4148](https://issues.redhat.com/browse/RHIDP-4148)


gashcrumb (approved):
:+1:

PR #2284, opened 2024-10-02
feat(kiali): adapt kiali to the new backend

aljesusg:
Fix. https://issues.redhat.com/browse/RHIDP-4228

Adapt kiali to the new Backend


plugins/kiali-backend/package.json
```diff
@@ -45,19 +41,16 @@
   },
   "configSchema": "config.d.ts",
   "dependencies": {
-    "@backstage/backend-common": "^0.23.3",
+    "@backstage/backend-common": "^0.25.0",
```
schultzp2020, line 44:
```suggestion
```
`@backstage/backend-common` is deprecated and needs to be removed where possible.

AndrienkoAleksandr (commented):
Packages `"@backstage/theme", "react-virtualized"`, `@backstage/errors`  are unused.

Dev dependencies `"@backstage/core-app-api", "@testing-library/react", "@testing-library/user-event", "cross-fetch", "msw":` looks unused. 

`"react-ace"` is present in the both sections: dependencies and devDependencies, but I think it should be only in the "dependencies"

`"@types/regression"` should be moved from "dependencies" to "devDependencies".

aljesusg:
> Packages `"@backstage/theme", "react-virtualized"`, `@backstage/errors` are unused.
> 
> Dev dependencies `"@backstage/core-app-api", "@testing-library/react", "@testing-library/user-event", "cross-fetch", "msw":` looks unused.
> 
> `"react-ace"` is present in the both sections: dependencies and devDependencies, but I think it should be only in the "dependencies"
> 
> `"@types/regression"` should be moved from "dependencies" to "devDependencies".

Thanks for this review... I'm going to make the changes but this plugin is going to be deprecated (We are migrating this to the community repo) we are adapting th backend side to not block

plugins/kiali-backend/src/index.ts
```diff
@@ -1,3 +1,2 @@
 export * from './service/router';
 export * from './plugin';
-export * from './dynamic/index';
```
davidfestal, line 3: There should be the following line (coming from [here](https://github.com/janus-idp/backstage-plugins/pull/2284/files#diff-5dc70c80b770e0fe0a290423fdb0018b26795c0224daeb7d3f2b6366f895aeeaL17)) added instead of the removed line:

```
export { kialiPlugin as default } from './plugin'
```

Without this default export the plugin would not be found by the new backend system affect.

davidfestal (changes requested):

plugins/kiali-backend/package.json
schultzp2020: Let's stick with your approach for now until https://github.com/backstage/backstage/issues/26933 is resolved.

plugins/kiali/package.json
```diff
@@ -80,21 +76,17 @@
...
-    "@testing-library/react": "14.3.1",
-    "@testing-library/user-event": "14.5.2",
     "@types/node": "20.14.2",
     "@types/react-copy-to-clipboard": "5.0.7",
-    "cross-fetch": "4.0.0",
-    "jest-canvas-mock": "2.5.2",
-    "msw": "1.3.3",
-    "react-ace": "9.5.0"
+    "@types/regression": "^2.0.6",
+    "cross-fetch": "^4.0.0",
```
schultzp2020, line 88:
```suggestion
    "@types/regression": "2.0.6",
    "cross-fetch": "4.0.0",
```

schultzp2020 (changes requested):

schultzp2020:
@aljesusg can you please rebase?

PR #2285, opened 2024-10-02
feat(lightspeed-backend)!: migrate to the new backend and remove deprecations

dzemanov:
## General
- removes legacy backend system
- removes deprecations
## Fixes
[RHIDP-4229](https://issues.redhat.com/browse/RHIDP-4229)


plugins/lightspeed-backend/src/service/router.test.ts
```diff
@@ -1,5 +1,4 @@
-import { getVoidLogger } from '@backstage/backend-common';
-// import { mockCredentials, mockServices } from '@backstage/backend-test-utils';
+import { mockServices } from '@backstage/backend-test-utils';
 import { MockConfigApi } from '@backstage/test-utils';
```
schultzp2020, line 2: We should use mock services [here](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/lightspeed-backend/src/service/router.test.ts#L58) instead.

plugins/lightspeed-backend/src/service/types.ts
```diff
@@ -2,10 +2,10 @@ import {
   // DiscoveryService,
```
schultzp2020, line 2: Is there a reason why the types are commented out?

dzemanov: I guess @yangcao77 added these for future reference? https://github.com/janus-idp/backstage-plugins/blob/main/plugins/lightspeed-backend/src/plugin.ts#L22-L24
I can remove them if @yangcao77 is ok with it.

yangcao77: I will  those for user authentication, but feel free to remove it if you would like to.

yangcao77 (approved):
lgtm, thanks for the change to migrate to new backend

plugins/lightspeed-backend/src/service/router.test.ts
```diff
@@ -1,6 +1,4 @@
-import { getVoidLogger } from '@backstage/backend-common';
-// import { mockCredentials, mockServices } from '@backstage/backend-test-utils';
-import { MockConfigApi } from '@backstage/test-utils';
+import { mockServices } from '@backstage/backend-test-utils';
```
schultzp2020, line 1: This test file should be refactored to follow the structure of the ocm `router.test.ts` file.

plugins/lightspeed-backend/src/service/types.ts
```diff
@@ -2,10 +2,10 @@ import {
...
   // HttpAuthService,
   // UserInfoService,
-  LoggerService,
+  type LoggerService,
```
schultzp2020, line 5:
```suggestion
  LoggerService,
```
Move `type` to the top level and remove the comments.

schultzp2020 (changes requested):

plugins/lightspeed-backend/src/service/router.test.ts
```diff
@@ -85,35 +46,50 @@ jest.mock('http-proxy-middleware', () => ({
...
-  });
+  afterAll(() => server.close());
 
-  beforeEach(() => {
+  afterEach(() => {
     jest.clearAllMocks();
+    server.resetHandlers();
   });
 
+  async function startBackendServer(config?: any) {
```
schultzp2020, line 75:
```suggestion
  async function startBackendServer(config?: Record<PropertyKey, unknown>) {
```

schultzp2020 (changes requested):

schultzp2020 (approved):
/approve
/lgtm

PR #2289, opened 2024-10-03
fix(ocm-common): fix catalog info (duplicate)

jerolimov:
Fixes a copy and paste error that this two components has the same name:

https://github.com/janus-idp/backstage-plugins/blob/5a4c3df939db940ab91f065cde22b250fee4d6cf/plugins/ocm-backend/catalog-info.yaml#L5-L7

and

https://github.com/janus-idp/backstage-plugins/blob/5a4c3df939db940ab91f065cde22b250fee4d6cf/plugins/ocm-common/catalog-info.yaml#L5-L7

This should fix:

image

### DIFF
--- a/plugins/ocm-common/catalog-info.yaml
+++ b/plugins/ocm-common/catalog-info.yaml
@@ -2,9 +2,9 @@
 apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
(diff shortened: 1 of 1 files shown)


Fortune-Ndlovu (approved):
/lgtm
/approve

PR #2297, opened 2024-10-04
fix(orchestrator): avoid empty links in workflow result

mareklibra:
A link needs to contain both key and value to be rendered, otherwise its skipped on the Workflow result card.

Fixes: [FLPATH-1749](https://issues.redhat.com//browse/FLPATH-1749)


gciavarrini (approved):
lgtm

PR #2299, opened 2024-10-04
fix(cli): improve local development experience

gashcrumb:
This change changes the CLI package-dynamic-plugins command to take advantage of some dynamic plugin projects that have followed the convention of maintaining a app-config.janus-idp.yaml in the root of the plugin source directory.  When this file is present, it will be read in as the command processes dynamic plugins in the monorepo and then included in the example dynamic-plugins.yaml configuration printed out when the command successfully completes.  This change also adjusts the control flow where this is printed out to ensure it's after successful completion of the container image.

This seemed like a nice enhancement to have while working on a dynamic plugin monorepo, here's an example of how this looks on a small project with a frontend plugin that contains an app-config.janus-idp.yaml:

Screenshot from 2024-10-04 13-13-57

And then since this has been open for a bit, I've gone ahead and added a second commit that adds a `--export-to` flag to `package-dynamic-plugin`.  This flag will skip the container creation, and instead copy the exported dynamic plugins to the specified directory, along with the generated metadata file.  Depending on if `--export-to` or `--tag` is used, the configuration that is displayed will use either a path that would be relevant for rhdh-local *or* the OCI container URL.

If you look sideways this is kind of related to [RHIDP-4336](https://issues.redhat.com/browse/RHIDP-4336)

### DIFF
--- a/packages/cli/src/commands/index.ts
+++ b/packages/cli/src/commands/index.ts
@@ -161,9 +161,13 @@ export function registerScriptCommand(program: Command) {
       '--preserve-temp-dir',
       'Leave the temporary staging directory on the filesystem instead of deleting it',
     )
-    .requiredOption(
+    .option(
+      '--export-to <directory>',
+      'Export the plugins to the specified directory, skips building the container image',
+    )
+    .option(
       '-t, --tag <tag>',
-      'Tag name to use when building the plugin registry image',
+      'Tag name to use when building the plugin registry image.  Required if "--export-to" is not specified',
     )
     .option(
       '--use-docker',
--- a/packages/cli/src/commands/package-dynamic-plugins/command.ts
+++ b/packages/cli/src/commands/package-dynamic-plugins/command.ts
@@ -12,18 +12,25 @@ import { paths } from '../../lib/paths';
 import { Task } from '../../lib/tasks';
 
 export async function command(opts: OptionValues): Promise<void> {
-  const { forceExport, preserveTempDir, tag, useDocker } = opts;
-  const containerTool = useDocker ? 'docker' : 'podman';
-  // check if the container tool is available
-  try {
-    await Task.forCommand(`${containerTool} --version`);
-  } catch (e) {
+  const { exportTo, forceExport, preserveTempDir, tag, useDocker } = opts;
+  if (!exportTo && !tag) {
     Task.error(
-      `Unable to find ${containerTool} command: ${e}\nMake sure that ${containerTool} is installed and available in your PATH.`,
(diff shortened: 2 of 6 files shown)


packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -143,6 +167,31 @@ COPY . .
...
     );
+    Task.log(`Successfully built image ${tag} with following plugins:`);
+    for (const plugin of pluginRegistryMetadata) {
+      Task.log(`  ${chalk.white(Object.keys(plugin)[0])}`);
+    }
+    // print out a configuration example based on available plugin data
+    try {
+      const configurationExample = YAML.stringify({
+        plugins: pluginRegistryMetadata.map(plugin => {
+          const pluginName = Object.keys(plugin)[0];
```
jesuino, line 178: Just a very minor suggestion that can be ignored: I would use `packageName` for this variable name to keep consistency.

jesuino (dismissed):
Looks good to me, thanks for adding this!

packages/cli/src/commands/package-dynamic-plugins/command.ts
gashcrumb: yes, good idea, thanks!

packages/cli/src/commands/index.ts
```diff
@@ -161,9 +161,13 @@ export function registerScriptCommand(program: Command) {
...
     )
-    .requiredOption(
+    .option(
+      '--export-to <directory>',
+      'Export the plugins to the specified directory, skips building the container image',
+    )
+    .option(
       '-t, --tag <tag>',
-      'Tag name to use when building the plugin registry image',
+      'Tag name to use when building the plugin registry image.  Required if "--write-to" is not specified',
```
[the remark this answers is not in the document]
gashcrumb: I sure did!  I'll update that.

packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -121,6 +135,27 @@ export async function command(opts: OptionValues): Promise<void> {
             keywords,
           },
         });
+        // some plugins include configuration snippets in an app-config.janus-idp.yaml
+        const pluginConfigPath = path.join(
+          packageDirectory,
+          'app-config.janus-idp.yaml',
```
gashcrumb, line 141: Is the convention to use the organization or repo name here?  I've seen some plugins in the community plugins repo that have this file named `app-config.community-plugins.yaml`  Maybe this will need to be more flexible.

gashcrumb: I went and added a function to the first commit that will look for a few possible known files, we can always refine this later.

jesuino (approved):
LGTM!

PR #2303, opened 2024-10-04
fix(keycloak-backend): delete unused dev dependencies and update test

dzemanov:
## General

- Delete unused dev dependencies
- Use `mockServices rootConfig` instead of `ConfigReader` in tests

## Related to

[RHIDP-4056](https://issues.redhat.com/browse/RHIDP-4056)

### DIFF
--- a/plugins/keycloak-backend/package.json
+++ b/plugins/keycloak-backend/package.json
@@ -51,25 +51,15 @@
     "uuid": "^9.0.1"
   },
   "devDependencies": {
-    "@backstage/backend-app-api": "0.8.0",
     "@backstage/backend-defaults": "0.3.3",
     "@backstage/backend-test-utils": "0.4.4",
     "@backstage/cli": "0.26.11",
-    "@backstage/config": "^1.2.0",
-    "@backstage/plugin-auth-node": "0.4.17",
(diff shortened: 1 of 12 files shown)


AndrienkoAleksandr:
Optional, but I think you can use `type` for all @keycloak/keycloak-admin-client library imports. And after that you can move this library to the dev dependency.

AndrienkoAleksandr (commented):
Optional, but I think you can use `type` for all @keycloak/keycloak-admin-client library imports. And after that you can move this library to the dev dependency.

dzemanov:
> Optional, but I think you can use `type` for all @keycloak/keycloak-admin-client library imports. And after that you can move this library to the dev dependency.

Thank you for the review. You are totally right, the imports can be changed to type, I will update them. However, `@keycloak/keycloak-admin-client` can't be moved to dev deps, look at for example this [section](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/keycloak-backend/src/providers/KeycloakOrgEntityProvider.ts#L201-L209).

AndrienkoAleksandr (commented):
Looks good to me, @schultzp2020 ?

schultzp2020 (approved):
/approve
/lgtm

PR #2315, opened 2024-10-07
fix(bulk-import): updated bulk import ui to show the correct import jobs count

debsmita1:
**Resolves:**
- https://issues.redhat.com/browse/RHIDP-4067
- https://issues.redhat.com/browse/RHIDP-3868

Note: Used `@tanstack/react-query` for querying the import jobs

**Screenshots:**

When the number of rows listed is less than the number of rows selected
(https://issues.redhat.com/browse/RHIDP-3868)
Screenshot 2024-10-16 at 1 45 48 PM


rm3l (commented):
@debsmita1 I just tested this PR, and can confirm both https://issues.redhat.com/browse/RHIDP-4067 and https://issues.redhat.com/browse/RHIDP-3868 are fixed.

However, I noticed what I think could be a regression: when trying to import new repos, the "Create pull request" button remains disabled regardless of the repos selected. See the screencast attached: 


https://github.com/user-attachments/assets/333beba3-437f-455a-a1e5-6fcf3a4f457a



Do you confirm this behavior on your end?

rm3l:
> @debsmita1 I just tested this PR, and can confirm both [issues.redhat.com/browse/RHIDP-4067](https://issues.redhat.com/browse/RHIDP-4067) and [issues.redhat.com/browse/RHIDP-3868](https://issues.redhat.com/browse/RHIDP-3868) are fixed.
> 
> However, I noticed what I think could be a regression: when trying to import new repos, the "Create pull request" button remains disabled regardless of the repos selected. See the screencast attached:
> 
>  review-pr-2315-2024-10-09_11.07.04.mp4 
> Do you confirm this behavior on your end?

I can also confirm this issue does not occur on the `main` branch.

debsmita1:
> @debsmita1 I just tested this PR, and can confirm both https://issues.redhat.com/browse/RHIDP-4067 and https://issues.redhat.com/browse/RHIDP-3868 are fixed.
> 
> However, I noticed what I think could be a regression: when trying to import new repos, the "Create pull request" button remains disabled regardless of the repos selected. See the screencast attached:
> 
>  review-pr-2315-2024-10-09_11.07.04.mp4 
> Do you confirm this behavior on your end?

Fixed it. PTAL

rm3l (commented):
> > @debsmita1 I just tested this PR, and can confirm both [issues.redhat.com/browse/RHIDP-4067](https://issues.redhat.com/browse/RHIDP-4067) and [issues.redhat.com/browse/RHIDP-3868](https://issues.redhat.com/browse/RHIDP-3868) are fixed.
> > However, I noticed what I think could be a regression: when trying to import new repos, the "Create pull request" button remains disabled regardless of the repos selected. See the screencast attached:
> > review-pr-2315-2024-10-09_11.07.04.mp4
> > Do you confirm this behavior on your end?
> 
> Fixed it. PTAL

It works better now - thanks. LGTM, but I'll let frontend experts take a look as well.

plugins/bulk-import/src/components/AddRepositories/OrganizationTableRow.tsx
```diff
@@ -31,7 +31,7 @@ export const OrganizationTableRow = ({
   ).length;
 
   return (
-    <TableRow hover>
+    <TableRow hover key={data.id}>
```
jerolimov, line 34: It would surprise me if this is necessary. Normally, it's required where the .map function is called.

plugins/bulk-import/src/components/AddRepositories/RepositoryTableRow.tsx
```diff
@@ -72,23 +72,17 @@ export const RepositoryTableRow = ({
...
         <Link to={data.repoUrl || ''}>
-          <>
-            {urlHelper(data?.repoUrl || '')}
-            <OpenInNewIcon
-              style={{ verticalAlign: 'sub', paddingTop: '7px' }}
-            />
-          </>
+          {urlHelper(data?.repoUrl || '')}
+          <OpenInNewIcon style={{ verticalAlign: 'sub', paddingTop: '7px' }} />
         </Link>
```
jerolimov, line 77: When there is no repoUrl, what do we should here?

Instead of showing an empty link, we could maybe show just a dash?

```tsx
{ data.repoUrl ?
  (
    <Link .... />
  ) :
  "-"
}
```

plugins/bulk-import/src/components/AddRepositories/RepositoryTableRow.tsx
```diff
@@ -72,23 +72,17 @@ export const RepositoryTableRow = ({
...
-            <>
-              {urlHelper(data?.organizationUrl || '')}
-              <OpenInNewIcon
-                style={{ verticalAlign: 'sub', paddingTop: '7px' }}
-              />
-            </>
+            {urlHelper(data?.organizationUrl || '')}
+            <OpenInNewIcon
+              style={{ verticalAlign: 'sub', paddingTop: '7px' }}
+            />
```
jerolimov, line 85: Same here, when there is no organizationUrl, we should skip rendering `Link` and `OpenInNewIcon`

plugins/bulk-import/src/components/BulkImportPage.tsx
```diff
@@ -27,6 +28,7 @@ export const BulkImportPage = () => {
     excludedRepositories: {},
     approvalTool: ApprovalTool.Git,
   };
+  const queryClient = new QueryClient();
```
jerolimov, line 31: When you move the `queryClient` const above `export const BulkImportPage`, then it will reuse the same instance also when the user switch to another page and come back later.

Both has pro and con. Its up to you.

But when you keep it here, you should ensure that it says the same instance also when rerendering this component. You should save it then in a useState or useRef.

plugins/bulk-import/src/components/Repositories/CatalogInfoAction.tsx
```diff
@@ -123,9 +121,9 @@ const CatalogInfoAction = ({ data }: { data: AddRepositoryData }) => {
           <IconButton
             target="_blank"
             href={
-              values?.repositories[data.id]?.catalogInfoYaml?.prTemplate
+              values?.repositories?.[data.id]?.catalogInfoYaml?.prTemplate
                 ?.pullRequestUrl ||
-              values?.repositories[data.id]?.repoUrl ||
+              values?.repositories?.[data.id]?.repoUrl ||
```
jerolimov, line 126: In this case we have a button that doesn't work right? Can we extract that code (`const link = ` and then when there is no link add an disabled button with a tooltip?

```tsx
if (!link) {
  return (
    <Tooltip title="Link is missing (or whatever)">
      <IconButton ... disabled />
    <Tooltip>
  )
}
```

plugins/bulk-import/src/components/Repositories/DeleteRepositoryDialog.tsx
```diff
@@ -69,27 +70,33 @@ const DeleteRepositoryDialog = ({
   closeDialog: () => void;
 }) => {
   const classes = useStyles();
-  const [error, setError] = React.useState<string>('');
-  const [isSubmitting, setIsSubmitting] = React.useState(false);
+  const [deleteError, setDeleteError] = React.useState<string>('');
```
jerolimov, line 73: useMutation returns an error. is it maybe simpler if you use `mutationDelete.error`. I would expect you can remove this error handling.

plugins/bulk-import/src/components/Repositories/DeleteRepositoryDialog.tsx
```diff
@@ -69,27 +70,33 @@ const DeleteRepositoryDialog = ({
...
   const bulkImportApi = useApi(bulkImportApiRef);
+  const deleteRepository = async (deleteRepo: AddRepositoryData) => {
+    return await bulkImportApi.deleteImportAction(
+      deleteRepo.repoUrl || '',
+      deleteRepo.defaultBranch || 'main',
+    );
+  };
+  const mutationDelete = useMutation(deleteRepository, {
+    onSuccess: (data: ImportJobStatus | Response) => {
+      if (get(data, 'err')) {
```
jerolimov, line 83: If the own error handling is required because sometimes the API returns HTTP 200 OK and an error, when you should handle this in `deleteRepository`.

plugins/bulk-import/src/components/Repositories/DeleteRepositoryDialog.tsx
```diff
@@ -69,27 +70,33 @@ const DeleteRepositoryDialog = ({
...
+    },
+  });
   const handleClickRemove = async () => {
-    setIsSubmitting(true);
     if (!repository.repoUrl || !repository?.defaultBranch) {
-      setIsSubmitting(false);
-      setError(
-        `Unable to remove repository. ${!repository?.repoUrl ? 'Repository URL missing' : 'Repository default branch is missing'}`,
+      setDeleteError(
+        `Unable to remove repository. ${!repository?.repoUrl ? 'Repository URL missing.' : 'Repository default branch is missing.'}`,
```
jerolimov, line 96: You know that case before the user press on delete, right?

Then you can also show a static alert below and disable the error button in that case.

```tsx
const missingRepository = !repository.repoUrl || !repository?.defaultBranch;

{ missingRepository
  <Alert /> :
  null}
```

plugins/bulk-import/src/components/Repositories/DeleteRepositoryDialog.tsx
```diff
@@ -127,19 +134,21 @@ const DeleteRepositoryDialog = ({
...
+          <Alert severity="error">{deleteError}</Alert>
         </Box>
       )}
       <DialogActions style={{ justifyContent: 'left', padding: '20px' }}>
         <Button
           variant="contained"
           className={`${classes.deleteButton} ${classes.button}`}
           onClick={() => handleClickRemove()}
-          disabled={isSubmitting}
+          disabled={mutationDelete.isLoading || !!deleteError}
```
jerolimov, line 147:
```suggestion
          disabled={missingRepository || mutationDelete.isLoading || mutationDelete.isError}
```

plugins/bulk-import/src/components/Repositories/RepositoriesList.tsx
```diff
@@ -1,48 +1,49 @@
 import React from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 
-import { ErrorPage, Table } from '@backstage/core-components';
+import { ErrorPage } from '@backstage/core-components';
```
jerolimov, line 4: Maybe just as a learning for me, why couldn't we use Backstage core-components Table here anymore?

plugins/bulk-import/src/components/Repositories/RepositoriesList.tsx
```diff
@@ -1,48 +1,49 @@
...
   const searchParams = new URLSearchParams(location.search);
+  const [order, setOrder] = React.useState<Order>('asc');
+  const [orderBy, setOrderBy] = React.useState<string>();
   const { openDialog, setOpenDialog, deleteComponent } = useDeleteDialog();
   const { openDrawer, setOpenDrawer, drawerData } = useDrawer();
   const [pageNumber, setPageNumber] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(5);
-  const [searchString, setSearchString] = React.useState('');
-  const classes = useStyles();
+  const [debouncedSearch, setDebouncedSearch] = React.useState('');
```
jerolimov, line 35: Imo: It would be nice to replace this with `useQueryParamState` from `@ackstage/core-components`. It has an optional debounceTime as well.

https://backstage.io/docs/reference/core-components.usequeryparamstate/

plugins/bulk-import/src/components/Repositories/RepositoriesList.tsx
```diff
@@ -55,33 +56,138 @@ export const RepositoriesList = () => {
...
+          />
+        )}
+      </Paper>
+    </Box>
+  );
+};
+
+/*
+
+<Table
```
jerolimov, line 154: Remove old code?

jerolimov (changes requested):
Some small comments and questions:

plugins/bulk-import/src/components/Repositories/CatalogInfoAction.tsx
debsmita1: The edit and open in new tab icons are displayed based on the user's permissions, depending on whether they can edit the pull request

plugins/bulk-import/src/components/Repositories/RepositoriesList.tsx
debsmita1: As of today, we do not add the search text in the query param of the bulk-import URL
I can create a separate ticket to address this

plugins/bulk-import/src/components/BulkImportPage.tsx
```diff
@@ -33,21 +36,27 @@ export const BulkImportPage = () => {
...
-        <Formik
-          initialValues={initialValues}
-          enableReinitialize
-          onSubmit={async (_values: AddRepositoriesFormValues) => {}}
-        >
-          <FormControl fullWidth>
-            <RepositoriesList />
-          </FormControl>
-        </Formik>
+        <QueryClientProvider client={queryClientRef.current as QueryClient}>
```
jerolimov, line 49: `as` always trigger me ;)

```suggestion
        <QueryClientProvider client={queryClientRef.current!}>
```

plugins/bulk-import/src/components/Repositories/CatalogInfoAction.tsx
jerolimov: Anyway, the code generates a link that isn't clickable in some cases.

With the recommended check we can ensure that we don't create a button that doesn't open a valid link.

jerolimov (changes requested):
Tiny nits

jerolimov (approved):
/lgtm

Issue #2112, opened 2024-08-30
Keycloak backend plugin is not working as expected


paralika:
### Describe the bug

I have setup the Keycloak backend plugin by following the steps given on below this link - https://janus-idp.io/plugins/keycloak/ . Plugin configure successfully and it started syncing users and groups, but  then abruptly, it stopped syncing with below error.

Processor BuiltinKindsEntityProcessor threw an error while validating the entity group:default/mainstream-engineering-admin; caused by TypeError: /spec/members/2 must be string - type: string entity=group:default/mainstream-engineering-admin [36mlocation[39m=url:https://authusqa.devops.sws.company.com/auth/admin/realms/chim/groups/480049ad-7a33-41ab-87a6-382a797831e0

I am using the new backend and followed the plugin configuration steps accordingly. 

### Expected Behavior
All users and groups from keyclock realm should be synced

### What are the steps to reproduce this bug?

1. Setup backstage
2. Configure the Keycloak plugin for new backup configuration (using steps from https://janus-idp.io/plugins/keycloak/)

### Versions of software used and environment

@janus-idp/backstage-plugin-keycloak-backend": "^1.13.1",

Backstage - 1.30.4 (create-app@0.5.18)

paralika:
Any comments here please?

kadel:
@paralika This looks like an invalid group definition. Is the `mainstream-engineering-admin` group something that you created, or is it imported by the plugin?

Can you share what this group looks like? Find this group in Catalog, lick on 3 dots in top right corner, and select inspect entity, there you can find Raw Yaml.

paralika:
@kadel "mainstream-engineering-admin" is available in Keycloak. The plugin is trying to import it, but it is not successful yet. I have ~1k groups in Keycloak. Out of that, only 341 groups were imported successfully and for others it is failing with this error.

kadel:
The error message complains about a third member of that group. Can you check the members of that group? Maybe you notice something suspicious. For example, isn't there a user with numeric username?

paralika:
@kadel I do not see any suspicious chars. Please see the screenshot below. The username is an alphanumeric value pointing to the users. 

![image](https://github.com/user-attachments/assets/1542ba6a-2fe4-4950-bb77-72133a28cf4b)

paralika:
No, I do not have any custom transformer set. I am following the instructions given at - https://janus-idp.io/plugins/keycloak/ 

[...]

PR #2319, opened 2024-10-08
fix(keycloak): avoid undefined values for keycloak group members

AndrienkoAleksandr:

Avoid undefined values for Keycloak group members to prevent type errors during group ingestion. Also, fix missing realms, as they could theoretically cause type errors, though very rarely. Log errors with a warning level when retrieving the Keycloak user list.

Fixes: https://issues.redhat.com/browse/RHIDP-3785
Upstream issue: https://github.com/janus-idp/backstage-plugins/issues/2112

### How to test it

Unfortunately, I reproduced the bug synthetically. I don't have access to a very large database to confirm the fix with 100% certainty.


PatAKnight (approved):
/lgtm

PR #2332, opened 2024-10-09
fix(deps): use * instead of version pinning

schultzp2020:


plugins/keycloak-backend/dist-dynamic/package.json
```diff
@@ -70,6 +70,6 @@
   },
   "resolutions": {
     "@aws-sdk/util-utf8-browser": "npm:@smithy/util-utf8@~2",
-    "@keycloak/keycloak-admin-client": "file:./embedded/keycloak-keycloak-admin-client"
+    "@keycloak/keycloak-admin-client": "file:./embedded\\keycloak-keycloak-admin-client"
```
gashcrumb: I don't think you want this change

gashcrumb (approved):

PR #2337, opened 2024-10-09
chore(deps): fix CVE-2024-37890 and update kubernetes-actions and ocm-backend

kim-tsao:
Fixes: https://issues.redhat.com/browse/RHIDP-2733

Changes:
* Needed to bump up kubernetes-actions and ocm-backend to their latest patch levels
~* Use `resolutions` to force resolution of ws to v8.17.1 (otherwise, we'd be picking up 8.18 on a maintenance branch which is greater than the one we have on 1.3)~
* This PR will not include version bumps and CHANGELOG.MD updates.  Those will be handled after another CVE fix in a separate PR.


Fortune-Ndlovu (approved):
/lgtm
/approve

PR #2339, opened 2024-10-10
feat(bulk-import): use react query to fetch repositories

debsmita1:
**Resolves:**
https://issues.redhat.com/browse/RHIDP-4035

### DIFF
--- a/plugins/bulk-import/src/components/AddRepositories/AddRepositoriesForm.test.tsx
+++ b/plugins/bulk-import/src/components/AddRepositories/AddRepositoriesForm.test.tsx
@@ -4,6 +4,7 @@ import { BrowserRouter as Router } from 'react-router-dom';
 import { configApiRef, identityApiRef } from '@backstage/core-plugin-api';
 import { MockConfigApi, TestApiProvider } from '@backstage/test-utils';
 
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { render, screen } from '@testing-library/react';
 import { useFormikContext } from 'formik';
 
@@ -11,8 +12,12 @@ import { useDrawer } from '@janus-idp/shared-react';
 
 import { bulkImportApiRef } from '../../api/BulkImportBackendClient';
 import { mockGetImportJobs, mockGetRepositories } from '../../mocks/mockData';
-import { ImportJobStatus, RepositorySelection } from '../../types';
-import { AddRepositoriesForm } from './AddRepositoriesForm';
+import {
+  ImportJobStatus,
+  OrgAndRepoResponse,
+  RepositorySelection,
+} from '../../types';
+import { AddRepositories } from './AddRepositories';
 
 jest.mock('formik', () => ({
   ...jest.requireActual('formik'),
@@ -44,6 +49,16 @@ jest.mock('@material-ui/core', () => ({
   },
 }));
 
+const createTestQueryClient = () =>
+  new QueryClient({
+    defaultOptions: {
+      queries: {
+        retry: false, // Disable retries for testing
+      },
+    },
+  });
+let queryClient: QueryClient;
+
 class MockBulkImportApi {
   async getImportAction(
     repo: string,
@@ -52,6 +67,13 @@ class MockBulkImportApi {
     return mockGetImportJobs.imports.find(
       i => i.repository.url === repo,
     ) as ImportJobStatus;
+  }
+  async dataFetcher(
+    _pageNo: number,
+    _size: number,
+    _searchString: string,
+  ): Promise<OrgAndRepoResponse> {
+    return mockGetRepositories;
   }
 }
 
@@ -70,9 +92,10 @@ beforeEach(() => {
     },
     setFieldValue: jest.fn(),
   });
+  queryClient = createTestQueryClient();
 });
 
-describe('AddRepsositoriesForm', () => {
+describe('AddRepositoriesForm', () => {
   it('should render the repositories list with the footer', async () => {
     (useDrawer as jest.Mock).mockImplementation(initial => ({
       initial,
@@ -97,7 +120,9 @@ describe('AddRepsositoriesForm', () => {
             ],
           ]}
         >
-          <AddRepositoriesForm error={null} />
+          <QueryClientProvider client={queryClient}>
+            <AddRepositories error={null} />
+          </QueryClientProvider>
         </TestApiProvider>
       </Router>,
     );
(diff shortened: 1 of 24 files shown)


its-mitesh-kumar:
@debsmita1 Could you please resolve quality gate failing and merge the latest change from main .

plugins/bulk-import/src/components/AddRepositories/AddRepositoriesForm.test.tsx
```diff
@@ -97,7 +109,9 @@ describe('AddRepsositoriesForm', () => {
             ],
           ]}
         >
-          <AddRepositoriesForm error={null} />
+          <QueryClientProvider client={queryClient}>
+            <AddRepositories error={null} />
```
christoph-jerolimov, line 113: I see that it's already before like this, but can you make the `error` prop optional and remove the prop here?

plugins/bulk-import/src/components/AddRepositories/AddRepositoriesForm.tsx
```diff
@@ -1,133 +1,112 @@
...
-    paddingLeft: '5px',
-  },
-}));
+export const AddRepositoriesForm = () => {
+  const bulkImportApi = useApi(bulkImportApiRef);
+  const navigate = useNavigate();
+  const [generalSubmitError, setGeneralSubmitError] = React.useState<{
+    message: string;
+    title: string;
+  } | null>(null);
```
christoph-jerolimov, line 32: You can use the error from the useMutation instead.

```suggestion
```

plugins/bulk-import/src/components/AddRepositories/AddRepositoriesForm.tsx
```diff
@@ -1,133 +1,112 @@
...
+      setGeneralSubmitError({
+        message:
+          error?.message ||
+          `${get(error, 'error.message')}\n${get(error, 'error.response.url')}` ||
+          'Failed to create pull request',
+        title: error?.name || get(error, 'error.name') || 'Error occured',
+      });
+      return error;
+    },
+  });
```
christoph-jerolimov, line 63: I think it's easier if you handle this later by using the mutationCreate.error

```suggestion
  const mutationCreate = useMutation(createImportJobs);
```

plugins/bulk-import/src/components/AddRepositories/AddRepositoriesForm.tsx
```diff
@@ -1,133 +1,112 @@
...
-        />
-      )}
-    </>
+    <DrawerContextProvider>
+      <Formik
+        initialValues={initialValues}
+        enableReinitialize
+        onSubmit={handleSubmit}
+      >
+        <AddRepositories error={generalSubmitError} />
```
christoph-jerolimov, line 108: Maybe you can pass just `mutationCreate.error` here. Or you can convert after `const mutationCreate` with

```tsx
  const generalSubmitError = mutationCreaten.errro ? ({
    message:
      error?.message ||
        `${get(error, 'error.message')}\n${get(error, 'error.response.url')}` ||
        'Failed to create pull request',
    title: error?.name || get(error, 'error.name') || 'Error occured',
  }) : null;
```

```suggestion
        <AddRepositories error={generalSubmitError} />
```

Wdyt?

plugins/bulk-import/src/hooks/useRepositories.test.ts
```diff
@@ -10,12 +9,18 @@ jest.mock('@backstage/core-plugin-api', () => ({
...
+}));
 
 describe('useRepositories', () => {
   it('should return repositories', async () => {
-    mockUseApi.mockReturnValue({
-      dataFetcher: jest.fn().mockReturnValue(mockGetRepositories),
+    (useQuery as jest.Mock).mockReturnValue({
+      data: mockGetRepositories,
+      isLoading: false,
+      error: '',
```
christoph-jerolimov, line 22: I would expect error is null when there is no error.

```suggestion
      error: null,
```

plugins/bulk-import/src/hooks/useRepositories.test.ts
```diff
@@ -32,8 +37,11 @@ describe('useRepositories', () => {
   });
 
   it('should return organizations', async () => {
-    mockUseApi.mockReturnValue({
-      dataFetcher: jest.fn().mockReturnValue(mockGetOrganizations),
+    (useQuery as jest.Mock).mockReturnValue({
+      data: mockGetOrganizations,
+      isLoading: false,
+      error: '',
```
christoph-jerolimov, line 43:
```suggestion
      error: null,
```

plugins/bulk-import/src/hooks/useRepositories.test.ts
```diff
@@ -47,13 +55,16 @@ describe('useRepositories', () => {
...
+    (useQuery as jest.Mock).mockReturnValue({
+      data: {
         ...mockGetRepositories,
         repositories: mockGetRepositories.repositories?.filter(
           r => r.organization === 'org/dessert',
         ),
-      }),
+      },
+      isLoading: false,
+      error: '',
```
christoph-jerolimov, line 66:
```suggestion
      error: null,
```

plugins/bulk-import/src/hooks/useRepositories.ts
```diff
@@ -182,8 +191,8 @@ export const useRepositories = (options: {
   const debouncedUpdateResources = useDebounceCallback(prepareData, 200);
 
   React.useEffect(() => {
-    debouncedUpdateResources?.(value, loading);
-  }, [debouncedUpdateResources, value, loading]);
+    debouncedUpdateResources?.(value, isQueryLoading);
+  }, [debouncedUpdateResources, value, isQueryLoading]);
 
   return useDeepCompareMemoize({
```
christoph-jerolimov, line 197: I see an issue that the pagination forward and backend shows always a loading indicator.

I guess that the `useRepository` doesn't use the caching benefits of react-query yet/correctly.

My recommendation is that you remove almost all useStates from this hook.

If you need calculate results, based on the useQuery hook, you can use `useMemo` instead.

The code should also get rid off the effect to call debouncedUpdateResources and the deepCompareMemoize function.

The order should be:

1. useQuery to fetch data
2. useMemo if you must transform "complex" data with a dependency just on the data.

christoph-jerolimov (changes requested):
Hi, from the user perspective I see two issues:

1. When I paginate between the pages, I always see a loading indicator.
2. The order by feature works just on the current page. I expect that this issue exists before, but server-side pagination works only when the server also supports server-side ordering.

debsmita1:
Created PR https://github.com/redhat-developer/rhdh-plugins/pull/23 , so closing this one

PR #2341, opened 2024-10-10
fix(orchestrator): remove docs feature

batzionb:
remove docs feature that breaks dynamic plugin build, and revert PR that attempted to fix that issue but broke multi-semantic-release


gciavarrini (approved):
lgtm

PR #2342, opened 2024-10-10
fix(orchestrator): add scrollbar to variables card

batzionb:
variables card scrollbar was removed following results feature


gciavarrini (approved):
lgtm

PR #2347, opened 2024-10-11
fix(release): rerun export-dynamic

schultzp2020:
Rerun `export-dynamic` to fix the `release` github action.


aljesusg (approved):
LGFM from kiali plugin side

PR #2349, opened 2024-10-11
fix(rbac): move admin creation to its own file

PatAKnight:
## Description

Moves the admin creation to its own file to simplify the code base a little bit. This will also help in the future if we ever decide to allow for the reloading of admins from within the app-config.

Also included is a number of other updates.

- Moves the test fixtures directory up a level
- Includes a new mock-utils.ts file that holds all of the different mocks that we use in all of our test files
    - This is so we can further refactor our tests in the future to remove the redundant mocks throughout
- Renames those redundant mocks to make the future update easier
- Includes a new test-utils.ts file that holds some of the common test util functions that are used in our test files
    - This is so we can further refactor our tests in the future to remove the redundant function declarations throughout

The future refactor should help with some of the duplication issues that we are seeing within SonarCloud


PatAKnight:
The sonarcloud analysis is failing because of the amount of duplication that is in this PR. My plan is to clean this up in a separate PR.

AndrienkoAleksandr (commented):
I really like this pull request and would like to merge it as soon as possible. However, I wonder if the admin-creation.ts file might be better placed in a separate package, perhaps something like 'admin-role' or similar, as it doesn’t seem directly related to file permissions. What are your thoughts on this?

AndrienkoAleksandr (approved):

PR #2353, opened 2024-10-14
fix(keycloak): include 'embedded' folder to keycloak-backend release files

AndrienkoAleksandr:

I tried to update backstage-plugins with newer keycloak-backend plugin, but failed with error:

 https://github.com/janus-idp/backstage-showcase/pull/1770#issuecomment-2412531097


gashcrumb (approved):
This is due to having to remove the `--no-embed-as-dependencies` to use the ESM keycloak client.  That embedded package is (apparently) referenced in the published paackage.json, so this should sort that.

PR #2356, opened 2024-10-15
fix(deps): update kubernetes-client version to fix a CVE

karthikjeeyar:
Fixes:
https://issues.redhat.com/browse/RHIDP-4441

This upgrades the following packages

- ocm-backend
- shared-react
- kubernetes-action
- topology
- tekton

Note: 1.3.x fix for Argocd will be contributed from backstage/community-plugin.

Fix CVE

[CVE-2024-21534
](https://bugzilla.redhat.com/show_bug.cgi?id=2317968)

### DIFF
--- a/plugins/ocm-backend/package.json
+++ b/plugins/ocm-backend/package.json
@@ -58,7 +58,7 @@
     "@backstage/plugin-permission-common": "^0.8.1",
     "@backstage/plugin-permission-node": "^0.8.3",
     "@janus-idp/backstage-plugin-ocm-common": "*",
-    "@kubernetes/client-node": "^0.20.0",
+    "@kubernetes/client-node": "^0.22.1",
     "express": "^4.18.2",
     "semver": "^7.5.4"
   },
--- a/plugins/ocm-backend/package.json
+++ b/plugins/ocm-backend/package.json
@@ -54,7 +54,6 @@
     "@backstage/catalog-model": "^1.7.0",
     "@backstage/errors": "^1.2.4",
     "@backstage/plugin-catalog-node": "^1.13.0",
-    "@backstage/plugin-kubernetes-common": "^0.8.3",
     "@backstage/plugin-permission-common": "^0.8.1",
     "@backstage/plugin-permission-node": "^0.8.3",
     "@janus-idp/backstage-plugin-ocm-common": "*",
--- a/plugins/ocm-backend/package.json
+++ b/plugins/ocm-backend/package.json
@@ -31,8 +31,8 @@
   "scripts": {
     "build": "backstage-cli package build",
     "clean": "backstage-cli package clean",
-    "export-dynamic": "janus-cli package export-dynamic-plugin --embed-package @backstage/plugin-kubernetes-common",
-    "export-dynamic:clean": "janus-cli package export-dynamic-plugin --embed-package @backstage/plugin-kubernetes-common --clean",
+    "export-dynamic": "janus-cli package export-dynamic-plugin",
+    "export-dynamic:clean": "janus-cli package export-dynamic-plugin --clean",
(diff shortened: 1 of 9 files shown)


karthikjeeyar:
we are using `plugin-kubernetes-common` dependancy in 4 out of 5 plugins changed in this PR, and that plugin is pinned to v0.20.0 https://github.com/backstage/backstage/blob/master/plugins/kubernetes-common/package.json#L58 causing issues in CI checks.

nickboldt (approved):

plugins/ocm-backend/package.json
```diff
@@ -31,7 +31,7 @@
   "scripts": {
     "build": "backstage-cli package build",
     "clean": "backstage-cli package clean",
-    "export-dynamic": "janus-cli package export-dynamic-plugin --embed-package --embed-package @backstage/plugin-kubernetes-common --no-embed-as-dependencies",
+    "export-dynamic": "janus-cli package export-dynamic-plugin --no-embed-as-dependencies",
```
karthikjeeyar, line 34: The CI tests were failing because the `ocm-backend` was embedding `@backstage/plugin-kubernetes-common` in export-dynamic script whereas other backend plugin does not embed it. The embedded package is internally pinned to a older version of `kubernetes/client-node` causing the conflict with the PR changes.

https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm-backend/package.json#L34

kim-tsao:
@PatAKnight can you or your team please review the ocm-backend plugin?  Can you consider the QE and docs impact this change may have?

plugins/ocm-backend/package.json
kim-tsao: @PatAKnight are you ok with this change?

gashcrumb: It's probably being embedded because the ocm plugin has a runtime dependency on this package, and it won't be present in the package search path when run as a dynamic plugin.  I recall there was a general issue to remove the dependency on the kubernetes plugin code from a few plugins, so if this has been changed as well then this is probably going to be fine.  Otherwise, if this code is still using the kubernetes plugin API client, this change will cause this plugin to break when run in RHDH.

gashcrumb: Took a slightly closer look at the OCM code in particular, I don't see it actually using this dependency at a quick glance and now I'm wondering why it's even listed in the [package.json](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/ocm-backend/package.json#L56) for this plugin.  It's worth trying to clean this up if possible.

karthikjeeyar: @gashcrumb Thanks for the review. I see there is a const being imported from this pacakge in [ManagedClusterProvider.ts](https://github.com/janus-idp/backstage-plugins/blob/218c4f0776cad63f62ade6489a5d612e4ab444a8/plugins/ocm-backend/src/providers/ManagedClusterProvider.ts#L33). I have now defined that const locally in `constants.ts` and removed the dependancy of `@backstage/plugin-kubernetes-common` from package.json.

PatAKnight:
Plugin still works from my side of things. 

But I was unable to test the ocm backend plugin dynamically. It might have to do with the plugins repo being on backstage 1.30 vs showcase repo being on backstage 1.29 and the various deprecations in `@backstage/backend-dynamic-feature-service`. @gashcrumb is that something that you can verify?

PatAKnight:
I ended up figuring out how to get the ocm backend dynamic plugin tested. It ended up being the version mismatch after some hacking of backstage-showcase. Looks good to me. @invincibleJai, does anyone from the frontend team want to take a look at this?

kim-tsao (approved):

PR #2228, opened 2024-09-23
fix(rbac): fix ownerRefs alias after upgrade backstage to 1.29

AndrienkoAleksandr:
### What does this pr do:

Fix ownerRefs alias after upgrade backstage to 1.29

### What does this pr fixes

Fixes: 

https://issues.redhat.com/browse/RHIDP-4069

PR #2143, opened 2024-09-09
feat(rbac): allow to disable group inheritance

AndrienkoAleksandr:
### What does this pull request do:

Allow to disable group inheritance feature.

PR #2375, opened 2024-10-17
chore(release): update RBAC to delivery bug fixes for RHDH 1.3.1

AndrienkoAleksandr:
### What does this pull request fixes:

Fixes https://issues.redhat.com/browse/RHIDP-4069

### DIFF
--- a/.changeset/fuzzy-dryers-lie.md
+++ b/.changeset/fuzzy-dryers-lie.md
@@ -1,0 +1,8 @@
+---
+"@janus-idp/backstage-plugin-rbac-backend": patch
+---
+
+Update RBAC dependencies to address the following issues:
+
+- fix broken conditional alias $ownerRefs
+- add the ability to disable RBAC group inheritance
--- a/plugins/rbac-backend/docs/group-hierarchy.md
+++ b/plugins/rbac-backend/docs/group-hierarchy.md
@@ -135,8 +135,10 @@ permission:
     maxDepth: 1
 ```
 
-The `maxDepth` must be greater than 0 to ensure that the graphs are built correctly. Also the graph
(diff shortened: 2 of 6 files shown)


.changeset/fuzzy-dryers-lie.md
```diff
@@ -0,0 +1,12 @@
+---
+"@janus-idp/backstage-plugin-audit-log-node": major
+"@janus-idp/backstage-plugin-rbac-backend": major
+"@janus-idp/backstage-plugin-rbac-common": major
+"@janus-idp/backstage-plugin-rbac-node": major
+"@janus-idp/backstage-plugin-rbac": major
```
PatAKnight, line 6: These should be consider patch instead of major

plugins/rbac-backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@janus-idp/backstage-plugin-rbac-backend",
-  "version": "4.10.6",
+  "version": "4.12.2",
```
PatAKnight, line 3: From my understanding, we don't need this change. Changesets should make this change for us after we merge this pr

PatAKnight (commented):
I believe that there are a number of unnecessary changes in this PR. We just need to backport the [fix ownerRefs](https://github.com/janus-idp/backstage-plugins/pull/2228/files) and [allow to disable group inheritance](https://github.com/janus-idp/backstage-plugins/pull/2143/files) PRs. Is there any reason that we need to include the changes to audit log, rbac common, rbac node, and rbac plugins?

plugins/rbac-backend/package.json
AndrienkoAleksandr: Ok, I removed this change. So next version for RHDH 1.3.1 will be 4.10.7?

PR #2401, opened 2024-10-18
chore(deps): change local package references to a '*'

schultzp2020:
continuation of https://issues.redhat.com/browse/RHIDP-4430


nickboldt (approved):
pretty sure we already did something very similar last week, so +1 for me

plugins/aap-backend/package.json
```diff
@@ -53,7 +53,7 @@
     "@backstage/backend-test-utils": "1.0.0",
     "@backstage/cli": "0.27.1",
     "@backstage/plugin-catalog-backend": "1.26.1",
-    "@janus-idp/cli": "1.15.2",
+    "@janus-idp/cli": "*",
```
gashcrumb, line 56: Would this work alright on the `release-1.3` branch?  I think probably, as it'd be using that branch of the source.  However this would mean we should be careful to always ensure CLI changes cause CLI releases, otherwise it's possible this could do a release using CLI code that is on the main branch but unreleased.

PR #2405, opened 2024-10-18
chore(yarn): upgrade to yarn v3

schultzp2020:


kim-tsao (approved):

.github/workflows/release.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          yarn install --frozen-lockfile
+          yarn install
```
nickboldt, line 47: we don't want to release with a frozen lock file? we want the random internet resolution to take effect on every release ?

nickboldt (approved):

.github/workflows/release.yaml
schultzp2020: `--frozen-lockfile` has been replace with `--immutable` which is automatically enabled when it detects a CI environment,

PR #2407, opened 2024-10-21
fix(rbac): prevent policy creation when rbac plugin is disabled

PatAKnight:
## Description
Adds a new allow all permission policy that can be used to whenever the permission framework is disabled. This allows us to skip the creation of a number of resources because we load in this allow all permission policy instead of the RBAC one. 

Also included is a number of other updates.

- Creates a new directory called `policies` and moves the `allow-all-policy.ts` and `permission-policy.ts` to this directory
- Cleans up the options for the `PoliciesServer` as there was some duplicates

### DIFF
--- a/.changeset/beige-parrots-act.md
+++ b/.changeset/beige-parrots-act.md
@@ -1,0 +1,5 @@
+---
+"@janus-idp/backstage-plugin-rbac-backend": patch
+---
+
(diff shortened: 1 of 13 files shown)


AndrienkoAleksandr (commented):
Tested, works like expected. Looks good to me.

AndrienkoAleksandr (approved):

PR #2408, opened 2024-10-21
fix(cli): allow mutable install for monorepo yarn.lock

gashcrumb:
This change updates the CLI to use `--no-immutable` when a plugin package or monorepo `yarn.lock` file is used during the `export-dynamic-plugin` command. Explicitly passing this flag avoids the default Yarn 3.x behavior of `--immutable` when the command is run in a CI environment.

@schultzp2020 @davidfestal fyi


schultzp2020 (approved):
/lgtm
/approve

PR #2415, opened 2024-10-22
chore(gh): update release action to trigger workflows

schultzp2020:
Update the release action to trigger the workflows by using the elevated gh token.


.github/workflows/update-changeset-prs.yaml
```diff
@@ -31,17 +54,7 @@ jobs:
...
-
       - name: Amend previous commit
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
-          commit_author: ${{ steps.last-commit.outputs.author }}
-          commit_message: ${{ steps.last-commit.outputs.message }}
-          commit_options: '--amend --no-edit'
-          push_options: '--force'
-          skip_fetch: true
+          commit_message: 'chore: update yarn.lock'
```
nickboldt, line 60: should we also set `file_pattern` to restrict this to only the yarn.lock file(s) ?

nickboldt: do you need the whole "is the previous commit message a `chore: update yarn.lock` ? because if the yarn lock regen shows no changes, there's nothing to commit.

.github/workflows/update-changeset-prs.yaml
```diff
@@ -6,20 +6,43 @@ on:
...
+        with:
+          # Fetch the latest two commits
+          fetch-depth: 2
+
+      # We do not want an infinite loop; therefore, we are checking if the previous
+      # commit was created by this action.
+      - name: Get previous commit message
+        id: prev-commit
+        run: |
+          echo "is-valid=$(if [[ $(git log -1 --pretty=%B HEAD^) != 'chore: update yarn.lock' ]]; then echo 'true'; else echo 'false'; fi)" >> $GITHUB_OUTPUT
```
nickboldt, line 26: bash logic looks OK, though I don't why you're checking the previous commit. should this not be the latest one?

nickboldt:
```suggestion
          # check the latest commit message in this branch for a yarn lock update message
          echo "is-valid=$(if [[ $(git log -1 --pretty=%B HEAD) != 'chore: update yarn.lock' ]]; then echo 'true'; else echo 'false'; fi)" >> $GITHUB_OUTPUT
```

.github/workflows/update-changeset-prs.yaml
```diff
@@ -6,20 +6,43 @@ on:
...
+    name: Verify previous commit
+    runs-on: ubuntu-latest
+    outputs:
+      is-valid: ${{ steps.prev-commit.outputs.is-valid }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+        with:
+          # Fetch the latest two commits
+          fetch-depth: 2
```
nickboldt, line 19:
```suggestion
          fetch-depth: 1
```

.github/workflows/update-changeset-prs.yaml
```diff
@@ -6,20 +6,40 @@ on:
...
 jobs:
+  prev-commit:
+    name: Verify previous commit
+    runs-on: ubuntu-latest
+    outputs:
+      valid: ${{ steps.prev-commit.outputs.valid }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
```
nickboldt, line 17:
```suggestion
        with:
          fetch-depth: 1
```

schultzp2020: The default is 1, so it's not needed.

nickboldt (approved):
sure, let's do this :success-kid:

PR #2416, opened 2024-10-22
chore(gh): update/remove gh actions

schultzp2020:
Changes:

- Removed the dirty `yarn.lock` check since Yarn v3 now uses the `--immutable` flag during `yarn install` in CI environments.
- Deleted `yarn-lock.yaml` as it's no longer necessary (same reason as above).
- Renamed `pr-playwright` to `pr-ui-test` and updated the action to use Turbo instead.
- Deleted `publish-backend-plugin-manager.yaml` as it is no longer required.
- Merge PR tests into a single file.


schultzp2020:
/override "Build with Node.js 20"

Merged into a single file.

schultzp2020:
/override "Test with Node.js 20"

Merged into a single file.

schultzp2020:
/override "SonarCloud Code Analysis"

Code analysis should not be looking at this PR.

PR #2429, opened 2024-10-22
chore(release): backport keycloak fixes but without upgrade keycloak client

AndrienkoAleksandr:
# What does this pull request fixes:

https://issues.redhat.com/browse/RHIDP-3785
https://issues.redhat.com/browse/RHIDP-4391

### DIFF
--- a/.changeset/early-trainers-dance.md
+++ b/.changeset/early-trainers-dance.md
@@ -1,0 +1,7 @@
+---
+"@janus-idp/backstage-plugin-keycloak-backend": major
+---
+
+Provide keycloak-backend fixes:
+- avoid undefined values for keycloak group members
+- retrieve full list group members using pagination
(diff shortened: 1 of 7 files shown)


.changeset/early-trainers-dance.md
```diff
@@ -0,0 +1,7 @@
+---
+"@janus-idp/backstage-plugin-keycloak-backend": major
```
PatAKnight, line 2: This should be a patch instead of major

plugins/keycloak-backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@janus-idp/backstage-plugin-keycloak-backend",
-  "version": "1.13.3",
+  "version": "1.13.4",
```
PatAKnight, line 3: From my understanding, we don't need this change. Changesets should make this change for us after we merge this pr

PatAKnight (commented):
Looks good to me, just a couple of minor changes needed

PR #2440, opened 2024-10-23
chore(release): revert the major update so that we can release under a patch

PatAKnight:
## Description

Looks like the keycloak plugin was released under a `major` update instead of a `patch`. This PR reverts back to the original version and adds a patch changeset so that we can release it properly.


schultzp2020 (approved):
/approve
/lgtm

PR #2451, opened 2024-10-25
feat(lightspeed-backend): send conversation history to langchain streams

karthikjeeyar:
This PR implements the following changes required for frontend.

1. sends the user conversation history from the store to the langchain stream method.
2. add message created timestamp. i.e `created_at` and `role` in  `response_metadata` object. 

image

---
**Unit tests:**

 POST /v1/query
     
      ✓ should not have any history for the initial conversation (25 ms)
      ✓ should call the stream with conversation history (17 ms)
      ✓ should contain ai and human message timestamp (218 ms)

### DIFF
--- a/plugins/lightspeed-backend/src/handlers/chatHistory.ts
+++ b/plugins/lightspeed-backend/src/handlers/chatHistory.ts
@@ -14,19 +14,35 @@ export async function saveHistory(
   conversation_id: string,
   role: string,
   message: string,
+  timestamp?: number,
 ): Promise<void> {
   let newMessage: BaseMessage;
   switch (role) {
     case Roles.AIRole: {
-      newMessage = new AIMessage(message);
+      newMessage = new AIMessage({
+        content: message,
+        response_metadata: {
+          created_at: timestamp || Date.now(),
+        },
+      });
       break;
     }
     case Roles.HumanRole: {
-      newMessage = new HumanMessage(message);
+      newMessage = new HumanMessage({
+        content: message,
+        response_metadata: {
+          created_at: timestamp || Date.now(),
+        },
+      });
       break;
     }
     case Roles.SystemRole: {
-      newMessage = new SystemMessage(message);
+      newMessage = new SystemMessage({
+        content: message,
+        response_metadata: {
+          created_at: timestamp || Date.now(),
+        },
+      });
       break;
     }
     default:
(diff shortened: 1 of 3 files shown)


karthikjeeyar:
cc: @yangcao77

plugins/lightspeed-backend/src/handlers/chatHistory.ts
```diff
@@ -14,19 +14,38 @@ export async function saveHistory(
...
 ): Promise<void> {
   let newMessage: BaseMessage;
   switch (role) {
     case Roles.AIRole: {
-      newMessage = new AIMessage(message);
+      newMessage = new AIMessage({
+        content: message,
+        response_metadata: {
+          created_at: timestamp || Date.now(),
+          role: Roles.AIRole,
```
yangcao77, line 26: it's already been stored as an `AIMessage`. the role is redundant here. 
we can get the `AIMessage` or `HumanMessage` from the id, as shown in your screenshot

plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -130,11 +130,30 @@ export async function createRouter(
...
+            : new AIMessage({ content: message.content }),
+        );
+
         const chain = prompt.pipe(openAIApi);
         let content = '';
+        const userMessageTimestamp = Date.now();
+        let botMessageTimestamp;
         for await (const chunk of await chain.stream({
-          messages: [new HumanMessage(query)],
+          messages: [...userHistory, new HumanMessage(query)],
```
yangcao77, line 151: since this PR feeds back the previous user history. we might want to introduce a couple new test cases to test the memory of previous conversations.

plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -145,8 +164,18 @@ export async function createRouter(
         }
         response.end();
 
-        await saveHistory(conversation_id, Roles.HumanRole, query);
-        await saveHistory(conversation_id, Roles.AIRole, content);
+        await saveHistory(
+          conversation_id,
+          Roles.HumanRole,
+          query,
+          userMessageTimestamp,
```
yangcao77, line 171: can we also simply test the timestamp as well? probably just check the query response contains a timestamp, and it's increased compare with previous messages.

plugins/lightspeed-backend/src/service/router.ts
karthikjeeyar: added test cases.

yangcao77 (approved):
lgtm

PR #2467, opened 2024-10-31
fix(ci): change fetch depth

schultzp2020:


schultzp2020:
git cherry-pick 1.2.x

PR #2473, opened 2024-11-01
fix(rbac): sync data from db on read operations

AndrienkoAleksandr:


AndrienkoAleksandr:
Close on favour: https://github.com/backstage/community-plugins/pull/2040

PR #2474, opened 2024-11-03
fix(orchestrator): fix issues with extendable workflow execution form

batzionb:
Resolved the following issues:

1. enabled validation using customValidate, and replaced extraErrors with getExtraErrors, since extraErrors is supposed to be populated when running onSubmit, and that isn't exposed to the user. Added busy handling while calling getExtraErrors.
2. moved FormComponent to a separate component, to avoid buggy behavior and code smells with component generated in a different component.
3. update formData on each change instead of when moving to next step, to avoid data being cleared.
4. fix bug in validator - it only worked in first step, because of issue in @rjsf form
5. removed unnecessary package json-schema that was used just for lint error, and fixed the root cause of lint error when importing types from @types/json-schema


batzionb:
@mareklibra PTAL

.changeset/brown-buttons-burn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@janus-idp/backstage-plugin-orchestrator-form-react": patch
+"@janus-idp/backstage-plugin-orchestrator-form-api": patch
+---
+
+fix several issues
```
mareklibra, line 6: The changelog will look beter with additional details :-)

plugins/orchestrator-form-api/README.md
```diff
@@ -14,7 +14,7 @@ The decorator will be provided through a factory method that leverages a [Backst
 
 ```typescript
 export type FormDecorator = (
-  FormComponent: React.ComponentType<Partial<FormProps>>,
+  FormComponent: React.ComponentType<FormDecoratorProps>,
```
mareklibra, line 17: nit: does the user know what is `FormDcoratorProps`? Where to import it from?

plugins/orchestrator-form-react/src/components/OrchestratorFormWrapper.tsx
```diff
@@ -25,55 +28,126 @@ type OrchestratorFormWrapperProps = {
...
+            : _extraErrors,
+        );
+      } catch (err) {
+        _validationError = err as Error;
+      } finally {
+        handleValidateEnded();
+      }
+    }
+    setValidationError(_validationError);
+    if (
```
mareklibra, line 94: How difficult is to disable the Next if there is a validation error? So the user can not issue the Submit.

batzionb: The default way RJSF form works, is that next is alwasy enabled, and the errors show after clicking on it.
This is also the way the backstage create from template works.
Otherwise, you show errors before user wants to see them, like while typing.
There are different approaches to this, but I chose to go with the consistent behavior with backstage create template.
This is also the behavior in many websites.

mareklibra: thanks, it makes sense

.changeset/brown-buttons-burn.md
batzionb: fixed

plugins/orchestrator-form-api/README.md
batzionb: added it to interface description
also mentioned the full plugin example location (although it isn't ready yet...)

mareklibra (approved):

PR #2477, opened 2024-11-04
feat(bulk-import): update the preview pull request form to have it's own formik context

debsmita1:
**Resolves:**
https://issues.redhat.com/browse/RHIDP-3890


plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
```diff
@@ -49,15 +49,11 @@ export const PreviewPullRequestForm = ({
   setPullRequest,
   formErrors,
   setFormErrors,
-  entityOwner,
-  setEntityOwner,
 }: {
   repoId: string;
   repoUrl: string;
   pullRequest: PullRequestPreviewData;
```
christoph-jerolimov, line 55: Couldn't you use here just one PullRequest Data (type `PulllRequestPreview`) instead of a Record?

Then it should be possible to remove the repoId type?

```suggestion
  pullRequest: PullRequestPreview;
```

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
```diff
@@ -49,15 +49,11 @@ export const PreviewPullRequestForm = ({
...
   formErrors: PullRequestPreviewData;
-  entityOwner: string;
-  setEntityOwner: (name: string) => void;
   setPullRequest: (pullRequest: PullRequestPreviewData) => void;
```
christoph-jerolimov, line 57: Same question here, couldn't we simplify the PreviewPullRequestForm when just return one PullRequest here?

```suggestion
  setPullRequest: (pullRequest: PullRequestPreview) => void;
```

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
```diff
@@ -68,6 +64,14 @@ export const PreviewPullRequestForm = ({
   const approvalTool =
     values.approvalTool === 'git' ? 'Pull request' : 'ServiceNow ticket';
 
+  const formik = useFormik<PullRequestPreviewData>({
+    enableReinitialize: true,
+    initialValues: pullRequest,
+    initialErrors: formErrors as any as FormikErrors<PullRequestPreviewData>,
```
christoph-jerolimov, line 70: :eyes:

Why can the form has initial errors?

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
```diff
@@ -271,31 +215,46 @@ export const PreviewPullRequestForm = ({
...
+    };
+
+    if (!formik.errors[repoId]) {
+      delete err[repoId];
+    }
+
+    setFormErrors(err);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [formik.errors]);
```
christoph-jerolimov, line 256: This tries to pass the errors from the sub-form into the parent form? Why?

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
```diff
@@ -308,10 +267,11 @@ export const PreviewPullRequestForm = ({
         variant="outlined"
         margin="normal"
         fullWidth
-        name={`repositories.${pullRequest[repoId]?.componentName}.prTitle`}
-        value={pullRequest[repoId]?.prTitle}
+        name={`${repoId}.prTitle`}
+        value={formik.values?.[repoId]?.prTitle}
```
christoph-jerolimov, line 271: Wouldn't it be easier if you can remove the `repoId` and the current formik state contains just information of the current form?

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
```diff
@@ -417,26 +358,16 @@ export const PreviewPullRequestForm = ({
...
         control={
           <Checkbox
-            checked={pullRequest[repoId]?.useCodeOwnersFile}
-            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
-              const pr = {
-                ...pullRequest,
-                [repoId]: {
-                  ...pullRequest[repoId],
-                  useCodeOwnersFile: event.target.checked,
+            name={`${repoId}.useCodeOwnersFile`}
```
christoph-jerolimov, line 361: Remove repoId

christoph-jerolimov (changes requested):
I think we should do one more step and don't use the same state as in the parent form. We should just pass a single PR to the sub-form-state.

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
debsmita1: The form uses a record instead of a single pull request due to the complexity in managing multiple preview forms in a tabbed view for organizations. In this view, it’s necessary to save changes across multiple preview forms and display field errors that might occur in each pull request preview.

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
debsmita1: This ensures that in the tabbed view, if there's an error in one tab, switching to another preview form within the same organization retains the errors and keeps the Save button disabled.

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
debsmita1: The sub-form can be a single preview form or a tabbed view of multiple preview forms and the save button is the common button in the main form that saves all the changes made in the sub form(s)

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
debsmita1: I initially started with this thought, but the complexity arises in the tabbed view

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
christoph-jerolimov: Got it, thanks

plugins/bulk-import/src/components/PreviewFile/PreviewPullRequestForm.tsx
debsmita1: I have updated the PR

debsmita1:
Created PR https://github.com/redhat-developer/rhdh-plugins/pull/22, so closing this one

PR #2484, opened 2024-11-04
feat(cli): discover additional app-config yamls

christoph-jerolimov:
Prefers:

1. examples overs non-examples
2. dynamic over janus-idp and backstage-community
3. fallback to example and non-example app-config

Wdyt? :)

### DIFF
--- a/packages/cli/src/commands/package-dynamic-plugins/command.ts
+++ b/packages/cli/src/commands/package-dynamic-plugins/command.ts
@@ -293,21 +293,16 @@ function discoverPluginConfigurationFile(
 ): string | undefined {
   // Possible file names, the first match will be used
   const supportedFilenames = [
+    'app-config.dynamic.example.yaml',
+    'app-config.dynamic.yaml',
     'app-config.janus-idp.yaml',
     'app-config.backstage-community.yaml',
+    'app-config.example.yaml',
     'app-config.yaml',
   ];
   return supportedFilenames
-    .map<boolean>((fileName: string) => {
-      const candidate = path.join(directory, fileName);
-      return fs.existsSync(candidate);
-    })
-    .reduce<string | undefined>((val, current, index) => {
-      if (typeof val === 'undefined' && current) {
-        return path.join(directory, supportedFilenames[index]);
-      }
-      return val;
-    }, undefined);
+    .map(fileName => path.join(directory, fileName))
+    .find(candidate => fs.existsSync(candidate));
 }
 
 /**


packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -293,21 +293,16 @@ function discoverPluginConfigurationFile(
...
-      return fs.existsSync(candidate);
-    })
-    .reduce<string | undefined>((val, current, index) => {
-      if (typeof val === 'undefined' && current) {
-        return path.join(directory, supportedFilenames[index]);
-      }
-      return val;
-    }, undefined);
+    .map(fileName => path.join(directory, fileName))
+    .find(cadidate => fs.existsSync(cadidate));
```
gashcrumb, line 305: typo here in "candidate"

packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -293,21 +293,16 @@ function discoverPluginConfigurationFile(
 ): string | undefined {
   // Possible file names, the first match will be used
   const supportedFilenames = [
+    'app-config.dynamic.example.yaml',
+    'app-config.dynamic.yaml',
     'app-config.janus-idp.yaml',
     'app-config.backstage-community.yaml',
+    'app-config.example.yaml',
```
gashcrumb, line 300: I'm not sure about looking for `app-config.example.yaml` and even the existing `app-config.yaml` entry that was originally here.  One thing I'd like to avoid is accidentally picking up one of these files that may contain a complete backstage configuration file for this.

christoph-jerolimov: I can remove `app-config.yaml`, but please be aware that some plugins uses this currently.

But since this feature is not documented (or?), I'm do what you prefer.

Should I remove it? :)

gashcrumb: Yep, we haven't documented this convention at all yet, so we can decide what we want to keep in the longer term.  I'm not sure of the origin of even the `app-config.janus-idp.yaml` convention, either the intent was it's a file that is meant to be handled by the `janus-idp` CLI or maybe the intent was just to match it with the organization name, like what was done when plugins were moved to the community repo and this `app-config.backstage-community.yaml` convention was introduced.

That being said, the more I think about it, the more I think maybe it's better to just keep it simple and maybe we should eventually just narrow down this list to `app-config.yaml` despite my worry that an entire yaml config gets picked up by the tooling, it's much easier to remember :-)

gashcrumb:
This probably needs a [changeset](https://github.com/janus-idp/backstage-plugins/blob/main/CONTRIBUTING.md#adding-a-changeset)

packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -293,21 +293,16 @@ function discoverPluginConfigurationFile(
...
-      return fs.existsSync(candidate);
-    })
-    .reduce<string | undefined>((val, current, index) => {
-      if (typeof val === 'undefined' && current) {
-        return path.join(directory, supportedFilenames[index]);
-      }
-      return val;
-    }, undefined);
+    .map(fileName => path.join(directory, fileName))
+    .find(candidate => fs.existsSync(candidate));
```
gashcrumb, line 305: :+1: to this change anyways :smile:

packages/cli/src/commands/package-dynamic-plugins/command.ts
gashcrumb: or better still as @davidfestal suggested `app-config.dynamic.yaml` as the one we document and favor over the other fallbacks, I think that naming makes it pretty clear it's not like the other config files.

gashcrumb: Maybe [this log statement](https://github.com/janus-idp/backstage-plugins/blob/main/packages/cli/src/commands/package-dynamic-plugins/command.ts#L155) could be updated as part of this too to mention that.

PR #2487, opened 2024-11-05
fix(orchestrator): remove try-catch that hide sonataflow errors

ydayagi:
[FLPATH-1726](https://issues.redhat.com//browse/FLPATH-1726)
https://issues.redhat.com/browse/FLPATH-1726


ydayagi:
@gciavarrini @pkliczewski @batzionb

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -193,14 +169,9 @@ export class SonataFlowService {
...
   }): Promise<boolean> {
-    try {
-      const urlToFetch = `${args.serviceUrl}/management/processes/${args.definitionId}`;
-      const response = await fetch(urlToFetch);
-      return response.ok;
-    } catch (error) {
-      this.logger.debug(`Error when pinging workflow service: ${error}`);
-    }
-    return false;
+    const urlToFetch = `${args.serviceUrl}/management/processes/${args.definitionId}`;
```
batzionb, line 172: you need to check if response is ok and if not throw an error

batzionb (changes requested):

plugins/orchestrator-backend/src/service/SonataFlowService.ts
batzionb: or is that a separate task?
@gciavarrini

plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -193,14 +169,9 @@ export class SonataFlowService {
...
+    const response = await fetch(urlToFetch);
+    return response.ok;
   }
 
   public async updateInstanceInputData(args: {
```
batzionb, line 177: this whole function should be deleted, it used to be wrongly used by retrigger
waste of time to fix issue in this code

plugins/orchestrator-backend/src/service/SonataFlowService.ts
ydayagi: this is handled in another story https://issues.redhat.com/browse/FLPATH-1725

plugins/orchestrator-backend/src/service/SonataFlowService.ts
ydayagi: we are trying to keep code consistent. removing this function should be done in a separate story

batzionb:
I was playing around today a lot with the error handling, 
Seems like we can't just have errors being thrown without catching them in the top level
We need to revisist this task

batzionb:
An example for when this PR being merged will cause an issue:
The API fetchWorkflowDefinition is called [here](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator-backend/src/service/router.ts#L530) without being surrounded by try or catch.
If errors aren't handled on the top level in router.ts, the API won't return any result and remain hanging.

ydayagi:
> An example for when this PR being merged will cause an issue: The API fetchWorkflowDefinition is called [here](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator-backend/src/service/router.ts#L530) without being surrounded by try or catch. If errors aren't handled on the top level in router.ts, the API won't return any result and remain hanging.
> 
> I do think we should remove the try catch from OrchestratorService, but all the API definitions in router.ts should have try catch.

something here does not make sense. it seems there's an async operation we are not waiting for. I will chck

batzionb:
> > An example for when this PR being merged will cause an issue: The API fetchWorkflowDefinition is called [here](https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator-backend/src/service/router.ts#L530) without being surrounded by try or catch. If errors aren't handled on the top level in router.ts, the API won't return any result and remain hanging.
> > I do think we should remove the try catch from OrchestratorService, but all the API definitions in router.ts should have try catch.
> 
> something here does not make sense. it seems there's an async operation we are not waiting for. I will chck

If an API is called and an error was throws in the backend and not caught, the API won't return anything, it'll remain pending.

batzionb (approved):

batzionb:
@gciavarrini I'm OK with the code changes
Can you PTAL at the test changes?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -75,47 +75,40 @@ describe('SonataFlowService', () => {
...
+      try {
+        await sonataFlowService.fetchWorkflowInfoOnService({
+          definitionId,
+          serviceUrl,
+        });
+      } catch (error) {
+        result = error;
+      }
+
+      expect(result).toBeDefined();
```
gciavarrini, line 88: why do not use `rejects.toThrow(<expected error message>);`?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -75,47 +75,40 @@ describe('SonataFlowService', () => {
...
+      let result;
+      try {
+        await sonataFlowService.fetchWorkflowInfoOnService({
+          definitionId,
+          serviceUrl,
+        });
+      } catch (error) {
+        result = error;
+      }
+
```
gciavarrini, line 105: same question about `.rejects`

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -155,32 +148,23 @@ describe('SonataFlowService', () => {
     beforeEach(() => {
       jest.clearAllMocks();
     });
-    it('should return workflow execution response when the request is successful', async () => {
+    it('should propagate thrown error when the request is successful', async () => {
```
gciavarrini, line 151: if the request is successful, why we have an error to propagate?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -155,32 +148,23 @@ describe('SonataFlowService', () => {
...
-        serviceUrl,
-        inputData: { var1: 'value1' },
-      });
-
-      // Then
-      expect(fetch).toHaveBeenCalledWith(
-        urlToFetch,
-        expectedFetchRequestInit(),
-      );
-      expect(result).toEqual({ id: definitionId, status: 'completed' });
```
gciavarrini, line 174: why removing this?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -203,68 +187,57 @@ describe('SonataFlowService', () => {
...
         json: { details: undefined, stack: undefined },
       });
 
-      const result = await runErrorTest();
+      let result;
+      try {
+        await runErrorTest();
+      } catch (error) {
+        result = error;
+      }
```
gciavarrini, line 204: same question about `.rejects`

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -203,68 +187,57 @@ describe('SonataFlowService', () => {
...
 
-      // Then
-      expect(fetch).toHaveBeenCalledWith(
```
gciavarrini, line 238: why removing this?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -203,68 +187,57 @@ describe('SonataFlowService', () => {
...
+      let result;
+      try {
+        await sonataFlowService.executeWorkflow({
+          definitionId,
+          serviceUrl,
+          inputData: inputData,
+        });
+      } catch (error) {
+        result = error;
+      }
```
gciavarrini, line 238: same comment about `.rejects`

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -494,115 +494,120 @@ function setupInternalRoutes(
...
-        : undefined;
-
-      const workflowData = instanceVariables
-        ? services.dataInputSchemaService.extractWorkflowData(instanceVariables)
-        : undefined;
-
-      const workflowInfo = await routerApi.v2
-        .getWorkflowInputSchemaById(workflowId, serviceUrl)
-        .catch((error: { message: string }) => {
-          auditLogRequestError(error, endpointName, endpoint, req);
```
gciavarrini, line 572: cayou explain how this change is related to error handling? I'm getting lost

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: it works so what's wrong with the current code?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: it works so what's wrong with the current code?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: ithink it is by mistake. checking

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: it is an error flow. we don't need to check for completed it is not the target of the test. expecting the fetch is absurd. we tell the mock to expect it and it will validate it . so why double test?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: it works so what's wrong with the current code?

ydayagi: it works so what's wrong with the current code?

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: expecting the fetch is absurd. we tell the mock to expect it and it will validate it . so why double test?

plugins/orchestrator-backend/src/service/router.ts
ydayagi: all V2 functions catch the errors, call audit and return result. only this API did not do it. some of the code is outside the call to the V2 object so we had to wrap it all with try-catch.

plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
ydayagi: fixed

plugins/orchestrator-backend/src/service/router.ts
batzionb: we can't remove try catch from sonataflow service unless all the calls to it in the router catch the errors. this was not the case with /inputSchema API so Yaron fixed it

batzionb (approved):

PR #2491, opened 2024-11-05
fix: use background button color from theme

mareklibra:
Fixes: [FLPATH-1825](https://issues.redhat.com//browse/FLPATH-1825)


batzionb:
IIUC this fix makes it look the same for disabled and enabled
In the first step the back button is disabled
It doesn't look that good but it's consistent with "create from template"

batzionb:
I think the bug is in RHDH them of .MuiDisabled
Also it's not consistent, sometimes it works fine

batzionb (approved):

PR #2495, opened 2024-11-06
fix: make long workflow result message fully visible

mareklibra:
Fixes: [FLPATH-1833](https://issues.redhat.com//browse/FLPATH-1833)

### DIFF
--- a/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
+++ b/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
@@ -50,6 +50,8 @@ const useStyles = makeStyles(_ => ({
   },
   middleRowCard: {
     height: '20rem',
+    overflow: 'auto',
+    wordBreak: 'break-word',
   },
   bottomRowCard: {
     minHeight: '40rem',
@@ -122,7 +124,6 @@ export const WorkflowInstancePageContent: React.FC<{
           <WorkflowResult
             assessedInstance={assessedInstance}
             className={styles.middleRowCard}
-            cardClassName={styles.autoOverflow}
           />
         </Grid>
 
--- a/.changeset/gentle-fishes-swim.md
+++ b/.changeset/gentle-fishes-swim.md
@@ -1,0 +1,5 @@
+---
+"@janus-idp/backstage-plugin-orchestrator": patch
+---
+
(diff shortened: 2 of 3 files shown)


mareklibra:
Screenshot from 2024-11-06 10-53-47

---

Screenshot from 2024-11-06 10-54-42

---

Screenshot from 2024-11-06 10-55-07

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -50,6 +50,8 @@ const useStyles = makeStyles(_ => ({
   },
   middleRowCard: {
     height: '20rem',
+    overflow: 'overlay',
```
batzionb, line 53: Reading about it, seemss like "overlay" value is deprecated
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow

mareklibra: good to know, changing to "auto" then

plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -122,7 +124,6 @@ export const WorkflowInstancePageContent: React.FC<{
           <WorkflowResult
             assessedInstance={assessedInstance}
             className={styles.middleRowCard}
-            cardClassName={styles.autoOverflow}
```
batzionb, line 125: are you sure we don't need the "overflow: auto" also for the cardClassName? What if the outputs are a long list? Does it work without it?
And if not, since WorkflwoResult is only called by one place, might as well remove the property all together

mareklibra: By the other change in this PR, we newly scroll whole card content, not just the inner part. It's better than having multiple scrollbars on a small place. 
I can remove the property for all. On the other hand, there is no harm in keeping it.

batzionb: I checked myself, it must have overflow: auto only in className as you did and it works well.
Please remove cardClassNameProperty instead of making it optional

batzionb (approved):

PR #2499, opened 2024-11-06
chore: deprecate rbac plugins

04kash:
Deprecate rbac plugins as they have been migrated to backstage/community-plugins


04kash:
To be merged after comfirming that the wrapper is correct: https://github.com/janus-idp/backstage-showcase/pull/1897

schultzp2020 (approved):
/lgtm

PR #2500, opened 2024-11-07
fix(orchestrator): improve alignment of key-value pairs on the Review step

mareklibra:
Fixes: [FLPATH-1315](https://issues.redhat.com//browse/FLPATH-1315)

### DIFF
--- a/.changeset/mean-parrots-sort.md
+++ b/.changeset/mean-parrots-sort.md
@@ -1,0 +1,5 @@
+---
+"@janus-idp/backstage-plugin-orchestrator-form-react": patch
+---
+
+Fix alignment of key-values on the Review step of the Run Workflow.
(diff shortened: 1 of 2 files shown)


mareklibra:
To properly fix the issue, we would need to either change the `StructuredMetadataTable` compoent of the `@backstage/core-components` or rewrite it for our proprietary purpose or find another component doing the same thing.

Contributing upstream would be nice but will take forever to propagate the fix for us.

Rewriting it from scratch is unnecessary effort, IMO.

The proposed patch balances between the requirements and effort.

mareklibra:
Screenshot from 2024-11-07 08-42-11

mareklibra:
Closing in favor of https://github.com/redhat-developer/rhdh-plugins/pull/102

PR #2501, opened 2024-11-07
feat(lightspeed): integrate lightspeed backend 

karthikjeeyar:
## Fixes

https://issues.redhat.com/browse/RHDHPAI-26
https://issues.redhat.com/browse/RHDHPAI-38

## Description

This PR replaces backstage proxy backend with lightspeed backend and supports following features:

1. Chat history
2. Multiple Chats
3. Create new Chat conversation
4. Delete chat conversation
5. Search conversation history
6. Auto scroll to the bottom when the message is streaming.

## Screenshots

image

Search results:

searchResult

welcome prompt:

image

## Video:

https://github.com/user-attachments/assets/871267d2-49a6-47cc-9711-a0af4262a3ef

## How to test:

1.  Follow the lightspeed backend plugin [README](https://github.com/janus-idp/backstage-plugins/tree/main/plugins/lightspeed-backend) to integrate lightspeed backend in your Backstage instance.

## Unit tests:

```  
   PASS  src/hooks/__tests__/useDeleteConversation.test.tsx
  useDeleteConversation
    ✓ calls deleteConversation API and updates cache correctly (20 ms)
    ✓ invalidates cache on success if invalidateCache is true (8 ms)
    ✓ reverts cache on error (4 ms)
  ```
  ```  
    PASS  src/hooks/__tests__/useConversationMessages.test.tsx
  useFetchConversations
    ✓ should return conversations data when fetch is successful (73 ms)
    ✓ should handle loading state (5 ms)
    ✓ should handle errors (56 ms)
  useConversationMesages
    ✓ should initialize conversations with the given conversationId (5 ms)
    ✓ should update current conversation when conversationId changes (6 ms)
    ✓ should call onComplete when streaming is done (7 ms)
    ✓ should handle the invalid json error (41 ms)
    ✓ should handle input prompt and update conversations with user and bot messages (4 ms)
    ✓ should update last bot message in conversation after API response (8 ms)
    ✓ should surface API error if last bot message failed (76 ms)
    ✓ should have scrollToBottomRef defined (55 ms)
```
```
 PASS  src/utils/__tests__/lightspeed-chatbot-utils.test.ts
  getTimestampVariablesString
    ✓ should add a leading zero if the number is less than 10 (2 ms)
    ✓ should return the number as a string if it is 10 or greater
  getTimestamp
    ✓ should format a given timestamp correctly
    ✓ should handle single-digit day, month, hour, minute, and second (1 ms)
    ✓ should handle end-of-year timestamps correctly
    ✓ should handle the beginning of the epoch (0 timestamp) (1 ms)
    ✓ should handle timestamps with daylight saving time shifts
  splitJsonStrings
    ✓ should return the entire string in an array if no `}{` pattern is found (1 ms)
    ✓ should split a concatenated JSON string into individual JSON strings
    ✓ should handle a JSON string with multiple concatenated objects correctly (1 ms)
    ✓ should handle a JSON string with edge case of empty objects
    ✓ should handle a JSON string with nested braces correctly
  createMessage
    ✓ should create a user message with default values (1 ms)
    ✓ should create a bot message with custom values
  createUserMessage
    ✓ should create a user message with default name if name is not provided
    ✓ should create a user message with provided name
  createBotMessage
    ✓ should create a bot message with provided properties (1 ms)
  getMessageData
    ✓ should return content and timestamp from message.kwargs
    ✓ should handle missing kwargs properties gracefully (33 ms)
  getCategorizeMessages
    ✓ categorizes messages correctly (37 ms)
 ```

### CODE REVIEW

**Comment by yangcao77** [comment]
tested on local backstage & worked for me
Screenshot 2024-11-08 at 1 52 34 PM

File: plugins/lightspeed/src/api/LightspeedApiClient.ts [hunk 1]
```diff
@@ -0,0 +1,154 @@
...
+
+  async createMessage(
+    prompt: string,
+    selectedModel: string,
+    conversation_id: string,
+  ) {
+    const baseUrl = await this.getBaseUrl();
+    const token = await this.getUserAuthorization();
+
+    const response = await fetch(`${baseUrl}/v1/query`, {
```
yangcao77, line 41: can just use the fetchAPI backstage provided.  for example:
```
  import { FetchApi
} from '@backstage/core-plugin-api';

... ...
export type Options = {
  configApi: ConfigApi;
  identityApi: IdentityApi;
  fetchApi: FetchApi;
};

... ...

        const resp = await this.fetchApi.fetch(`${baseUrl}/conversations`, {
          method: 'POST',
          headers: {
            'Content-Type': 'application/json',
          },
        });

```

according to https://backstage.io/docs/auth/#identity-for-plugin-developers, frontend plugin should call backend plugin API via the fetchAPI, and by-default, the login information will be appended as part of the call

karthikjeeyar:
> otherthan that, I'm wondering if we should get a review from UX as well?

Yes, I have already shared the PR with April in our team channel and updated the PR based on her and Ben's comments, tagging @aprilma419 again for getting +1 in the PR



> I noticed in the screen recording @karthikjeeyar uploaded. the model list takes a fairly long time to display for a new chat session. I'm wondering do we need to fetch the list upon every new session?

No, it is not fetched upon every new session, models are fetched only on initial load but conversations in sidebar loaded immediately as I have cached that request. I can add caching to the models fetch call as well, as it will not be frequently changing.

> also the chat session shows after received the first AI response. I'm wondering if a session should be created once after a new conversation_id has been generated and responded, even without any summary?

As we have discussed in the slack, summary will not be re-analyzed once it is created and stored in db right?  so I am waiting for the first AI response to be complete to generate the summary at the end of first conversation between human and AI.

yangcao77 (approved):
changes looks good to me

rohitkrai03 (commented):
/approve

PR #2502, opened 2024-11-07
fix(orchestrator): handle fetch error that doesn't throw error

ydayagi:
FLAPTH-1725
https://issues.redhat.com/browse/FLPATH-1725


ydayagi:
@gciavarrini @batzionb

batzionb:
@ydayagi 
please remove the empty changeset file

batzionb:
use this method to generate the error, so we'll have a consistent error message structure 
https://github.com/janus-idp/backstage-plugins/blob/orchestrator-1.3/plugins/orchestrator-backend/src/service/SonataFlowService.ts#L193

ydayagi:
change not needed

PR #2503, opened 2024-11-07
fix(orchestrator): make error handling consistent in backend and UI

batzionb:
In backend: all error handling should be done by backstage error handler, this means our APIs just need to call the express next function, which will send it to the error handling middleware.
In UI: All APIs should generate an error with the message received from the backend.
includes  in addition:

- minor code improvement - removed the await on promises that call .then and .catch, that is bad practice.
- removed unused code, retrigger and MockOrchestratorClient

resolves https://issues.redhat.com/browse/FLPATH-1595 and https://issues.redhat.com/browse/FLPATH-1810


plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -297,15 +303,12 @@ function setupInternalRoutes(
       if (decision.result === AuthorizeResult.DENY) {
         manageDenyAuthorization(endpointName, endpoint, req);
       }
-      await routerApi.v2
+      return routerApi.v2
```
gciavarrini, line 306: but getWorkflowsOverview is `async` is it ok to return a Promise?

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -385,14 +384,12 @@ function setupInternalRoutes(
 
       const executeWorkflowRequestDTO = req.body;
 
-      await routerApi.v2
+      return routerApi.v2
```
gciavarrini, line 387: same as before

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -455,77 +454,59 @@ function setupInternalRoutes(
...
             definitionId: workflowId,
             cacheHandler: 'throw',
           });
 
         if (!workflowDefinition) {
-          auditLogRequestError(
-            new Error(`Couldn't fetch workflow definition ${workflowId}`),
-            endpointName,
-            endpoint,
-            req,
```
gciavarrini, line 509: why removing this?

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -455,77 +454,59 @@ function setupInternalRoutes(
...
-          return;
         }
-
         const serviceUrl = workflowDefinition.serviceUrl;
         if (!serviceUrl) {
-          auditLogRequestError(
-            new Error(`Service URL is not defined for workflow ${workflowId}`),
-            endpointName,
-            endpoint,
-            req,
```
gciavarrini, line 523: why removing this?

plugins/orchestrator-backend/src/service/router.ts
batzionb: yes

batzionb: it shouldn't be with .then and .catch, but I didn't have time to change that

plugins/orchestrator-backend/src/service/router.ts
batzionb: the error is handled in the catch. this should just throw an error. Otherwise you have all this code duplication. We want the error handling to be consistent in one place

plugins/orchestrator-backend/src/service/router.ts
batzionb: the way it should be with an async function is:
try {
  const result = await getWorkflowsOverview(buildPagination(req), getRequestFilters(req))
  res.json(result)
}  catch (err) {
    auditLogRequestError(error, endpointName, endpoint, req);
     next(error);
}

these router functions combine classic promise coding (with .then and .catch) with async await. This is bad practice.

but removing the await is the least of the evil, at least this line of code doesn't combine the two.

gciavarrini: ok, please add a jira item the improvement epic to track this.

gciavarrini (approved):

plugins/orchestrator-backend/src/service/router.ts
batzionb: done
https://issues.redhat.com/browse/FLPATH-1844

PR #2511, opened 2024-11-11
fix(orchestrator): remove completedWith status 

mareklibra:
Closes: [FLPATH-1845](https://issues.redhat.com//browse/FLPATH-1845)

This virtual "completedWithError" state is difficult to handle with pagination where we lack support for complex filtering.
So removing it fo 1.3.

The next 1.4 version will have support for the `warning` workflow result state which will semantically replace that virtual state.


ydayagi:
@mareklibra how does pagination influence completedWith?

mareklibra:
I was told that the backend search (filter) can not be done on both workflows and workflow instances. To calculate the `completedWithError`, we need result status from the instance's output.

batzionb (approved):

PR #2521, opened 2024-11-12
fix(orchestrator): remove v2/workflows/:workflowId endpoint

ydayagi:
[FLPATH-1563](https://issues.redhat.com//browse/FLPATH-1563)
https://issues.redhat.com/browse/FLPATH-1563


.changeset/neat-onions-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@janus-idp/backstage-plugin-orchestrator-common": major
+---
```
batzionb, line 3: I think an empty or patch changeset are enough

batzionb:
This UI code should also be deleted, it's not used but it calls this API
https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator/src/api/OrchestratorClient.ts#L118

ydayagi:
> This UI code should also be deleted, it's not used but it calls this API https://github.com/janus-idp/backstage-plugins/blob/main/plugins/orchestrator/src/api/OrchestratorClient.ts#L118

done

batzionb (approved):

PR #2522, opened 2024-11-12
fix(orchestrator): return instance ID also on failed execute workflow API call

batzionb:
In some cases sonataflow returns response.ok false for execute workflow API, but it does return the instance ID. In this case the UI should navigate to the instance page and the user will be able to see the issue clearly.

Includes also error handling in workflow cache service, for when the service URl is wrong.

resolves [FLPATH-1850](https://issues.redhat.com/browse/FLPATH-1850)


gciavarrini (approved):
/lgtm

PR #2532, opened 2024-11-14
fix(orchestrator): implement retrigger instance

ydayagi:
[FLPATH-1567](https://issues.redhat.com//browse/FLPATH-1567)
https://issues.redhat.com/browse/FLPATH-1567


ydayagi:
@batzionb @gciavarrini @pkliczewski

plugins/orchestrator-backend/src/service/OrchestratorService.ts
```diff
@@ -182,6 +182,22 @@ export class OrchestratorService {
...
+    cacheHandler?: CacheHandler;
+  }): Promise<boolean | undefined> {
+    const { definitionId, cacheHandler } = args;
+    const isWorkflowAvailable = this.workflowCacheService.isAvailable(
+      definitionId,
+      cacheHandler,
+    );
+    return isWorkflowAvailable
+      ? await this.sonataFlowService.retriggerInstance(args)
+      : undefined;
```
gciavarrini, line 198: In the past there was an invocation of `orchestratorService.updateInstanceInputData`. I know that there were issue, but can you explain why we don't needed anymore? To be sure we're doing the right thing.

batzionb: This is definitely the right thing
Also discussed it with @masayag so he can confirm
updating the instance with input data is a bug
Consider the following scenario: The user requested to scale up the cluster according to certain specs. He got that approved. Then the scaling failed. Now he can run retrigger, update the scale to much bigger specs, invoke retrigger and the workflow will skip the Jira ticket and go directly to the scale up!

plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -394,6 +394,46 @@ function setupInternalRoutes(
...
+      if (decision.result === AuthorizeResult.DENY) {
+        manageDenyAuthorization(endpointName, endpoint, req);
+      }
+
+      await routerApi.v2
+        .retriggerInstance(workflowId, instanceId)
+        .then(result => res.status(200).json(result))
+        .catch(error => {
+          auditLogRequestError(error, endpointName, endpoint, req);
+          res
```
batzionb, line 430:
```suggestion
next()
```

batzionb, line 430: @ydayagi I worked hard on unifying the error handling
instead of 
res.status(500).json({ message: error.message || INTERNAL_SERVER_ERROR_MESSAGE });
It should just call (next)
That would move the error handling to the standard backstage error handler and create the error in the structure the UI can handle
The UI is implemented to expect a different structure than {message: <text>}

plugins/orchestrator-backend/src/service/OrchestratorService.ts
masayag: To generalize what @batzionb said, editing input parameters during retrigger may enable the re-execution of a workflow with unapproved parameters.

ydayagi: so this is resolved?

plugins/orchestrator-backend/src/service/router.ts
ydayagi: fixed

batzionb (approved):

PR #2547, opened 2024-11-20
fix(cli): use cli directly as a fallback

gashcrumb:
This commit changes the CLI to try generating a command to export a plugin package as a dynamic plugin if the plugin package doesn't have an export-dynamic script defined.

This fixes [RHIDP-4828](https://issues.redhat.com/browse/RHIDP-4828)

### DIFF
--- a/packages/cli/src/commands/package-dynamic-plugins/command.ts
+++ b/packages/cli/src/commands/package-dynamic-plugins/command.ts
@@ -3,6 +3,7 @@ import { PackageRoles } from '@backstage/cli-node';
 import chalk from 'chalk';
 import { OptionValues } from 'commander';
 import fs from 'fs-extra';
+import { PackageJson } from 'type-fest';
 import YAML from 'yaml';
 
 import os from 'os';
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -56,20 +58,42 @@ export async function command(opts: OptionValues): Promise<void> {
...
+          });
+        } catch (err) {
+          Task.log(
+            `Encountered an error running 'yarn export-dynamic' on plugin package ${packageFilePath}, this plugin will not be packaged.  The error was ${err}`,
+          );
+        }
+      } else {
         Task.log(
-          `Encountered an error running 'yarn export-dynamic on plugin package ${packageFilePath}, this plugin will not be packaged.  The error was ${err}`,
+          `Using npx @janus-idp/cli to export plugin package ${packageFilePath}`,
```
davidfestal, line 85: Do we really need to use `npx` here ?
Can't we automatically find the place of the janus-cli being called, and use it to call the `export-dynamic-plugin` ?

gashcrumb: Yeah, that would probably be better anyways, I'll give it a go.  Thanks!

jesuino (approved):
LGTM!

PR #2550, opened 2024-11-21
fix(orchestrator): execute should allow no inputs

batzionb:
The bug is that execute fails also if field inputData is passed but is empty. 
This causes workflows without required inputs to fail.

Resolves [FLPATH-1880](https://issues.redhat.com/browse/FLPATH-1880)


batzionb:
@gciavarrini 
Please review
I didn't create a PR on main branch because of migration to rhdh-plugins
I'll create a PR there once it's in

mareklibra (approved):

PR #2564, opened 2024-11-25
chore: remove orchestrator plugin references

04kash:
- Remove references to the orchestrator plugin from renovate and catalog-info.yaml.
- Remove some of the remaining source code from orchestrator-common.


batzionb:
wait - why are the openapi files removed from orchestrator plugins?

plugins/orchestrator-common/src/openapi/openapi.yaml
```diff
@@ -1,733 +0,0 @@
-openapi: 3.1.0
```
batzionb, line 1: please don't delete the orchestrator openapi files

batzionb (approved):

PR #2573, opened 2024-11-27
fix(orchestrator): multiple bug fixes related to execution form

batzionb:
1. businessKey fixed to be only a parent assessment workflow, to avoid having workflow runs shown as assessed by themselves, and then rerun shows the inputs as readonly.
2. fixed rerendering of form on every change, which was caused by passing formData to decorator. This constant rerendering causes the form to be unusable. instead only the initial data should be passed to the decorator.


mareklibra (approved):
I prefer having spaghetti rather on a plate.

PR #2597, opened 2024-12-09
fix(cli): support single packages better

gashcrumb:
This change fixes an issue in the `package-dynamic-plugins` command that prevents it from running properly in a single plugin project which does not have workspaces.


jesuino (approved):
LGTM!

jesuino (approved):
LGTM!

PR #2601, opened 2024-12-12
fix(shared-react): move css styles from css file to the Status component

debsmita1:
**Fixes:**
https://issues.redhat.com/browse/RHIDP-5140

**Test setup:**
- Go to the shared-react package
- Run `npm run tsc`
- Run `npm run build`
- Copt the generated `dist` folder under  community-plugins -> topology ->  node_modules -> @janus-idp -> shared-react
- Run the topology plugin to view the changes

**Screenshots:**
Before:
Screenshot 2024-12-04 at 3 25 16 AM

After:
Screenshot 2024-12-12 at 5 46 04 PM

**Note:**
Update the `shared-react` package version in the topology plugin under community plugins


plugins/shared-react/src/components/common/StatusIconAndText.css
```diff
@@ -22,9 +15,3 @@
 .bs-shared-dashboard-icon {
   font-size: 1.2rem;
 }
```
ciiay, line 17: Is it okay to leave other styles in the CSS files? If we leave them in where they are now, later we may have other UI issues, what do you think? We can address those in a different story for sure, because I think it could be a common issue for other plugins which depends on shared-react.

cc. @christoph-jerolimov

ciiay (approved):
/lgtm

Thanks for the fix 💯 This solution resolved the injection issue perfectly and we can probably apply it to other missing styles issues due to CSS injection.
image

plugins/shared-react/src/components/common/StatusIconAndText.css
debsmita1: removed the css file, I checked those classes aren't required anymore

christoph-jerolimov (approved):

PR #2616, opened 2025-01-14
fix(cli): extend --ignore-version-check

gashcrumb:
This change updates the cli's --ignore-version-check flag to also ignore the check performed on a embedded transitive dependencies that are moved to peer dependencies, when it is known that the mismatch is runtime compatible.

Fixes [RHIDP-4505](https://issues.redhat.com/browse/RHIDP-4505)

@04kash @schultzp2020 FYI

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/backend-utils.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/backend-utils.ts
@@ -23,11 +23,17 @@ import path from 'path';
 
 import { Task } from '../../lib/tasks';
 
-export function addToDependenciesForModule(
-  dependency: { name: string; version: string },
-  dependencies: { [key: string]: string },
-  module: string,
-): void {
+export function addToDependenciesForModule({
+  dependency,
+  dependencies,
+  ignoreVersionCheck = [],
+  module,
+}: {
+  dependency: { name: string; version: string };
+  dependencies: { [key: string]: string };
+  ignoreVersionCheck?: string[];
+  module: string;
+}): void {
   const existingDependencyVersion = dependencies[dependency.name];
   if (existingDependencyVersion === undefined) {
     dependencies[dependency.name] = dependency.version;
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/export-dynamic-plugin/backend-utils.ts
```diff
@@ -61,9 +67,15 @@ export function addToDependenciesForModule(
...
-  throw new Error(
-    `Several incompatible versions ('${existingDependencyVersion}', '${dependency.version}') of the same transitive dependency ('${dependency.name}') for embedded module ('${module}')`,
-  );
+  if (!ignoreVersionCheck.includes(dependency.name)) {
+    throw new Error(
+      `Several incompatible versions ('${existingDependencyVersion}', '${dependency.version}') of the same transitive dependency ('${dependency.name}') for embedded module ('${module}')`,
+    );
+  } else {
+    Task.log(
+      `Several incompatible versions ('${existingDependencyVersion}', '${dependency.version}') of the same transitive dependency ('${dependency.name}') for embedded module ('${module}') however this has been overridden`,
```
davidfestal, line 76: Should we mention which version is finally kept in the peer dependencies, as it is done in the other messages above ?

04kash:
Tested these changes locally against the [1.34 Backstage Update PR](https://github.com/redhat-developer/rhdh/pull/2053) and the export-dynamic command ran without errors for all wrappers. Had to make these changes:
1. Updated @janus-idp/cli to 1.19.1
2. Added the --ignore-version-check flag to the export-dynamic command for the following wrappers:
backstage-plugin-catalog-backend-module-gitlab-org-dynamic: `--ignore-version-check @backstage/backend-defaults`
backstage-plugin-techdocs-backend-dynamic:  `--ignore-version-check @backstage/backend-plugin-api`
roadiehq-backstage-plugin-argo-cd-backend-dynamic: `--ignore-version-check @backstage/backend-common`
roadiehq-scaffolder-backend-argocd-dynamic: `--ignore-version-check @backstage/backend-common --ignore-version-check @backstage/backend-plugin-api`

kadel (approved):

PR #2630, opened 2025-02-05
chore: bump rollup to v4

kim-tsao:
Dependency bump to fix issue: https://github.com/janus-idp/backstage-plugins/pull/2234
Not a security issue, but we want to keep the package updated.  
https://issues.redhat.com/browse/RHIDP-4438


gashcrumb (approved):
:+1:

PR #2636, opened 2025-02-10
fix(shared-react): use minor version instead of major for next release

AndrienkoAleksandr:
Use minor version instead of major for next shared-react release


PatAKnight (approved):
/lgtm

PR #2645, opened 2025-03-06
fix: cli incompatibility with yarn v1

davidfestal:
Yarn v1 support for file protocol used in conjunction with resolution is limited, so that in some cases the `export-dynamic-plugin` command may fail on `yarn v1`.

This PR  accommodates this limitation without changing the existing behavior and generated output for yarn v3.

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/backend-embed-as-dependencies.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/backend-embed-as-dependencies.ts
@@ -42,6 +42,7 @@ export async function backend(opts: OptionValues): Promise<string> {
   const targetRelativePath = 'dist-dynamic';
   const target = path.join(paths.targetDir, targetRelativePath);
   const yarn = 'yarn';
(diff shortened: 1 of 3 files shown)


gashcrumb (approved):

packages/cli/src/commands/export-dynamic-plugin/backend-embed-as-dependencies.ts
```diff
@@ -659,6 +655,7 @@ function checkWorkspacePackageVersion(
 
 export function customizeForDynamicUse(options: {
   embedded: ResolvedEmbedded[];
+  isYarnV1: boolean;
```
gashcrumb, line 658: Actually you'll need to maybe [update this](https://github.com/janus-idp/backstage-plugins/blob/main/packages/cli/src/commands/export-dynamic-plugin/frontend.ts#L113-L126) as part of this PR too

gashcrumb (approved):

PR #2646, opened 2025-03-06
chore: remove audit-log-node changeset to fix release workflow

04kash:
Since the `audit-log-node` plugin was [removed from the repository](https://github.com/janus-idp/backstage-plugins/pull/2641), [this changeset](https://github.com/janus-idp/backstage-plugins/blob/main/.changeset/plugins-audit-log-node.md) needs to be deleted as well. The release workflow is attempting to release this non-existent package because of the changeset, which is causing [the workflow to fail](https://github.com/janus-idp/backstage-plugins/actions/runs/13707841310/job/38337388474).

### DIFF
--- a/.changeset/plugins-audit-log-node.md
+++ b/.changeset/plugins-audit-log-node.md
@@ -1,6 +1,0 @@
----
-"@janus-idp/backstage-plugin-audit-log-node": minor
----
-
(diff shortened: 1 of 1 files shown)


nickboldt (approved):
/lgtm

PR #2651, opened 2025-03-28
fix(cli): use the correct package name

gashcrumb:
This change ensures package-dynamic-plugin takes the package name as-is from the exported plugin package, rather than derive the name based on the original plugin package and role.  This change also ensures that plugin metadata comes from the exported package.  Finally this change removes some output that isn't necessary when detecting plugin configuration files.

Fixes [RHIDP-6652](https://issues.redhat.com/browse/RHIDP-6652)

### DIFF
--- a/.changeset/fresh-mirrors-help.md
+++ b/.changeset/fresh-mirrors-help.md
@@ -1,0 +1,5 @@
+---
+"@janus-idp/cli": patch
+---
+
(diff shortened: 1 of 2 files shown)


davidfestal (commented):
If I'm not mistaken, this will create the folder (inside the image) as `packageName`, which means, if now it is taken from the `dist-dynamic/package.json`, that it will contain the `-dynamic` suffix for both frontend and backend plugins.

I like the idea of taking all the metadata from the derived package (in `dist-dynamic`), but I think that the `-dynamic` suffix should still be removed from the folder inside the image.

gashcrumb:
Updated that @davidfestal, though I wonder, this means there's still a discrepancy between what's recorded in the `index.json` and annotations and the `package.json` for each exported plugin.

davidfestal (approved):

PR #2661, opened 2025-04-15
fix: role retrieval bug in the `metadata` command, and introduce backstage `support-versions` check during export.

davidfestal:
Fix role retrieval bug in the `metadata` command, and introduce backstage `support-versions` check during export.

### DIFF
--- a/packages/cli/src/commands/export-dynamic-plugin/command.ts
+++ b/packages/cli/src/commands/export-dynamic-plugin/command.ts
@@ -19,6 +19,7 @@ import { PackageRoles } from '@backstage/cli-node';
 import chalk from 'chalk';
 import { OptionValues } from 'commander';
 import fs from 'fs-extra';
+import * as semver from 'semver';
 
 import path from 'path';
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/export-dynamic-plugin/command.ts
```diff
@@ -74,5 +75,50 @@ export async function command(opts: OptionValues): Promise<void> {
...
   await applyDevOptions(opts, rawPkg.name, roleInfo, targetPath);
 }
+
+async function checkBackstageSupportedVersions(targetPath: string) {
+  const targetPackageFile = path.join(targetPath, 'package.json');
+  const targetPackage = await fs.readJSON(targetPackageFile);
+  const supportedVersions: string | undefined =
+    targetPackage.backstage?.['supported-versions'];
+  const backstageJson = path.join(paths.targetRoot, '/backstage.json');
+  if (fs.existsSync(backstageJson)) {
```
gashcrumb, line 89: would it better to reduce nesting and just return at this point if it doesn't exist?

packages/cli/src/commands/export-dynamic-plugin/command.ts
```diff
@@ -74,5 +75,50 @@ export async function command(opts: OptionValues): Promise<void> {
...
+      const singleVersionInSupportedVersions = semver.valid(
+        supportedVersions,
+        true,
+      );
+      const supportedVersionsRange = singleVersionInSupportedVersions
+        ? `~${supportedVersions}`
+        : supportedVersions;
+
+      if (!semver.subset(`~${backstageVersion}`, supportedVersionsRange)) {
+        const errorMessage = `The ${chalk.cyan('backstage.supported-versions')} field in the package descriptor is not compatible with the backstage version specified in the ${chalk.cyan('backstage.json')} file: ${chalk.cyan(supportedVersions)} vs ${chalk.cyan(backstageVersion)}.`;
```
gashcrumb, line 101: just a nit, but this string could probably be created on line 103 without the extra variable no?

packages/cli/src/commands/export-dynamic-plugin/command.ts
davidfestal: You're right.
I couldn't do it before, before the code wasn't extracted as a function.
I'll do it now.

packages/cli/src/commands/export-dynamic-plugin/command.ts
davidfestal: I also reuse it in line 107

packages/cli/src/commands/export-dynamic-plugin/command.ts
davidfestal: Fixed it, and rebased the code

gashcrumb (approved):

PR #2667, opened 2025-05-08
feat(cli): extend container tool to buildah

durandom:
Add `buildah` as an option to build container images, since it doesnt require a running daemon such a podman or docker

Signed-off-by: Marcel Hild <hild@b4mad.net>

### DIFF
--- a/packages/cli/src/commands/package-dynamic-plugins/command.ts
+++ b/packages/cli/src/commands/package-dynamic-plugins/command.ts
@@ -13,22 +13,59 @@ import { paths } from '../../lib/paths';
 import { Task } from '../../lib/tasks';
 
 export async function command(opts: OptionValues): Promise<void> {
-  const { exportTo, forceExport, preserveTempDir, tag, useDocker, platform } =
-    opts;
+  const {
+    exportTo,
+    forceExport,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/package-dynamic-plugins/command.ts
```diff
@@ -13,22 +13,61 @@ import { paths } from '../../lib/paths';
...
+    const allowedTools = ['docker', 'podman', 'buildah'];
+    if (!allowedTools.includes(_containerTool)) {
+      Task.error(
+        `Invalid value for --container-tool: ${_containerTool}. Allowed values are: ${allowedTools.join(', ')}`,
+      );
+      return;
+    }
+
+    // Determine the build command and version check command
+    containerBuildCommand = _containerTool === 'buildah' ? 'bud' : 'build';
```
kadel, line 64: Why do we need to use `bud` command in `buildah`, isn't this just a shortcut for `build`?

durandom: Indeed. Looks like I was looking at a previous version of buildah documentation where this was build using docker file. I'll update the PR

PR #2674, opened 2025-06-12
fix: add CEL property to WhenExpression type

Historyman:
There is a current issue with the tekton backstage plugin that requires adding the CEL property.

This cel when expression is outlined in the Tekton documentation and is currently a preview feature.

https://tekton.dev/docs/pipelines/pipelines/#use-cel-expression-in-whenexpression

### DIFF
--- a/.changeset/tiny-otters-deliver.md
+++ b/.changeset/tiny-otters-deliver.md
@@ -1,0 +1,5 @@
+---
+"@janus-idp/shared-react": patch
+---
+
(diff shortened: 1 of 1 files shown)


Historyman:
@dzemanov or @its-mitesh-kumar are you available to review this? Its a small change so I can make an enhancement to https://github.com/backstage/community-plugins/tree/main/workspaces/tekton/plugins/tekton

PR #2686, opened 2026-01-15
fix(cli): remove unnecessary singleton entries

gashcrumb:
This change removes a few entries for backstage libraries from the webpack singleton config, as these libraries can handle duplicate instances being loaded at runtime.  This change fixes [RHDHBUGS-2495](https://issues.redhat.com/browse/RHDHBUGS-2495).

A way to test this is to `yarn link` the `packages/cli` directory in your rhdh clone both at the root and in the `dynamic-plugins` folder, then rebuild rhdh and build/export the wrappers under `dynamic-plugins`

### DIFF
--- a/.changeset/late-forks-decide.md
+++ b/.changeset/late-forks-decide.md
@@ -1,0 +1,5 @@
+---
+"@janus-idp/cli": minor
+---
+
+Remove backstage packages from the webpack singleton configuration.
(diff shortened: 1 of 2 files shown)


nickboldt (commented):
lgtm but curious about the major version bump requirement here.

gashcrumb:
> lgtm but curious about the major version bump requirement here.

yeah, I changed that to minor, that's probably better.

nickboldt (approved):
j'approve! 

/lgtm
/approve

christoph-jerolimov (approved):
Thanks @gashcrumb for this deep investigation! 👍 

/lgtm
/approve

